PR #9, opened 2020-07-07
Prefer const and fix some typos

Hyuchia:
This is just mainly a visual cleanup and will probably require to make a similar PR based on the `new-drive-desktop` branch though since many of the files have been removed there.

Changes:

- Added an eslint rule to prefer `const` instead of `let` whenever a variable is not reassigned and updated all existing cases that matched that rule.
-  Fixed a few typos throughout the code
- Removed all instances of `var` from the code and changed them for `const`/`let`.

### DIFF
--- a/.electron-vue/dev-runner.js
+++ b/.electron-vue/dev-runner.js
@@ -114,7 +114,7 @@ function startMain () {
 }
 
 function startElectron () {
-  var args = [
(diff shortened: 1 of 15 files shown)


fvsegarra:
Merged, thanks a lot!

PR #12, opened 2020-07-15
Use RegExp for ignored files

Hyuchia:
**This has not been tested, do not merge before testing it first**

I have not been able to set up a working local version for development yet so this is mostly a _Request For Comments_.

1. The `.nvmrc` will allow people to just do `nvm use` and the correct node version will be selected
2. Using regex to match files to ignore, while being slower, should allow matching more files such as:

Old Windows temporal files:
```
.thumbs
.Thumbs
```

macOS loves creating garbage, apart from the `.DS_Store` file, it also creates temporary files when copying files specially from and to external drives. They follow this pattern:
```
._*
```

For example:

```
._Some Video.mp4
._my-document.docx
```

Linux and some apps also create temporal files, specially when a file is currently open and has some changes that have not been saved, these files usually follow this patterns:

```
.*~
~.*
```

For example:

```
some-file.txt~
~some-file.txt
```


fvsegarra:
nice pr; please take a look at the conflicts mentioned above and we'll merge it as soon as those are resolved. Thanks!

PR #28, opened 2021-05-31
Dv2 11 event sending UI

JoanMora:
WIP


JoanMora:
It should be merged into beta.

PR #57, opened 2021-12-22
Updated onboarding

aboredvaro:
Used `openOnboarding` instead `openOnboarding()`


AlexMenor (changes requested):
`openOnboarding()` works because of vue's magic but you actually have to provide a function there so please use `openOnboarding` instead

AlexMenor (approved):

PR #61, opened 2021-12-27
Do not skip macOS notarize

JoanMora:
Avoid skipping notarize by default.


sg-gs:
Should this still be merged @AlexMenor?

AlexMenor:
As I said this is good as is, because we only notarize in our build machine.

PR #145, opened 2022-10-26
[WT-159] fix: store encrypted bearer token

JoanVicens:
Adds Encryption to the bearer token. The token is encrypted when:
- An user logs in to the app
- The user is already logged in and starts the application

Added also added `bearerTokenEncrypted` to control if it has been encrypted or not. Mainly to avoid possible errors when people update the application when it's already running.


src/main/auth/service.ts
```diff
@@ -1,19 +1,52 @@
+import { safeStorage } from 'electron';
+import Logger from 'electron-log';
 import ConfigStore, { defaults, fieldsToSave } from '../config';
 import packageConfig from '../../../package.json';
 import { User } from '../types';
 
+const TOKEN_ENCODING = 'latin1';
```
sg-gs, line 7: does latin1 support every character a JWT can have?

src/main/auth/service.ts
```diff
@@ -1,19 +1,52 @@
...
+  ConfigStore.set('bearerToken', encryptedToken);
+  ConfigStore.set('bearerTokenEncrypted', true);
+}
+
 export function setCredentials(
   userData: User,
   mnemonic: string,
   bearerToken: string
 ) {
+  if (!safeStorage.isEncryptionAvailable()) {
```
sg-gs, line 34: Could we try to test where (SOs) this feature is not available? Maybe we are forcing some users to change the SO, which could be fatal

src/main/auth/service.ts
JoanVicens: The `latin1` is to encode only the buffer resultant of the encryption. when it's decrypted the buffer it's created with the same encoding so it should be ok

src/main/auth/service.ts
JoanVicens: There are two cases where this could be false:
- We are not on the main process: `safeStorage` is only available from the main process.
- Depending on the OS: only on MacOS can return false if the system does not have access to Keychain. On Windows and Linux once the app is ready is available.


The first one is manageable since we can use the event bus to communicate with the main process. If it fails with the second it should be fixable by the user, I'm not quite sure how the Keychain functions but the user should be able to allow the app to use it

PR #151, opened 2022-10-31
[TICK-367] fix: define transports connection option on socket connection

JoanVicens:
Seems like the transports option needs to be defined to be able to connect reliably. https://github.com/socketio/socket.io/issues/1995#issuecomment-74017995


sg-gs:
This should not be necessary if the `socket.io-client` library is up to date, for reference: 
https://socket.io/docs/v3/client-initialization/#low-level-engine-options

JoanVicens:
Yeah, this seems not to be a problem. We are not using the latest but close. 
It's version 4 tho

PR #154, opened 2022-11-09
[WT-193] feat: refresh tokens

JoanVicens:
Add a refresh token scheduler to renew tokens when they start to expire.

Added also axios HTTP client to authorize and handle authorization errors of the freshes calls


src/main/token-scheduler/token-scheduler.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+      const schedule = scheduler.schedule(taks);
+
+      const nextInvocation = schedule?.nextInvocation();
+
+      expect(nextInvocation?.getDay()).toBe(data.day.getDay());
+    }
+  );
+
+  it('shedules to refresh even if a token does not expire', () => {
```
sg-gs, line 40: typo here: _schedules_ instead of shedules

src/main/token-scheduler/token-scheduler.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    const schedule = scheduler.schedule(taks);
+
+    const nextInvocation = schedule?.nextInvocation();
+
+    expect(schedule).toBeDefined();
+    expect(nextInvocation?.getDay()).toBe(expectedExpireDay.getDay());
+  });
+
+  it('shedules to refresh even if a token is not valid', () => {
```
sg-gs, line 56: same here

src/main/token-scheduler/token-scheduler.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  it('does not schedules if any token expires', () => {
+    const scheduler = new TokenScheduler(7, [jwtWithoutExpiration]);
+
+    const schedule = scheduler.schedule(taks);
+
+    expect(schedule).not.toBeDefined();
+  });
+
+  it('does not schedules if any token is valid', () => {
```
sg-gs, line 87: _schedule_ instead of _schedules_ as _does_ is used here

sg-gs (changes requested):

PR #161, opened 2022-11-21
[_]: Fix windows auto startup

JoanVicens:
- Fixes auto-start on windows ⚠️  Could not be tested yet
- Hides auto-start on Linux since it's not implemented


sg-gs (approved):

JoanVicens:
Tested the changes on a windows machine. The application started correctly but was not hidden.

PR #164, opened 2022-11-30
[WT-246] fix: sanitize paths before validating them

JoanVicens:
Nested folders are not correctly validated on windows due to the validation method was not prepared to handle OS specific paths.

Solved by sanitizing/converting the path to a UNIX like path before sanitazing

### DIFF
--- a/src/workers/utils/name-verification.ts
+++ b/src/workers/utils/name-verification.ts
@@ -1,13 +1,21 @@
+import path from 'path';
+
 const isWindowsRootDirectory = /[a-zA-Z]:[\\/]/;
 const containsNullCharacter = /\0/g;
 
 const validations = [
   (name: string) => name.includes('../'),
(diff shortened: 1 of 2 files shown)


sg-gs:
When an issue as delicate as this happens, please add examples to see the problem in deeper detail. @JoanVicens

src/workers/utils/name-verification.ts
```diff
@@ -1,13 +1,21 @@
+import path from 'path';
+
 const isWindowsRootDirectory = /[a-zA-Z]:[\\/]/;
 const containsNullCharacter = /\0/g;
 
 const validations = [
   (name: string) => name.includes('../'),
```
fabioespinosa, line 7: One little thing I can think of here, not urgent though is

`validations: ((name: string) => boolean)[] = [`

So you don't have to type all name as strings below @JoanVicens

PR #169, opened 2022-12-15
[_] feat: improve e2e test

JoanVicens:
Added e2e test for the panel issue. Along the way made a few changes:
- Extracted `BackupFatalError` to a separate folder to be able to import it on test cases
- Extracted wait method
- Made unauthorized tests independent of each other


src/test/issues.e2e.ts
```diff
@@ -0,0 +1,149 @@
...
+      );
+
+      const numberOfIssues = await page.innerHTML(
+        '[data-test=number-sync-issues]'
+      );
+
+      expect(numberOfIssues).toBe(`${emitNErrors} files`);
+    });
+
+    test('does not display any async issue is none is emited', async () => {
```
sg-gs, line 102: typo: `if none is emitted`

sg-gs (changes requested):

src/test/issues.e2e.ts
JoanVicens: changed

PR #177, opened 2022-12-21
[_] feat: reenable auto start on linux

JoanVicens:
To enable the application auto start on Linux a file is needed to be created on `.config/autostart`.

Adds and removes the .desktop file and checks if it is present to determine if auto-start is enabled.

This solution has the downside that we cannot clean this configuration when the app is uninstalled, so a further installation would maintain the configuration of the previous installation


JoanVicens:
We could solve the downside by storing that info also on the app config. But not sure if its worth

PR #179, opened 2022-12-21
[_] test: run ok test with ruddestak js

JoanVicens:
When the @rudderstack/rudder-sdk-node dependency was added the pipeline to run the test started to fail on the rebuild after installing the dependencies.

On local sometimes work some not.

Tested to use the js library version of Rudderstak to avoid post-install errors. It compiles and test run ok but **fails to run the app itself**:
```
A JavaScript error occurred in the main process
Uncaught Exception:
ReferenceError: window is not defined
    at e.value (/home/jvalles/internxt/drive-desktop/release/app/node_modules/rudder-sdk-js/index.js:1:52943)
    at new e (/home/jvalles/internxt/drive-desktop/release/app/node_modules/rudder-sdk-js/index.js:1:52706)
    at /home/jvalles/internxt/drive-desktop/release/app/node_modules/rudder-sdk-js/index.js:1:53454
    at /home/jvalles/internxt/drive-desktop/release/app/node_modules/rudder-sdk-js/index.js:1:69
    at Object.<anonymous> (/home/jvalles/internxt/drive-desktop/release/app/node_modules/rudder-sdk-js/index.js:1:216)
    at Module._compile (node:internal/modules/cjs/loader:1116:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1169:10)
    at Object.require.extensions.<computed> [as .js] (/home/jvalles/internxt/drive-desktop/node_modules/ts-node/src/index.ts:1361:43)
    at Module.load (node:internal/modules/cjs/loader:988:32)
    at Module._load (node:internal/modules/cjs/loader:829:12)
```


fabioespinosa:
It seems the library is meant for browsers and has no indication to support electron 🤔

PR #183, opened 2023-01-03
[DT-32] feat: move files to trash when deleted

JoanVicens:
In order to move files to the trash I had to use the Drive Server WIP API on the sync and backups process.


src/workers/filesystems/remote-filesystem.ts
```diff
@@ -218,15 +221,22 @@ export function getRemoteFilesystem({
...
-          { method: 'DELETE', headers }
-        );
-      } catch (err) {
+      const result = await clients.newDrive.post(
+        `${process.env.NEW_DRIVE_URL}/drive/storage/trash/add`,
+        {
+          items: [
+            {
+              type: 'file',
+              id: fileInCache.fileId,
```
sg-gs, line 230: Does the trash require the `fileId` from the network instead of the `file.id`?

sg-gs (changes requested):

src/workers/filesystems/remote-filesystem.ts
JoanVicens: Yes

PR #184, opened 2023-01-03
[DT-33] Feat: move folders to trash when deleted

JoanVicens:
- Changed the behavior when a folder is deleted to move it to the trash.
- Changed also the files that need to be deleted to exclude the files that are in a deleted folder, this is needed to be able to restore the full folder with its contents from the web.

ℹ️ This depends on #183 to have the authorized client on the remote file system. (This branch is based on the one from #183)


src/workers/sync/sync.ts
```diff
@@ -79,15 +79,36 @@ class Sync extends Process {
       this.consumePullQueue(pullFromLocal, this.local, this.remote),
       this.consumePullQueue(pullFromRemote, this.remote, this.local),
     ]);
+
+    const foldersDeletedInLocal = await this.listDeletedFolders(
```
sg-gs, line 83: Please explain here why a change on the sync is required and how it affects to the sync process, also, remember to update the Desktop documentation according to this

JoanVicens: When a folder was deleted, the folder and all its contents were deleted, but with trash, we have to move only the one deleted. Not doing so will make it impossible to restore a folder from the trash since the contents will be still deleted.

So the change is only to get the deleted folders before actually deleting anything and filtering the files that don't need to be deleted because they are in a folder that has been deleted

PR #199, opened 2023-02-06
[DT-38] Handle folders independently of files

JoanVicens:
The main objective of this PR is to start decoupling the creation of folders from the files.

### Why I did the changes:
Previously the folders were created "on demand" by the `pullFile` methods from the file systems. This needs to change in order to be able to create empty folders and handle folder renames. Having a simpler method on `pullFile` will also improve its mutability, this will be handy since empty files will be added in the future.

During the task found that Sync and Backups share some logic (both extend from Process) to interact with the file systems, that's why to avoid unwanted changes on backups I've created a new Remote and Local file system. They are basically the same with the minor change with the Listing object that they use.

### Changes

Each of these points its module inside sync

#### ItemState
Added a class that contains the delta value of an item. More properties will be added soon.
Extracted the generation of the previously deltas (`Record<string, Delta>`) to ItemDelta(`Record<string, ItemState>`)

#### Listing

The Listing object that was the same for the "Remote" and "Local" now there are 3 different types:
- Remote Listings: Listings generated from the server response. It contains the server Id
- Local Listings: Listings generated from the local file system. It contains a local Id
- Listings: Listings generated from the synchronization result. It contains the local and server Id

##### ItemMetaData
Extracted class from the current value of the Listing type plus the attribute `isFolder`

##### Listing store
 - Minor changes due to folder structure changed
 - When a stored listing does not comply with the desired one it simply returns null. This will make the sync process recalculate them. This is needed since there is missing info that we cannot obtain

#### Actions

To generate the actions needed to sync the file systems I've changed the current private method for the function `generateHierarchyActions`. This function determines the action needed for each item applying the [Strategy Pattern](https://en.wikipedia.org/wiki/Strategy_pattern). 

It generates an action for each item that needs it, creating the first one that matches the criteria. This will be needed for the rename and move actions since items affected by those actions match the delete and create criteria.


src/workers/filesystems/infrastructure/SyncRemoteFileSystem.ts
```diff
@@ -0,0 +1,666 @@
...
+              fileId: file.fileId,
+              modificationTime,
+              size: file.size,
+            };
+          });
+
+        foldersInThisFolder.forEach((folder) => {
+          const name =
+            currentName +
+            crypt.decryptName(
```
sg-gs, line 187: Worth to start using `plainName` if the field exists here. Therefore we will avoid having to publish a new Desktop App to remove the decrypt-name-always assumption.

src/workers/filesystems/infrastructure/SyncRemoteFileSystem.ts
```diff
@@ -0,0 +1,666 @@
...
+    async renameFile(oldName: string, newName: string): Promise<void> {
+      const fileInCache = cache[oldName];
+      const newNameBase = path.parse(newName).name;
+
+      try {
+        const res = await httpRequest(
+          `${process.env.API_URL}/api/storage/file/${fileInCache.fileId}/meta`,
+          {
+            method: 'POST',
+            headers: { ...headers, 'internxt-mnemonic': mnemonic },
```
sg-gs, line 244: Definitely remove the header 'internxt-mnemonic'

src/workers/filesystems/infrastructure/SyncRemoteFileSystem.ts
```diff
@@ -0,0 +1,666 @@
...
+      if (route.length > 0) {
+        for (const [i, folderName] of route.entries()) {
+          const routeToThisPoint = route.slice(0, i + 1).join('/');
+
+          const folderInCache = cache[routeToThisPoint];
+
+          if (folderInCache) lastParentId = folderInCache.id;
+          else {
+            if (createFolderQueue.listeners(routeToThisPoint).length === 0) {
+              httpRequest(`${process.env.API_URL}/api/storage/folder`, {
```
sg-gs, line 305: Use `await` instead of `then` chaining.

src/workers/filesystems/infrastructure/SyncRemoteFileSystem.ts
```diff
@@ -0,0 +1,666 @@
...
+                method: 'POST',
+                headers,
+                body: JSON.stringify({
+                  folderName,
+                  parentFolderId: lastParentId,
+                }),
+              })
+                .then(async (res) => {
```
sg-gs, line 313: Mixing `async` and `then` is dangerous

src/workers/filesystems/infrastructure/SyncRemoteFileSystem.ts
```diff
@@ -0,0 +1,666 @@
...
+                  };
+                  createFolderQueue.emit(routeToThisPoint, createdFolder.id);
+                })
+                .catch(async (err) => {
+                  createFolderQueue.emit(routeToThisPoint, err);
+                });
+            }
+
+            try {
+              lastParentId = await new Promise((resolve, reject) =>
```
sg-gs, line 343: I see what are you doing here. You are not using async and await to have time to listen to the event finished. This approach could be done, but **always** put the listener **before** the emit that you are doing, otherwise the risk of missing an event is unavoidable. 

Suggestion: I would recommend using `queue` from the `async` library which is a queue that works with async-await or callbacks and can do what you are trying to do here in a cleaner, abstracted and easier-to-use way.

src/workers/filesystems/infrastructure/SyncRemoteFileSystem.ts
```diff
@@ -0,0 +1,666 @@
...
+      const n = route.at(-1);
+
+      const parentFolderRoute = route.slice(0, route.length - 1).join('/');
+
+      const folderInCache = cache[parentFolderRoute];
+
+      const lastParentId =
+        folderInCache !== undefined ? folderInCache.id : baseFolderId;
+
+      httpRequest(`${process.env.API_URL}/api/storage/folder`, {
```
sg-gs, line 496: same here, avoid the async-then mix

src/workers/sync/ItemState/domain/ItemState.ts
```diff
@@ -0,0 +1,15 @@
+// import { ItemKind } from '../../../../shared/ItemKind';
```
sg-gs, line 1: remove comments from here

sg-gs (changes requested):

src/workers/filesystems/infrastructure/SyncLocalFileSystem.ts
```diff
@@ -0,0 +1,363 @@
...
+
+    async smokeTest() {
+      try {
+        await fs.access(localPath, constants.R_OK | constants.W_OK);
+        await fs.lstat(localPath);
+      } catch (err) {
+        const systemError = err as { code?: string };
+        if (systemError.code === 'ENOENT') {
+          throw new ProcessFatalError(
+            'BASE_DIRECTORY_DOES_NOT_EXIST',
```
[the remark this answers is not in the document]
JoanVicens: We can but at the moment this parameter is a union type (`ProcessFatalErrorName`) used on several points of the app, from the file systems to the UI and the test. So I would prefer to do this change isolated

src/workers/filesystems/infrastructure/SyncLocalFileSystem.ts
```diff
@@ -0,0 +1,363 @@
...
+      return new Promise((resolve, reject) => {
+        const tmpFilePath = getTempFilePath();
+
+        const { stream, ...sourceWithoutStream } = source;
+
+        Logger.debug(`Downloading ${name} to temp location ${tmpFilePath}`);
+
+        const writeStream = createWriteStream(tmpFilePath);
+
+        stream.on('data', (chunk) => writeStream.write(chunk));
```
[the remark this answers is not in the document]
JoanVicens: Changed

src/workers/filesystems/application/PullFolderQueueConsumer.ts
```diff
@@ -0,0 +1,82 @@
...
+    }, [] as QueueByNestLevel);
+  }
+
+  async consume(queue: Array<string>): Promise<void> {
+    const queuesByLevel = this.divideByLevel(queue);
+
+    const tasks = queuesByLevel.map(
+      (folders: Array<string>, level: number) => async (): Promise<void> => {
+        const results = await Promise.allSettled(
+          folders.map(async (folder) => {
```
sg-gs, line 49: is this concurrency caped in chunks of X folders? We cannot just `pullFolder` at crazy and assume that problems are not going to arise. Please split them into chunks of 10 and then iterate as is currently being done.

src/workers/filesystems/infrastructure/SyncLocalFileSystem.ts
```diff
@@ -0,0 +1,341 @@
...
+        if (err.code !== 'ENOENT')
+          throw new ProcessError(
+            'UNKNOWN',
+            createErrorDetails(err, 'Deleting a file locally', `Name: ${name}`)
+          );
+      }
+    },
+
+    pullFile(name: string, source: Source) {
+      return new Promise(async (resolve, reject) => {
```
sg-gs, line 171: do not mix async and Promise, please use try-catch plus async-await

sg-gs: If you need to catch the error to convert it, do this inside the catch:

`throw new ProcessError(...`

src/workers/filesystems/infrastructure/SyncLocalFileSystem.ts
```diff
@@ -0,0 +1,341 @@
...
+    renameFile(oldName: string, newName: string) {
+      const oldActualPath = getActualPath(oldName);
+      const newActualPath = getActualPath(newName);
+      return saferRenameFile(oldActualPath, newActualPath);
+    },
+
+    async existsFolder(name: string): Promise<boolean> {
+      const actualPath = getActualPath(name);
+      try {
+        await fs.access(actualPath);
```
sg-gs, line 230: why not use `fs. existsSync ` ?

src/workers/filesystems/infrastructure/SyncRemoteFileSystem.ts
JoanVicens: Removed

src/workers/sync/ItemState/domain/ItemState.ts
JoanVicens: Cleaned

src/workers/filesystems/infrastructure/SyncRemoteFileSystem.ts
JoanVicens: I've actually deleted this part since the folders are created beforehand. On the new versions of the file systems (`SyncRemoteFileSystem` and `SyncLocalFileSystem`), the method pull file does only that, pull the file into the file system.

sg-gs (changes requested):

src/workers/sync/listing-store.ts
```diff
@@ -1,39 +0,0 @@
-import ConfigStore from '../../main/config';
-import { Listing } from '../types';
-import { ListingStore } from './sync';
-
-export default function getListingStore(): ListingStore {
-  function isOldListingFormat(listing: Record<string, any>): boolean {
```
sg-gs, line 6: Is the old listing format transformation no longer required?

sg-gs (changes requested):

src/workers/sync/listing-store.ts
JoanVicens: > When a stored listing does not comply with the desired one it simply returns null. This will make the sync process recalculate them. This is needed since there is missing info that we cannot obtain

To know if a local file is the same when moved or renamed, we need to store the files `ino` and `dev`. We could loop thru the stored listing, get the new info and update it, but it's simpler to redo it

src/workers/filesystems/infrastructure/SyncRemoteFileSystem.ts
JoanVicens: Created a pr to return the plain name on drive-server https://github.com/internxt/drive-server/pull/385

src/workers/filesystems/infrastructure/SyncLocalFileSystem.ts
JoanVicens: Not sure why `fs.access` was chosen. The only thing it comes to me is to ignore system folders

PR #205, opened 2023-03-09
[DT-41]:(fix) Handle unable to get device name issue

PixoDev:
When unable to get the device name, for whatever reason:

1. Show a fallback name instead of an error, will be "Your device"
2. Display an issue in a new Issues tab called "General"

### DIFF
--- a/src/main/device/service.ts
+++ b/src/main/device/service.ts
@@ -1,45 +1,63 @@
 import { aes } from '@internxt/lib';
 import { dialog } from 'electron';
 import fetch from 'electron-fetch';
+
 import os from 'os';
 import path from 'path';
-
+import { addProcessIssue } from '../background-processes/process-issues';
 import { getHeaders } from '../auth/service';
 import configStore from '../config';
 
 export type Device = { name: string; id: number; bucket: string };
 
+const addUnknownDeviceIssue = (error: Error) => {
+  addProcessIssue({
+    errorName: 'UNKNOWN_DEVICE_NAME',
+    kind: 'LOCAL',
+    name: error.name,
+    action: 'GET_DEVICE_NAME_ERROR',
+    errorDetails: {
+      action: '',
+      message: error.message,
+      code: '',
+      stack: error.stack || '',
+    },
+    process: 'GENERAL',
+  });
+};
+
+async function createDevice(deviceName: string) {
+  return fetch(`${process.env.API_URL}/api/backup/deviceAsFolder`, {
+    method: 'POST',
+    headers: getHeaders(true),
+    body: JSON.stringify({ deviceName }),
+  });
+}
+
+async function tryToCreateDeviceWithDifferentNames(): Promise<Device> {
+  let res = await createDevice(os.hostname());
+
+  let i = 1;
+
+  while (res.status === 409 && i <= 10) {
+    res = await createDevice(`${os.hostname()} (${i})`);
+    i++;
+  }
+
+  if (!res.ok)
+    res = await createDevice(
+      `${os.hostname()} (${new Date().valueOf() % 1000})`
+    );
+
+  if (res.ok) {
+    return res.json();
+  } else {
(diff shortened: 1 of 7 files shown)


src/main/device/service.ts
```diff
@@ -1,45 +1,63 @@
...
+      message: error.message,
+      code: '',
+      stack: error.stack || '',
+    },
+    process: 'GENERAL',
+  });
+};
 
-    let i = 1;
+async function createDevice(deviceName: string) {
```
sg-gs, line 29: `async` is not required here

src/main/device/service.ts
```diff
@@ -1,45 +1,63 @@
...
-    let res = await createDevice(os.hostname());
+const addUnknownDeviceIssue = (error: Error) => {
+  addProcessIssue({
+    errorName: 'UNKNOWN_DEVICE_NAME',
+    kind: 'LOCAL',
+    name: error.name,
+    action: 'GET_DEVICE_NAME_ERROR',
+    errorDetails: {
+      action: '',
+      message: error.message,
```
sg-gs, line 21: Will this issue provide enough context about what happened?

src/main/device/service.ts
```diff
@@ -1,45 +1,63 @@
...
 
+  if (!res.ok)
+    res = await createDevice(
+      `${os.hostname()} (${new Date().valueOf() % 1000})`
+    );
+
+  if (res.ok) {
+    return res.json();
+  } else {
+    const error = new Error('Could not create device trying different names');
```
sg-gs, line 55: Questions reading this error on the logs, that could arise:
- Which names have been tried? 
- How many?

sg-gs (changes requested):

src/main/device/service.ts
PixoDev: The operation is retried up to 10 times

Names that we try:
- os.hostname
- os.hostname + UNIX timestamp

src/main/device/service.ts
PixoDev: Yes, the http error is proxied to this function, so if the problem is due to server problems such 500, we will see the error in the logs. Since this is a plain error, some properties are empty because we simply need to match the TS type to match the "issue" type

src/workers/types.ts
```diff
@@ -190,16 +193,19 @@ type ProcessInfoBase = {
...
+export type GeneralProcessIssue = 'GET_DEVICE_NAME_ERROR';
 export type ProcessIssue = ProcessInfoBase & {
   action:
     | 'PULL_ERROR'
     | 'RENAME_ERROR'
     | 'RENAME_ERROR'
     | 'DELETE_ERROR'
-    | 'METADATA_READ_ERROR';
+    | 'METADATA_READ_ERROR'
+    | GeneralProcessIssue;
```
JoanVicens, line 204: I think it would be better to have an entirely separate list with the `GENERAL` issues since there is data that is not relevant to a general issue as the `FileSystemKind` and the actions make reference to file systems actions

PixoDev: ok, sounds better

src/main/device/service.ts
sg-gs: Okay, but maybe adding this feedback to the logs, even in a simple manner could help to understand which names were tried and why this failed. So is not my doubt but anyone's doubt when the time to provide support comes. 

For instance, if some user has a non-supported or not handled hostname by the app (strange characters), we should be able to have all the context on that and realize the issue in a simple manner. If that situation comes, without feedback on which names have been tried, we should need another release to see what is going on. So, in the case of the desktop platform, I suggest having more rich context even if it sounds obvious than having just a bit and not being able to determine what is happening, at least until we have a faster way of releasing and more flexibility to avoid some of these details

sg-gs (changes requested):

JoanVicens (approved):

PR #231, opened 2023-04-26
[_] fix: stop uploading on connection lost

JoanVicens:


JoanVicens:
@sg-gs this should solve solve the problem

src/workers/filesystems/infrastructure/SyncRemoteFileSystem.ts
```diff
@@ -361,6 +365,13 @@ export function getRemoteFilesystem({
             fileSize: source.size,
             source: source.stream,
           });
+
+          abortSignal.addEventListener('abort', () => {
+            Logger.debug(`[SYNC REMOTE FS] Aborting upload for ${name}`);
+            // localUpload.uploadCancel(state);
```
sg-gs, line 371: Should this be commented?

sg-gs (changes requested):

PR #235, opened 2023-05-08
[PB-458]:(feature) Add Sentry for main and renderer process

PixoDev:
Add Sentry to desktop app, now errors can be reported using `reportError` function from the main and renderer process. We should add source maps uploading at some points however given the current webpack configuration that would involve more development time + investigation.


PixoDev:
I see in the diff there's multiple white spaces changes in the diff, do you know why it could be? @JoanVicens

JoanVicens (approved):

PR #236, opened 2023-05-11
[PB-405]:(feature) Sync remote files and folders with a local db

PixoDev:
Allow to sync progressively the remote files and folders with a local db.

A remote sync can be triggered any time and will pull the items updatedAt from a given date, if no date is provided, all the items will be returned and synced which will lead to a big initial sync where the local db is initially populated.


src/main/database/driveFoldersDB.ts
```diff
@@ -0,0 +1,110 @@
...
+    type TEXT NOT NULL,
+    id INTEGER UNIQUE NOT NULL,
+    parentId INTEGER,
+    bucket TEXT,
+    userId INTEGER, 
+    createdAt TEXT,
+    updatedAt TEXT NOT NULL,
+    uuid TEXT NOT NULL,
+    plainName TEXT,
+    status TEXT NOT NULL,
```
PixoDev, line 16: @sg-gs This will fail right now as the API is not returning a folder status yet

src/main/analytics/user-handlers.ts
```diff
@@ -2,15 +2,43 @@ import { ipcMain } from 'electron';
...
+    },
+    {
+      fetchFilesLimitPerRequest: 50,
+      fetchFoldersLimitPerRequest: 50,
+      syncFiles: true,
+      syncFolders: true,
+    }
+  );
+
+  remoteSyncManager.onStatusChange((newStatus) => {
```
PixoDev, line 27: This is for demo purposes, we will need to connect this with some UI feedback at some point

src/main/database/adapters/base.ts
```diff
@@ -0,0 +1,39 @@
+export abstract class DatabaseAdapter<DatabaseItemType> {
```
PixoDev, line 1: This is a base class to implement database adapters, is an initial concept but the idea behind this is that you can plug in whatever underlying database you want, even an in memory one, or a remote one

tsconfig.test.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
```
PixoDev, line 3: @JoanVicens No more TS errors for tests

src/main/database/driveFoldersDB.ts
sg-gs: Still working on it. Could we set it now, using the fields deleted and removed? I am still applying this change to files table which is 20x bigger than the folders one, so it will last a bit more to finish.

src/main/database/driveFilesDB.ts
```diff
@@ -0,0 +1,126 @@
+import { DatabaseAdapter } from './adapters/base';
+import { RemoteSyncedFile } from '../remote-sync/helpers';
+import SQLite from 'better-sqlite3';
+
+export function generateCreateTableQuery(): string {
+  const query = `CREATE TABLE IF NOT EXISTS drive_files (
```
sg-gs, line 6: Should we use Sequelize to create a mini-backend as a layer to make this easier to maintain?

sg-gs: Take into count that all the migrations history will be easy to run and keep track of, so each time we can run the migrations if the app autoupdates.

src/main/database/driveFoldersDB.ts
```diff
@@ -0,0 +1,110 @@
...
+  async connect(): Promise<{ success: boolean }> {
+    this._db = SQLite('drive_synced.db', {});
+
+    this.db.prepare(generateCreateTableQuery()).run();
+    return {
+      success: this._db.open,
+    };
+  }
+
+  async get(folderUuid: string) {
```
sg-gs, line 34: Could we use direct references to the entities as types?

So, instead of:
`async get(folderUuid: string) {`

We can do:
`async get(folderUuid: RemoteSyncedFolder['uuid']) {`

Also, as this is the folders database, I find it a bit verbose to say `folderThis` `folderThat` to the namings. So, as you have called the methods get instead of getFolder, I would suggest calling params as `uuid` instead of `folderUuid`, just adding the entity before if it is a foreign entity, like `fileUuid`, etc

src/main/remote-sync/RemoteSyncManager.test.ts
```diff
@@ -0,0 +1,284 @@
...
+  const result: RemoteSyncedFolder = {
+    plainName: `folder_${Date.now()}`,
+    id: Date.now(),
+    type: 'folder',
+    bucket: `bucket_${Date.now()}`,
+    userId: 567,
+    createdAt: new Date().toISOString(),
+    updatedAt: new Date().toISOString(),
+    parentId: null,
+    uuid: `xxxxx-xxxxx-${Date.now()}`,
```
sg-gs, line 57: Could we use `v4()` from 'uuid' to generate UUIDs and therefore, make it nearer to the real cases?

src/main/remote-sync/RemoteSyncManager.test.ts
```diff
@@ -0,0 +1,284 @@
...
+  });
+
+  it('Should retry N times and then stop if sync does not succeed', async () => {
+    (global.fetch as jest.Mock).mockImplementation(() =>
+      Promise.reject('Fail on purpose')
+    );
+
+    await sut.startRemoteSync();
+
+    await sleep(50);
```
sg-gs, line 213: why?

src/main/remote-sync/RemoteSyncManager.ts
```diff
@@ -0,0 +1,347 @@
...
+    },
+    private config: {
+      fetchFilesLimitPerRequest: number;
+      fetchFoldersLimitPerRequest: number;
+      syncFiles: boolean;
+      syncFolders: boolean;
+    }
+  ) {}
+
+  onStatusChange(callback: (newStatus: RemoteSyncStatus) => void) {
```
sg-gs, line 35: Could be useful to use an array to store the listeners of this event in case they are multiple? Replacing it could be a bit confusing.

Using an `EventEmitter` could provide this way of working by default (like extending it)

src/main/remote-sync/RemoteSyncManager.ts
```diff
@@ -0,0 +1,347 @@
...
+    return this.status === 'SYNCED';
+  }
+
+  /**
+   * Triggers a remote sync so we can populate the localDB, this sync
+   * is global and starts pulling all the files the user has in remote.
+   *
+   * Throws an error if there's a sync in progress for this class instance
+   */
+  async startRemoteSync() {
```
sg-gs, line 52: Will add an `smokeTest` here:
- Check the database is okay and exists

So, as the current sync, we ensure that nothing strange is going on when we try to sync something

src/main/remote-sync/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+  remoteSyncStore.set(
+    'lastFoldersSyncAt',
+    new Date(date.getTime() - offsetMs).toISOString()
+  );
+  return date;
+}
+
+export type RemoteSyncedFile = {
+  id: number;
+  uuid?: string;
```
sg-gs, line 43: This always exists

src/main/remote-sync/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+  fileId: string;
+  type: string;
+  size?: string;
```
sg-gs, line 46: This always exists and it is a number

src/main/remote-sync/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+  bucket?: string;
```
sg-gs, line 47: This always exists, could be null, but the field exists always, is not undefined

src/main/remote-sync/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+  folderId: number;
+  folderUuid?: string;
+  encryptVersion?: string;
+  userId?: number;
```
sg-gs, line 51: This always exists

src/main/remote-sync/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+  modificationTime?: string;
+  createdAt?: string;
```
sg-gs, line 53: This always exists

src/main/remote-sync/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+export type RemoteSyncedFile = {
+  id: number;
+  uuid?: string;
+  fileId: string;
+  type: string;
+  size?: string;
+  bucket?: string;
+  folderId: number;
+  folderUuid?: string;
+  encryptVersion?: string;
```
sg-gs, line 50: Avoid knowing about this field, it is deprecated

sg-gs (changes requested):

src/main/remote-sync/helpers.ts
JoanVicens: Size is a number but comes as a string from the backend. The bug is being reported

src/main/analytics/user-handlers.ts
```diff
@@ -2,15 +2,43 @@ import { ipcMain } from 'electron';
...
+import { DriveFilesDB } from '../database/driveFilesDB';
+import { DriveFoldersDB } from '../database/driveFoldersDB';
+import Logger from 'electron-log';
+import { clearRemoteSyncStore } from '../remote-sync/helpers';
+import { reportError } from '../bug-report/service';
 
 eventBus.on('USER_LOGGED_IN', () => {
-	userSignin();
+  userSignin();
+  const remoteSyncManager = new RemoteSyncManager(
```
JoanVicens, line 14: We should not initialize the local db on the analytics "module". The structure is not perfect, let's try to keep it as clean as possible. `USER_LOGGED_IN` can be listened to as long as you are on the main process.

src/main/database/adapters/base.ts
JoanVicens: If the class is abstract and all the methods are abstract as well why don't use an interface?

PixoDev: because interfaces doesn't allow you to add base methods like in an abstract class that you will need to implement otherwise, imagine we need to add a common method to all the classes that uses this interface, we would need to manually add it to each class, with an abstract class we can simply add the common method to the abstract class

src/main/analytics/user-handlers.ts
PixoDev: just added this here since I don't know yet how are we going to use this and when, but we can move to your way

src/main/remote-sync/RemoteSyncManager.ts
JoanVicens: +1 to have an event emitter, but I would prefer to have it injected rather than extending from it.

src/main/remote-sync/RemoteSyncManager.ts
PixoDev: This is a promise, when you call .connect() on the database adapter, it can throw an error if something goes wrong, will fail to start this in that case

src/main/remote-sync/helpers.ts
PixoDev: uuid is not coming in the file response right now, can you add it?

src/main/remote-sync/RemoteSyncManager.ts
```diff
@@ -0,0 +1,347 @@
...
+export class RemoteSyncManager {
+  private foldersSyncStatus: RemoteSyncStatus = 'IDLE';
+  private filesSyncStatus: RemoteSyncStatus = 'IDLE';
+  private status: RemoteSyncStatus = 'IDLE';
+  private onStatusChangeCallback: (newStatus: RemoteSyncStatus) => void =
+    () => {
+      return;
+    };
+  constructor(
+    private db: {
```
JoanVicens, line 23: Why don't limit the RemoteSyncManager to one database? on the methods, we are doing the same for files and folders. We could extract the `syncRemoteFiles` and `syncRemoteFolders` this we are not coupled to the data fetching. It will be easier to test since will be no need to mock the global fetch and will need less than half of the test since there will be no combinations to test (I'm thinking of cases where `syncFiles` and `syncFolders` are different of `true`)

src/main/remote-sync/RemoteSyncManager.ts
```diff
@@ -0,0 +1,347 @@
...
+
+    let query = `?limit=${this.config.fetchFilesLimitPerRequest}&offset=0&status=ALL`;
+
+    if (updatedAtCheckpoint) {
+      query += `&updatedAt=${updatedAtCheckpoint.toISOString()}`;
+    }
+
+    const path = `${process.env.NEW_DRIVE_URL}/drive/files${query}`;
+
+    const response = await fetch(path, {
```
JoanVicens, line 270: We added an Axios client with the authentication built-in and with unauthorized response handler. It can be obtained by the `getClient` function on `src/shared/HttpClient/main-process-client.ts`
 [src/shared/HttpClient/HttpClient.ts](https://github.com/internxt/drive-desktop/blob/8cd7c83f18f7ac064a10fffb440f9e67453df9b7/src/shared/HttpClient/HttpClient.ts#L12)

package.json
```diff
@@ -253,8 +258,10 @@
     "@internxt/sdk": "^1.4.22",
     "@sentry/electron": "^4.5.0",
     "@sentry/react": "^7.51.0",
+    "@types/better-sqlite3": "^7.6.4",
     "async": "^3.2.4",
     "axios": "^1.1.3",
+    "better-sqlite3": "^8.3.0",
```
JoanVicens, line 264: This is a native dependency, it should be installed on "./release/app"

src/main/remote-sync/RemoteSyncManager.ts
PixoDev: but this is coupled to the data fetching, is on purpose, that's why is called RemoteSyncManager, cause it syncs from remote, if we are going to sync from another source, we would need another system don't you think? I understand the idea but if we need to place stuff in the local db in another way, you can simply use the DriveFilesDB or the DriveFoldersDB

src/main/database/adapters/base.ts
JoanVicens: Hmm, but we are assuming all extensions of this class will need that and they might not. If the goal is to be able to "plug in whatever underlying database you want" we need to have only the methods needed to interact with the database. In this regard, the method `connect` should not be part of it either as not all databases might require to connect. There is an example on the test, you need to declare a method that is not being used anywhere since the adapter is memory and it does not need to connect.

Said that I don't really mind keeping it as an abstract class but then the classes 
`DriveFilesDB` and `DriveFoldersDB` should extend and not implement. By implementing the class any method even the non-abstract needs to be implemented.

JoanVicens:
Following up with some of the comments, and share how the app is organized.

src/main/database/adapters/base.ts
PixoDev: I don't get the point of extending an abstract class instead of implementing it, it was on purpose, all the adapters need to implement the methods described in the abstract class (create, get, update and remove), if you don't need `connect` for your adapter case, just return a promise like this `Promise.resolve({success: true})` and you are done, but the other methods are mandatory for the adapter to work. There's always cases where a class doesn't fit, but I believe this implementation is flexible enough to solve those cases, if you create an inMemoryAdapter, just return a promise as I said and implement the other methods as the abstract class defines.

The test case doesn't fail because all the methods are mocked using `jest.fn` but if you check the DriveFilesDB, when you try to do an operation without calling connect first, it throws an error, that's on purpose because you need to open a connection first

src/main/remote-sync/RemoteSyncManager.ts
```diff
@@ -0,0 +1,347 @@
...
+  }
+
+  private checkRemoteSyncStatus() {
+    // We only syncing files
+    if (
+      this.config.syncFiles &&
+      !this.config.syncFolders &&
+      this.filesSyncStatus === 'SYNCED'
+    ) {
+      this.changeStatus('SYNCED');
```
JoanVicens, line 97: Wouldn't we need an early return after changing the status? If we are only syncing files but the `foldersSyncStatus` is `SYNC_FAILED` we are going to change it here but will be overridden after `Files OR Folders sync failed, RemoteSync Failed`

src/main/remote-sync/RemoteSyncManager.ts
JoanVicens: The goal of extracting the data fetching is to avoid having to redo things and improve testability. As the file fetching is already implemented on the remote filesystem bus as it's coupled to the file system you need to implement it again on the manager and I'm having to implement it for Webdav. Plus being coupled makes it harder to extend the functionality, and keep it maintained

PixoDev: Ok, so for example the RemoteSyncManager should receive maybe an input of `RemoteSyncedFile[]` and `RemoteSyncedFolder[]` but without containing the logic to retrieve those inputs?

src/main/remote-sync/RemoteSyncManager.ts
PixoDev: you are right, modified it

src/main/remote-sync/helpers.ts
sg-gs: Done and deployed. Ref: https://github.com/internxt/drive-server-wip/commit/58250b1cedb87990e0d2e129ac57f8a9ffc9318a

PixoDev:
> Following up with some of the comments:
> 
> On the main process, we handle user-related events, like login and logout, token refreshes, drive usage, metrics, etc.
> 
> All sync-related things are on the workers/sync. Therefore the database initialization should be made there. The workers/sync/index.ts will be loaded and executed when the sync process starts.
> 
> Instead of doing that, what could be done as well is initialize it on the main process and listen to the sync actions done to interact with the database. Every sync action (pull file/folder, delete, etc) has its event (or should have its event) so they can be listened from the main process. This way we avoid initializing it on a tab.
> 
> The second option is being used to create the thumbnails so you can see what could look like

I added a separated handlers file for the remote sync, right now it listens to USER_LOGGED_IN event only, once we define better when this will be called we can add the events we want

src/main/remote-sync/RemoteSyncManager.ts
PixoDev: @JoanVicens  what do you mean by injected rather than extending from it?

src/main/remote-sync/RemoteSyncManager.ts
PixoDev: I added an smokeTest function that checks for the RemoteSyncManager status, database issues will be handled by the database adapter in the connect() method. Anyways database operations are success or die, if something goes wrong during database insert, the sync fails

src/main/analytics/user-handlers.ts
PixoDev: Moved this to own remote sync handlers file

src/main/remote-sync/RemoteSyncManager.ts
PixoDev: I added an httpClient that we pass via constructor

PixoDev:
@JoanVicens I noticed that the tests are failing on Github Actions now due to `yarn --ignore-scripts`, then electron is not installed, the test fails even if I mock electron-store, do you have any idea about how to solve this?

src/main/database/driveFilesDB.ts
sg-gs: We can use TypeORM then, which seems to have support and is the alternative to Sequelize: 
https://www.npmjs.com/package/typeorm-with-better-sqlite3

sg-gs: Repo here, as that package seems to be an adaptation: 
https://github.com/typeorm/typeorm#readme

src/main/remote-sync/RemoteSyncManager.ts
PixoDev: Now callbacks are stored in an array so multiple callbacks can be passed

src/main/database/entities/DriveFile.ts
```diff
@@ -0,0 +1,99 @@
...
+  async connect(): Promise<{ success: boolean }> {
+    return {
+      success: true,
+    };
+  }
+
+  async get(fileId: DriveFile['fileId']) {
+    const match = await this.repository.findOneBy({ fileId });
+    return {
+      success: true,
```
sg-gs, line 64: Why are we returning success: true if we are not catching to return succes: false? Also, we should add the return type here, as the get will or will not found the desired `file`. I suggest using `uuid` as the key to identify files as the autoincremental id will disappear in the future and the `fileId` is a field from the network which is not directly related to Drive and therefore, could change in the future.

src/main/database/entities/DriveFile.ts
```diff
@@ -0,0 +1,99 @@
...
+      result: match,
+    };
+  }
+
+  async update(fileId: DriveFile['fileId'], updatePayload: Partial<DriveFile>) {
+    const match = await this.repository.update(
+      {
+        fileId,
```
sg-gs, line 72: Same here, better update by `uuid`, `fileId` is susceptible to change

src/main/database/entities/DriveFile.ts
```diff
@@ -0,0 +1,99 @@
...
+    const createResult = await this.repository.save(creationPayload);
+
+    return {
+      success: createResult ? true : false,
+      result: createResult,
+    };
+  }
+
+  async remove(fileId: DriveFile['fileId']) {
+    const result = await this.repository.delete({ fileId });
```
sg-gs, line 93: Same here

src/main/main.ts
```diff
@@ -93,6 +96,7 @@ ipcMain.on('user-quit', () => {
 app
   .whenReady()
   .then(async () => {
+    await AppDataSource.initialize();
```
sg-gs, line 99: Should we retry or add something like an smoke test to ensure that if it fails the first time, we try to recover from it?

src/main/remote-sync/helpers.ts
```diff
@@ -0,0 +1,85 @@
...
+    'lastFilesSyncAt',
+    new Date(date.getTime() - offsetMs).toISOString()
+  );
+  return date;
+}
+
+export function getLastFoldersSyncAt(): Date | undefined {
+  const value = remoteSyncStore.get('lastFoldersSyncAt');
+
+  if (!value) return undefined;
```
sg-gs, line 28: May we return the new Date with the beggining of the epoch to isolate here all the logic of returning a Date depending on if exists or not?

src/main/remote-sync/RemoteSyncManager.ts
sg-gs: I think @JoanVicens is referring to injecting the `EventEmitter` via constructor @PixoDev

sg-gs (changes requested):

src/main/remote-sync/helpers.ts
PixoDev: I think we should not here, since this basically gives you a date, or doesn't give you a date (which means no sync has ever happened), if we are going to return always a date, we won't be able to determine if there was a previous sync at some Date

src/main/database/entities/DriveFile.ts
PixoDev: types are inferred from the base abstract class, and are modified using a TS generic, see here: https://github.com/internxt/drive-desktop/blob/4edc8ce9ad4c31f299a5e7bdf11cd11372eedaad/src/main/database/adapters/base.ts

I moved the primary key to uuid.

About the success true, this operation cannot fail (there's a method called findOneByOrFail that does throw if no entry is found, is not being used here), so this basically means that the operation succeded, but maybe no entry was found in the db, which would result in `result:null` 

I added the success field to unify those adapters that maybe under some circumstances could lead to failed operations, but is not related to the result of the operation itself

src/main/main.ts
PixoDev: I find smoke testing useful when you can recover from the smoke test failing case. This is not the case, if we fail to run this promise, we are sold and the app won't work properly, so I think a retry or an smoke test won't help us here. This method initializes your data source, runs schema migrations if needed and prepares the db file in case it doesn't exists

src/main/remote-sync/RemoteSyncManager.ts
PixoDev: ok, but why? I added a simple callback pattern here (now with arrays we can handle N callbacks) because this is the only method that uses callbacks in this class, maybe in the future we add more stuff that needs a more sophisticated system, but I believe the callbacks pattern gets the job done here

sg-gs (approved):

sg-gs:
LGTM. Is someone else @JoanVicens?

src/main/database/entities/DriveFile.ts
```diff
@@ -0,0 +1,99 @@
...
+  folderId!: number;
+
+  @Column({ nullable: true })
+  folderUuid?: string;
+
+  @Column({ nullable: false })
+  userId!: number;
+
+  @Column({ nullable: true })
+  modificationTime?: string;
```
JoanVicens, line 38: `modificationTime` is not deprecated?

PixoDev: I believe is the same as updatedAt since the dates I receive are the same, can you confirm it @sg-gs ?

JoanVicens:
Would be neat to store also the path for the files and folders, for example:
- fodler A
 - file.txt
 
On the entry of file.txt the path would be folder A/file.txt

This way when we need to search for the path it's already calculated.

src/main/database/entities/DriveFile.ts
PixoDev: I confirm that updatedAt and modificationTime are different fields, modificationTime is not deprecated

src/main/database/adapters/base.ts
```diff
@@ -0,0 +1,39 @@
+export abstract class DatabaseCollectionAdapter<DatabaseItemType> {
```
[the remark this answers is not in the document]
PixoDev: I like the idea, but what if the file is in A/FOLDER/FOLDER_NESTED/A.jpg we will need to go back recursively until we build the path right?

JoanVicens (approved):

lalberts:
Hello,
may the wrong place but:
can I change the path of the local sync drive?  
Internxt-Drive is more than 35GB big locally? this is not what I call a virtual drive?
Thanks
Lukas

PR #241, opened 2023-05-19
[PB-324]:(feature) Populate listing using local db

PixoDev:
Populate listing using the RemoteSyncManager and the local db entities, so we avoid config.json serialization


JoanVicens (approved):

src/workers/filesystems/infrastructure/SyncRemoteFileSystem.ts
```diff
@@ -46,625 +50,747 @@ type CacheData = {
...
+            size: driveFile.size,
+            type: driveFile.type,
+            updatedAt: driveFile.updatedAt,
+            userId: driveFile.userId,
+          };
+          return serverFile;
+        }
+      );
+
+      const serverFolders: ServerFolder[] = allDriveFolders.map<ServerFolder>(
```
JoanVicens, line 127: The map can be expressed like: 
allDriveFolders.map<ServerFolder>((folder) => ({bucket: folder.bucket ....})

And there is no need to declare a variable and return it

JoanVicens (approved):

PR #243, opened 2023-05-22
[PB-436] feat: handle main process errors

JoanVicens:
Added unhandled listener on the main process. When an unhandled error the app will close. Fixed the following problems that appeared on a local Linux build:
- Change import from `@Sentry/electron` to `@Sentry/electron/main`
- Downgrade the electron version due to the on-click event on the tray icon not working
- Change the jest configuration to run on `node` environment (Sentry checks if the imported module it is on the correct environment)
- Mocked the sentry calls on `RemoteSyncManager` to avoid internal Sentry error

### DIFF
--- a/package.json
+++ b/package.json
@@ -194,7 +194,7 @@
     "detect-port": "^1.3.0",
     "dotenv": "^10.0.0",
     "dotenv-webpack": "^7.0.3",
-    "electron": "23.0.0",
+    "electron": "^21.0.0",
     "electron-builder": "^23.0.3",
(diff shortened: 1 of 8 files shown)


PixoDev:
is this still relevant? We fixed a lot of main process bugs/logout bugs, also this PR downgrades the Electron version, can you explain why? are we sure we are not going to break extra things with the downgrade? @JoanVicens

JoanVicens:
I dont' think so

PR #245, opened 2023-05-29
[_] fix: errors on logout

JoanVicens:


src/main/usage/service.ts
```diff
@@ -1,76 +1,77 @@
...
+  const [driveUsage, photosUsage, limitInBytes] = await Promise.all([
+    getDriveUsage(),
+    getPhotosUsage(),
+    getLimit(),
+  ]);
+
+  return {
+    usageInBytes: driveUsage + photosUsage,
+    limitInBytes,
+    isInfinite: limitInBytes >= 108851651149824,
```
PixoDev, line 74: What is this hardcoded value?

src/main/usage/service.ts
```diff
@@ -1,76 +1,77 @@
...
-async function getLimit(headers: HeadersInit): Promise<number> {
-	const response = await fetch(`${driveUrl}/api/limit`, {
-		headers,
-	});
-
-	if (!response.ok) {
-		Logger.error('Limit request failed');
-	}
+async function getLimit(): Promise<number> {
+  const response = await httpClient.get(`${driveUrl}/api/limit`);
```
PixoDev, line 52: I think is worth the time to start moving this raw http methods to use the SDK ones, this method is already exposed in the SDK. 

If we use the SDK ones, we don't need to maintain this if the endpoint implementation changes, also we have TS signatures out of the box.

Get Drive Limit endpoint:
https://github.com/internxt/sdk/blob/c164a03d4778786046e4818be261222e7c438f21/src/drive/storage/index.ts#L253

src/main/usage/service.ts
```diff
@@ -1,76 +1,77 @@
...
-async function getDriveUsage(headers: HeadersInit): Promise<number> {
-	const response = await fetch(`${driveUrl}/api/usage`, {
-		headers,
-	});
-
-	if (!response.ok) {
-		Logger.error('Usage request failed');
-	}
+async function getDriveUsage(): Promise<number> {
+  const r = await httpClient.get(`${driveUrl}/api/usage`);
```
PixoDev, line 39: This also exists in the SDK, maybe worth the refactor time

Get Drive Usage:
https://github.com/internxt/sdk/blob/c164a03d4778786046e4818be261222e7c438f21/src/drive/storage/index.ts#L246

src/main/preload.d.ts
```diff
@@ -112,7 +118,7 @@ declare interface Window {
     getBackups: typeof import('../main/device/service').getBackupsFromDevice;
 
     addBackup: typeof import('../main/device/service').addBackup;
-    addBackupsFromLocalPaths: typeof import('../main/device/service').createBackupsFromLocalPaths;
```
PixoDev, line 115: This is needed for the onboarding, why is being removed here? It will make the onboarding crash

https://github.com/internxt/drive-desktop/blob/4ef2cd07563f917a9c80dadf2e3c032fe6ffa974/src/renderer/pages/Onboarding/index.tsx#L27

JoanVicens: Probably this was because of the linter. I do not remember deleting those lines, either is necessary for the task. Will undo them

src/main/preload.d.ts
```diff
@@ -153,10 +154,10 @@ declare interface Window {
     ): () => void;
 
     changeBackupPath: typeof import('../main/device/service').changeBackupPath;
-    getFolderPath: typeof import('../main/device/service').getPathFromDialog;
```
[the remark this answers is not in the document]
JoanVicens: Don't know why it's marked as deleted. It is not for my local

src/main/preload.d.ts
JoanVicens: Don't know why it's marked as deleted. It is not for my local

PixoDev (approved):

PR #247, opened 2023-05-30
[_]: Fix codebase formatting

PixoDev:
Formatted the whole project with eslint rules (included prettier ones) so we avoid more eslint changes during code reviews

 Note that eslint itselfs formats the code, prettier is not needed, eslint formatting runs on precommit hook


JoanVicens:
Seems good to me

larry-internxt (approved):

PR #251, opened 2023-06-02
[PB-403]: VirtualDrive improvements

larry-internxt:
- Fixed and improved windows mount and unmount (virtual drive letter related)
- Fixed macos/linux mount
- Added useful functionality


src/main/config.ts
```diff
@@ -112,6 +116,7 @@ export const defaults: AppStore = {
   backupList: {},
   clientId: uuid.v4(),
   preferedLanguage: '',
+  virtualdriveWindowsLetter: 'I',
```
JoanVicens, line 119: I could be totally wrong. But I think that when the user already have a config.json file with the stored values if you add a new entry it's default value will not be added and will return undefined until is `set`

JoanVicens (changes requested):

src/main/config.ts
larry-internxt: I chequed and it worked good, anyways I've added a default value just in case

src/main/main.ts
```diff
@@ -89,6 +89,7 @@ const installExtensions = async () => {
 };
 
 app.on('window-all-closed', () => {
+  unmountDrive();
```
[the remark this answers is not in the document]
larry-internxt: i've leaved only de will-quit event for better performance

PR #253, opened 2023-06-02
[PB-497] feat: add storage manager

JoanVicens:
Based on the weekly branch.

Adds the storage manager to the server


JoanVicens:
@sg-gs Even I'm returing 3GB (3221225472 to be extacts) as avaliable space I it displays 3.2GB
Tried to return 0 and the returns 0 so doesn't seem to be doing any calculation.
Do you have any idea why can this happen?

image

larry-internxt (approved):

PR #261, opened 2023-06-12
fix: use gio to mount to nautilus

JoanVicens:
Changes the mount on Linux systems to gio (GNOME Input/Output) since Ubuntu 22.04.1 LTS `mount` is not working.

The culprit is that to mount a `webdav` type of file system server davfs2 need to be installed. And on 22.04.1 LTS does not come by default.

The fix mounts the server to the file explorer and creates a symlink to the "mounted" folder

### DIFF
--- a/src/workers/webdav/VirtualDrive.ts
+++ b/src/workers/webdav/VirtualDrive.ts
@@ -1,5 +1,5 @@
 import Logger from 'electron-log';
-import { exec } from 'child_process';
+import { exec, execSync } from 'child_process';
 import configStore from '../../main/config';
 import { homedir } from 'os';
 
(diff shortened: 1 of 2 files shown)


src/workers/webdav/VirtualDrive.ts
```diff
@@ -1,5 +1,5 @@
 import Logger from 'electron-log';
-import { exec } from 'child_process';
+import { exec, execSync } from 'child_process';
```
larry-internxt, line 2: i think this new import is not being used

JoanVicens: yep, solved

PR #262, opened 2023-06-12
[PB-424]: (feature) New migration flow UI

PixoDev:
Migration flow UI, the system that performs the actual migration is not included in this PR, a fake migration for demo purposes has been added here.

Translations for EN, ES and FR are included


src/main/device/service.ts
```diff
@@ -8,6 +8,7 @@ import path from 'path';
 import { getHeaders } from '../auth/service';
 import { addGeneralIssue } from '../background-processes/process-issues';
 import configStore from '../config';
+import { getVirtualDrivePath } from 'workers/webdav/VirtualDrive';
```
larry-internxt, line 11: i think this new imports are not being used

src/main/sync-root-folder/handlers.ts
```diff
@@ -9,8 +9,10 @@ ipcMain.handle('get-sync-root', () => {
...
 ipcMain.handle('choose-sync-root-with-dialog', chooseSyncRootWithDialog);
 
-ipcMain.handle('open-sync-folder', () => {
+ipcMain.handle('open-sync-folder', async () => {
   const syncFolderPath = configStore.get('syncRoot');
 
-  return shell.openPath(syncFolderPath);
+  const errorMessage = await shell.openPath(syncFolderPath);
+
+  if (errorMessage) throw new Error(errorMessage);
```
JoanVicens, line 17: We cannot throw errors on the handlers https://github.com/electron/electron/issues/24427

src/renderer/pages/Onboarding/index.tsx
```diff
@@ -13,6 +13,11 @@ export default function Onboarding() {
   const [backupFolders, setBackupFolders] = useState<BackupFolder[]>([]);
   const [slideIndex, setSlideIndex] = useState<number>(0);
   const [backupsModalOpen, setBackupsModalOpen] = useState(false);
+  const [platform, setPlatform] = useState<string>('');
```
JoanVicens, line 16: ℹ️  There is a hook to get the platform `useClientPlatform`

src/renderer/pages/Onboarding/slides/AvailableOnlineSlide.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
-import { OnboardingSlideProps } from '../helpers';
+
 import { useTranslationContext } from 'renderer/context/LocalContext';
 
-export type AvailableOnlineSlideProps = OnboardingSlideProps;
+export type AvailableOnlineSlideProps = object;
```
JoanVicens, line 5: If the props are of type `object` do we need them? I guess this is for the demo

src/renderer/pages/Onboarding/index.tsx
PixoDev: glorious, thx didn't see it

src/renderer/pages/Onboarding/slides/AvailableOnlineSlide.tsx
PixoDev: I think will be better to not use props here yeah

PR #263, opened 2023-06-13
[PB-577] fix: don't hide the login window when it loses focus

JoanVicens:
The login popup window stays open on the desktop after losing focus


larry-internxt (approved):
seems good for me

PR #265, opened 2023-06-14
[_]:(enhancement) Remove desktop dead code for V2

PixoDev:
This PR removes a lot of dead code due to new systems introduced on desktop codebase

- The sync worker has been removed since we have now the RemoteSyncManager and the webdav server implementation
- Sync related stuff like listings and actions are removed too, since we don't use them anymore
- The lock service is no longer used since now the RemoteSyncManager takes care of maintaining local and remote updated

Tests are passing, and the app works as expected, but some tests are appreciated since this removes a lot of code


larry-internxt (approved):
🎉

JoanVicens (approved):

PR #266, opened 2023-06-14
[PB-553]: Fix double folder opened on folder icon clicking

larry-internxt:
- Fixed double folder opened on folder icon clicking (windows)

### DIFF
--- a/src/renderer/pages/Widget/Header.tsx
+++ b/src/renderer/pages/Widget/Header.tsx
@@ -149,8 +149,6 @@ export default function Header() {
       <div
         role="button"
         tabIndex={0}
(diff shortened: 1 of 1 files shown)


src/renderer/pages/Widget/Header.tsx
```diff
@@ -149,8 +149,6 @@ export default function Header() {
       <div
         role="button"
         tabIndex={0}
-        onClick={window.electron.openSyncFolder}
-        onKeyPress={window.electron.openSyncFolder}
```
PixoDev, line 153: I would say is better to remove the onClick from the child since this html item is the one with role="button"

PixoDev (approved):

PR #268, opened 2023-06-14
[PB-496]:(feature) Generate tree for webdav using local db

PixoDev:
Retrieve files and folders that exists on remote using the local db every time getTree is called


JoanVicens:
Just a nitpicks that I did not catch on before, the IPC on the Repositories should be injected.

JoanVicens (approved):

JoanVicens (approved):

PR #269, opened 2023-06-14
[PB-480] Feat: Track times

JoanVicens:
Changes the repository approach to have a class that handles every action. The repository is still there but can be removed after (not included in this PR to not block other PRs that need this)


src/main/analytics/webdav-handlers.ts
```diff
@@ -1,38 +1,92 @@
...
+      size,
+      elapsedTime: uploadInfo.elapsedTime,
+    });
+  });
+}
+
+function subscribeToFlowErrors(ipc: IpcWebdavFlowErrors) {
+  ipc.on('WEBDAV_FILE_UPLOADED_ERROR', (_, payload) => {
+    const { name, error } = payload;
+    trackWebdavError('Upload Error', { name, error });
```
PixoDev, line 84: maybe add `reportError` call too so we have the data available on Sentry?

src/workers/webdav/modules/files/test/infrastructure/storage/EnvironmentFileContentRepository.test.ts
```diff
@@ -40,7 +40,7 @@ describe.skip('Environment File Content Repository', () => {
         cbs.finishedCallback(undefined, new Readable());
       });
 
-      const stream = await reposiotry.download(file);
+      const stream = await reposiotry.downloader(file).download();
```
PixoDev, line 43: repository?

src/main/analytics/webdav-handlers.ts
JoanVicens: The report to Sentry is made on the Webdav worker, the analytics need to be on the main process as it relies on a NodeJs process

src/workers/webdav/modules/files/test/infrastructure/storage/EnvironmentFileContentRepository.test.ts
JoanVicens: fixex

src/main/analytics/service.ts
```diff
@@ -273,6 +273,19 @@ export function trackWebdavEvent(
 }
 
 export function trackWebdavError(
+  event: TrackedWebdavServerErrorEvents,
+  properties: Record<string, any>
+) {
+  const { uuid: userId } = ConfigStore.get('userData');
+  client.track({
+    userId,
```
PixoDev, line 281: Can you check this? I believe something got ignored during the rebase, this will overwrite a fix I pushed some days ago, here is already in master-erb https://github.com/internxt/drive-desktop/blob/master-erb/src/main/analytics/service.ts#L290

Here is the PR https://github.com/internxt/drive-desktop/pull/279

PixoDev (changes requested):

PixoDev (approved):

PR #270, opened 2023-06-15
[PB-586]: Fix download & upload notifications

larry-internxt:
- Fix status should not say Downloading infinity% by changing the way upload and download notifications are triggered


src/workers/webdav/modules/files/domain/RemoteFileContentsRepository.ts
```diff
@@ -7,5 +7,5 @@ export interface RemoteFileContentsRepository {
 
   download(file: WebdavFile): Promise<Readable>;
 
-  upload(size: FileSize, contents: Readable): Promise<WebdavFile['fileId']>;
+  upload(name: string, size: FileSize, contents: Readable): Promise<WebdavFile['fileId']>;
```
JoanVicens, line 10: If the interface changes, when it's used should change also

JoanVicens (commented):
When a file is cloned and event should be sended

src/workers/webdav/modules/files/domain/RemoteFileContentsRepository.ts
JoanVicens: Also, we should avoid passing the name only to add it to the event. If we need to add more info on the event we will have to modify the contract. For the tracking on ruddestack we will need the size and type of the file for example

PR #272, opened 2023-06-19
[PB-498] feat: force upade usage when selecting the account Section

JoanVicens:


src/renderer/hooks/useUsage.tsx
```diff
@@ -14,20 +13,23 @@ export default function useUsage() {
...
+    setUsage('loading');
     updateUsage();
     const listener = window.electron.onRemoteChanges(updateUsage);
-
     return listener;
-  }, []);
+  }, [refresh]);
+
+  const refreshUsage = () => {
+    setRefresh(Date.now());
```
PixoDev, line 31: why don't call `updateUsage` here directly?

PixoDev (approved):

PR #274, opened 2023-06-19
[PB-506]:(feature) Send feedback window

PixoDev:
Allow users to send feedback via feedback window, can be accesed from widget dropdown menu.

Also HttpClient error interceptor has been modified to throw the received error correctly


src/main/remote-sync/RemoteSyncManager.ts
```diff
@@ -294,17 +308,28 @@ export class RemoteSyncManager {
...
+          2
+        )}`
+      );
+
+      throw new Error('Did not receive an array of files');
+    }
+
     const hasMore =
-      isArray(response.data) &&
+      Array.isArray(response.data) &&
```
JoanVicens, line 326: This check is already made a few lines above. If it needed to have `response.data ` typed the the check can be converted into a guard clause

PixoDev: fixed

JoanVicens (approved):

PR #275, opened 2023-06-20
[_]:(fix) Connect correctly to notifications websocket

PixoDev:
Fix how we connect to notifications websocket, tested on local

- Use websocket transport instead of polling
- Use newToken instead of bearerToken
- Remove polling code (it was not working either)


JoanVicens:
I'm not seeing the new token change you mention.
Also, not sure about the transport change I think I talked about it with @sg-gs but not remember if we wanted to avoid the WebSocket

src/main/realtime.ts
```diff
@@ -12,42 +12,23 @@ function cleanAndStartRemoteNotifications() {
...
   socket = io(process.env.NOTIFICATIONS_URL, {
+    transports: ['websocket', 'polling'],
     auth: {
       token: obtainToken('bearerToken'),
     },
     withCredentials: true,
   });
 
-  socket.io.on('open', () => {
-    socket?.io.engine.transport.on('pollComplete', () => {
```
sg-gs, line 22: Setting the cookie is required for the load balancer of any provider to be able to connect always against the same replica which holds the websocket

sg-gs (changes requested):

src/main/realtime.ts
PixoDev: I believe this was for sticky sessions, but the polling transport was not connecting either, if we use the websocket transport, which is working, the connection is always open right?

PixoDev: I restored the ingress cookie

PixoDev:
So after investigating for a bit why the polling is failing, I noticed that the handler in the notifications server is waiting for a X-API-KEY header, since we are not sending it, because that endpoint is used for requesting notifications to send messages, and not to receive the messages itself, polling won't work right now:

https://github.com/internxt/notifications/blob/master/src/auth-api-middleware.ts

This is what happens when I try to connect to my local notifications from desktop, even with Postman

https://gyazo.com/4bc7aad0dd6d42455adf3fa27b702665

Even setting the X-API-KEY header in the request, it will fail because we are missing body (we get a 400), because we cannot connect to that endpoint to get notifications from, so right now only websocket transport will work

src/main/remote-sync/handlers.ts
```diff
@@ -54,6 +55,10 @@ ipcMain.handle('START_REMOTE_SYNC', async () => {
 });
 
 eventBus.on('RECEIVED_REMOTE_CHANGES', async () => {
+  // Wait before checking for updates, could be possible
+  // that we received the notification, but if we check
+  // for new data we don't receive it
+  await sleep(500);
```
sg-gs, line 61: Why 500ms?

src/main/realtime.ts
```diff
@@ -12,16 +15,22 @@ function cleanAndStartRemoteNotifications() {
...
   socket = io(process.env.NOTIFICATIONS_URL, {
+    transports: ['websocket', 'polling'],
     auth: {
       token: obtainToken('bearerToken'),
     },
     withCredentials: true,
   });
 
-  socket.io.on('open', () => {
+  socket.on('open', () => {
```
sg-gs, line 25: Why is now directly listening from the socket and not from io?

src/main/realtime.ts
```diff
@@ -12,16 +15,22 @@ function cleanAndStartRemoteNotifications() {
   stopRemoteNotifications();
 
   socket = io(process.env.NOTIFICATIONS_URL, {
+    transports: ['websocket', 'polling'],
```
sg-gs, line 18: If we are not using polling, remove it from the allowed methods. If not, it will fallback to polling, which we do know in advance is not going to work

sg-gs (changes requested):

src/main/realtime.ts
PixoDev: I believe this was incorrect before, socket.io is a readonly property with underlying logic (they call it a Manager, it holds all the connection logic), socket is a class that extends from EventEmitter, idk if they are the same, but since the rest of the methods uses `socket.on` instead of `socket.io.on` I decided to move it to `socket.on`

src/main/remote-sync/handlers.ts
sg-gs: Okay

sg-gs (approved):

PR #279, opened 2023-06-21
[_]:(fix) Fix webdav events errors when logged out

PixoDev:
Fixed some cases where we attempt to track an event, pulling the user from the config store, but no user is found, so this error pops up
08d9c818-f1cc-41b3-8f4c-fb13fb34b0a9

### DIFF
--- a/src/main/analytics/service.ts
+++ b/src/main/analytics/service.ts
@@ -257,7 +257,11 @@ export function trackWebdavEvent(
   event: TrackedWebdavServerEvents,
   properties: Record<string, any>
 ) {
-  const { uuid: userId } = ConfigStore.get('userData');
+  const userData = ConfigStore.get('userData');
+
+  if (!userData) return;
+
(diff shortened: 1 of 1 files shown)


PixoDev:
Should we modify the TS signature for ConfigStore.get('userData')? Because it can return nully values, and this can happen @JoanVicens

JoanVicens:
This error seems to have occurred because the Webdav server threw an error after the user logged out. I don't think we should avoid throwing an error in that case (it should be captured tho) since the server should not be running if the user logged out or is not even logged in

PixoDev:
This is going to fail if we try to access the userData object. I think for now, I'll check if we have an user, if not, we can send an anonymousId generated with uuid.v4(), wdyt? @JoanVicens

JoanVicens:
The anonymousId has to be unique, we are using the clientId from the storage

JoanVicens (approved):

PR #282, opened 2023-06-22
[PB-633]:(feature) Display status of the remote sync, display error if it fails

PixoDev:
- Display remote sync status in the widget
- Display error message if remote sync fails

### DIFF
--- a/src/renderer/pages/Widget/index.tsx
+++ b/src/renderer/pages/Widget/index.tsx
@@ -3,15 +3,30 @@ import Header from './Header';
 import SyncAction from './SyncAction';
 import SyncErrorBanner from './SyncErrorBanner';
 import SyncInfo from './SyncInfo';
+import useSyncStatus from '../../hooks/SyncStatus';
+import { useState } from 'react';
+import { SyncStatus } from '../../../main/background-processes/sync';
+import { SyncFailed } from './SyncFailed';
 
 export default function Widget() {
+  const [syncStatus, setSyncStatus] = useState<SyncStatus>('RUNNING');
+  useSyncStatus(setSyncStatus);
+  const handleRetrySync = () => {
+    window.electron.startRemoteSync().catch((err) => {
+      reportError(err);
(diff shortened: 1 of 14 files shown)


src/renderer/pages/Widget/index.tsx
```diff
@@ -3,15 +3,30 @@ import Header from './Header';
...
 import SyncInfo from './SyncInfo';
+import useSyncStatus from '../../hooks/SyncStatus';
+import { useState } from 'react';
+import { SyncStatus } from '../../../main/background-processes/sync';
+import { SyncFailed } from './SyncFailed';
 
 export default function Widget() {
+  const [syncStatus, setSyncStatus] = useState<SyncStatus>('RUNNING');
+  useSyncStatus(setSyncStatus);
+  const handleRetrySync = () => {
```
JoanVicens, line 14: This could be declared on `SyncFailed`, there is no need to pass it as a prop from here

PixoDev: note that props are distributed to two childs, SyncFailed and SyncAction component, is just easier to maintain the logic in the parent and distribute it to the childs than distributing logic between childs just to avoid props drilling. 

Now if I need to modify or improve this logic I have everything centralized in this component, the other ones are just dummy components. Do you think it has any benefit to distribute the logic through multiple components in a simple component tree like this one?

JoanVicens: but handleRetrySync is only passed to SyncFailed or I'm missing something?

JoanVicens:
Could not get the status to display RUNNING after uploading a file on remote, but not seeing an obvious cause

JoanVicens (approved):

PixoDev:
> Could not get the status to display RUNNING after uploading a file on remote, but not seeing an obvious cause

This is because notifications is not working correctly so you are not receiving events that triggers a sync again, theres an open PR for that

PR #288, opened 2023-06-26
[PB-669]:(fix) Allow to close feedback window

PixoDev:
Allow to close feedback window

### DIFF
--- a/src/main/windows/feedback.ts
+++ b/src/main/windows/feedback.ts
@@ -22,7 +22,6 @@ export const openFeedbackWindow = () => {
       nodeIntegration: true,
     },
     titleBarStyle: process.platform === 'darwin' ? 'hidden' : undefined,
-    frame: process.platform !== 'darwin' ? false : undefined,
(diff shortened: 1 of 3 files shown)


PixoDev:
@JoanVicens Can you check this on windows? Seems like the feedback window cannot be closed on win32 platform, I added the closable: true property, but the docs says the default is true, can you verify if it can be closed on win32 now? I don't have one to test this on

JoanVicens:
We are not supporting win32 and will not support it in the future either as discussed [here](https://inxt.atlassian.net/browse/PB-325).

To show a top bar with a close button you can use `WindowTopBar` component

PixoDev:
> We are not supporting win32 and will not support it in the future either as discussed [here](https://inxt.atlassian.net/browse/PB-325).
> 
> To show a top bar with a close button you can use `WindowTopBar` component

I added the WindowTopBar component

JoanVicens (approved):

PR #290, opened 2023-06-26
[_]: Added virtual drive unmount error handler

larry-internxt:
- Added virtual drive unmount error handler

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -40,6 +40,8 @@ import packageJson from '../../package.json';
 import eventBus from './event-bus';
 import * as Sentry from '@sentry/electron/main';
 import { AppDataSource } from './database/data-source';
+import { reportError } from './bug-report/service';
+
 Logger.log(`Running ${packageJson.version}`);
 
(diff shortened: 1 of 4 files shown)


src/main/main.ts
```diff
@@ -91,11 +91,11 @@ const installExtensions = async () => {
 };
 
 app.on('will-quit', () => {
-  unmountDrive();
+  unmountDrive().catch(Logger.error);
```
PixoDev, line 94: can you also call reportError() to report this to Sentry?, this way we can see how many users are getting errors here

src/main/main.ts
```diff
@@ -91,11 +91,11 @@ const installExtensions = async () => {
...
 });
 
 app.on('window-all-closed', () => {
-  unmountDrive();
+  unmountDrive().catch(Logger.error);
```
PixoDev, line 98: Same here, add reportError call too please

PixoDev (changes requested):

PixoDev (approved):

PR #291, opened 2023-06-26
[PB-632]:(fix) Separate widget and login windows, improve widget cleanup

PixoDev:
- Fix widget cleanup logic
- Improve windows cleanup when closed
- Separate widget and login windows

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -40,6 +40,16 @@ import packageJson from '../../package.json';
 import eventBus from './event-bus';
 import * as Sentry from '@sentry/electron/main';
 import { AppDataSource } from './database/data-source';
+import { getIsLoggedIn } from './auth/handlers';
+import {
+  createWidget,
+  getWidget,
+  setBoundsOfWidgetByPath,
+} from './windows/widget';
+import { createAuthWindow, getAuthWindow } from './windows/auth';
+import configStore from './config';
+import { getTray } from './tray';
+import { openOnboardingWindow } from './windows/onboarding';
 Logger.log(`Running ${packageJson.version}`);
 
 Logger.log('Initializing Sentry for main process');
@@ -106,7 +116,15 @@ ipcMain.on('user-quit', () => {
 app
   .whenReady()
   .then(async () => {
+    if (!AppDataSource.isInitialized) {
+      await AppDataSource.initialize();
+    }
     eventBus.emit('APP_IS_READY');
+    const isLoggedIn = getIsLoggedIn();
+
+    if (!isLoggedIn) {
+      await createAuthWindow();
+    }
     if (process.env.NODE_ENV === 'development') {
       await installExtensions();
     }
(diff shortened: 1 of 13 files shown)


src/main/windows/auth.ts
```diff
@@ -0,0 +1,67 @@
...
+export const createAuthWindow = async () => {
+  authWindow = new BrowserWindow({
+    width: 300,
+    height: 474,
+    show: false,
+    webPreferences: {
+      preload: preloadPath,
+      nodeIntegration: true,
+    },
+    movable: true,
```
JoanVicens, line 22: even if `movable` is set to true on the page need to have a section of it `-webkit-app-region: drag;` The `WindowTopBar` has it (the class draggable) but `Login` don't

src/main/windows/widget.ts
```diff
@@ -2,18 +2,20 @@ import { BrowserWindow, screen } from 'electron';
 
 import { isAutoLaunchEnabled } from '../auto-launch/service';
 import eventBus from '../event-bus';
-import { getTray } from '../tray';
+import { TrayMenu } from '../tray';
 import { preloadPath, resolveHtmlPath } from '../util';
 import { setUpCommonWindowHandlers } from '.';
 import { getIsLoggedIn } from '../auth/handlers';
 
+const widgetConfig: { width: number; height: number; placeUnderTray: boolean } =
```
JoanVicens, line 10: Do we want to keep the window config like before if this window will only display the widget? We could simplify some logic. Will add a comment of what I mean

src/main/windows/widget.ts
```diff
@@ -121,8 +99,19 @@ function getLocationUnderTray(
...
+  const { placeUnderTray, ...size } = widgetConfig;
 
-  eventBus.emit('WIDGET_IS_READY');
-});
+  const bounds = tray.bounds;
+
+  if (placeUnderTray && bounds) {
+    const location = getLocationUnderTray(size, bounds);
+    widgetWindow.setBounds({ ...size, ...location });
+  } else {
```
JoanVicens, line 113: This is a case we could simplify (from the comment above). I belive the window was only centered when it was displaying the login page

PixoDev: agreed, will remove this

src/main/main.ts
```diff
@@ -115,9 +133,35 @@ app
...
+  const tray = getTray();
+  if (widget && tray) {
+    setBoundsOfWidgetByPath(widget, tray);
+  }
+
+  const lastOnboardingShown = configStore.get('lastOnboardingShown');
+
+  if (!lastOnboardingShown) {
+    openOnboardingWindow();
+  } else if (widget) {
```
JoanVicens, line 151: This will be not enough to avoid displaying the widget before the onboarding as when the widget is ready it bee shown and the auto-launch is not enable it's shown

PixoDev: I removed the widget.show when widget is ready since we open it imperatively now

src/main/main.ts
```diff
@@ -106,7 +116,15 @@ ipcMain.on('user-quit', () => {
...
   .whenReady()
   .then(async () => {
+    if (!AppDataSource.isInitialized) {
+      await AppDataSource.initialize();
+    }
     eventBus.emit('APP_IS_READY');
+    const isLoggedIn = getIsLoggedIn();
+
+    if (!isLoggedIn) {
+      await createAuthWindow();
```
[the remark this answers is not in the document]
PixoDev: fixed, now is created here only

larry-internxt (approved):

src/main/windows/auth.ts
PixoDev: removed this

JoanVicens (approved):

PR #293, opened 2023-06-28
[PB-481]: Feat/UI reporting changes

larry-internxt:
- Used webdav events to send changes to sync notifications
- Added renaming logic
- Removed some old sync notification parts

### DIFF
--- a/src/main/analytics/webdav-handlers.ts
+++ b/src/main/analytics/webdav-handlers.ts
@@ -12,7 +12,7 @@ function subscribeToFlowEvents(ipc: IpcWebdavFlow) {
 
     broadcastToWindows('sync-info-update', {
       action: 'DELETED',
-      name: name + type,
+      name: name + '.' + type,
     });
 
     trackWebdavEvent('Delete', {
@@ -27,7 +27,7 @@ function subscribeToFlowEvents(ipc: IpcWebdavFlow) {
 
     broadcastToWindows('sync-info-update', {
       action: 'DOWNLOADING',
-      name: name + type,
+      name: name + '.' + type,
       progress,
     });
   });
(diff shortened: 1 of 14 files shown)


PixoDev (changes requested):
@larry-internxt no tests no party mate, unit tests work correctly on master-erb, can you give it a look?

src/main/analytics/webdav-handlers.ts
```diff
@@ -10,27 +10,52 @@ function subscribeToFlowEvents(ipc: IpcWebdavFlow) {
...
+    });
+  });
+
   ipc.on('WEBDAV_FILE_DOWNLOADED', (_, payload) => {
-    const { name, type, size, uploadInfo } = payload;
+    const { name, type, size, processInfo } = payload;
+
+    broadcastToWindows('sync-info-update', {
+      action: 'DOWNLOADED',
+      name: name + '.' + type,
```
PixoDev, line 40: what if file doesn't has type? we display 'filename.' ?

src/main/analytics/webdav-handlers.ts
```diff
@@ -48,39 +73,59 @@ function subscribeToFlowEvents(ipc: IpcWebdavFlow) {
   ipc.on('WEBDAV_FILE_RENAMED', (_, payload) => {
     const { name } = payload;
 
+    broadcastToWindows('sync-info-update', {
+      action: 'RENAMED',
+      name,
```
PixoDev, line 78: we are not missing the type here?

src/main/analytics/webdav-handlers.ts
larry-internxt: nice catch, I've added a new property nameWithExtension for a better management

src/main/analytics/webdav-handlers.ts
larry-internxt: on this one, the name comes with extension, so i've renamed it for a better understanding

PixoDev (approved):

PR #296, opened 2023-07-03
[PB-634](fix): Use webContents API to send events to the webdav worker

PixoDev:
Send event to webdav worker using webContents API instead of ipcWebdav


larry-internxt (approved):
seems good for me

PR #297, opened 2023-07-04
[PB-680](fix): Exlude native OS files

PixoDev:
Exclude files from being created mostly

### DIFF
--- a/src/workers/webdav/worker/InternxtFileSystem/InternxtFileSystem.ts
+++ b/src/workers/webdav/worker/InternxtFileSystem/InternxtFileSystem.ts
@@ -118,7 +118,7 @@ export class InternxtFileSystem extends FileSystem {
     }
 
     if (ctx.type.isFile) {
-      const isValidName = this.fileValidator.validateName(path);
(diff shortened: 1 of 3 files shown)


JoanVicens:
The validation should also be added on `FilePath.ts` to ensure all paths are valid since users could already have that file on their drive.
This way we can catch it whenever it appears and it also allows for being tested

PixoDev:
I don't know If I understood you, but updated the PR, let me know if that's what you mean @JoanVicens

src/workers/webdav/worker/InternxtFileSystem/InternxtFileSystem.ts
```diff
@@ -132,6 +137,9 @@ export class InternxtFileSystem extends FileSystem {
     }
 
     if (item.isFile()) {
+      const isValidName = this.fileValidator.validatePath(path);
```
JoanVicens, line 140: If we already have an instance of a Webdav we know that the file is correct, since the path is validated when its created

JoanVicens (approved):

PR #300, opened 2023-07-07
[_]: Reset traverser correctly

PixoDev:
Reset the traverser correctly, this was causing a lot of files and folder to being cached incorrectly


larry-internxt (approved):
very nice catch there!!!

JoanVicens (approved):

PR #306, opened 2023-07-13
[_]: Use Internxt webdav-server

PixoDev:
Use @internxt/webdav-server from https://github.com/internxt/webdav-server

### DIFF
--- a/package.json
+++ b/package.json
@@ -260,6 +260,7 @@
     "@internxt/inxt-js": "^2.0.7",
     "@internxt/lib": "^1.1.6",
     "@internxt/sdk": "^1.4.33",
+    "@internxt/webdav-server": "^1.0.0",
     "@sentry/electron": "^4.5.0",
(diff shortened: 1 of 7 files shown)


JoanVicens (approved):

larry-internxt (approved):

claell:
@PixoDev @larry-internxt @JoanVicens Sorry to misuse this place to ask a question. Unfortunately, this project doesn't offer issues or discussions.

The Internxt support has promised WebDAV or comparable features for quite some time. Is this something related to it? An if yes, is this sensible for end users? The expected use case would be to be able to use WebDAV without installing any software on our computers. But that might clash with Internxt encryption, I fear?

Also, I saw https://api.internxt.com/drive/. Might that be a way to improve the current situation? Possibly, rclone support could be built up on that API, drastically reducing the need to have WebDAV or other open protocols available and really increasing the usability and attractiveness of Internxt for many users.

PR #308, opened 2023-07-14
[_]: Use optimistic folders and files

PixoDev:
Use optimistic folders and files when doing webdav operations


src/main/main.ts
```diff
@@ -110,10 +110,6 @@ app.on('will-quit', () => {
 });
 
 app.on('window-all-closed', () => {
-  unmountDrive().catch((error) => {
```
larry-internxt, line 113: needed for windows unmount

larry-internxt (approved):

PR #312, opened 2023-07-17
[PB-717]: Eject installer disks on app start

larry-internxt:
- Moved the ejection of the MacOS installer disks to the 'APP_IS_READY' event

### DIFF
--- a/src/main/background-processes/webdav.ts
+++ b/src/main/background-processes/webdav.ts
@@ -61,7 +61,9 @@ eventBus.on('USER_LOGGED_IN', () => {
   }
 });
 
-eventBus.on('APP_IS_READY', ejectMacOSInstallerDisks);
+if (process.platform === 'darwin') {
+  eventBus.on('APP_IS_READY', ejectMacOSInstallerDisks);
+}
 
 ipcMain.handle('retry-virtual-drive-mount', () => {
   webdavWorker?.webContents.send('RETRY_VIRTUAL_DRIVE_MOUNT');
--- a/src/main/background-processes/webdav.ts
+++ b/src/main/background-processes/webdav.ts
@@ -3,6 +3,7 @@ import { ipcWebdav } from '../ipcs/webdav';
 import path from 'path';
 import Logger from 'electron-log';
 import eventBus from '../event-bus';
+import { ejectMacOSInstallerDisks } from '../../workers/webdav/VirtualDrive';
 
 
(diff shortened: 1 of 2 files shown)


src/main/background-processes/webdav.ts
```diff
@@ -60,6 +61,8 @@ eventBus.on('USER_LOGGED_IN', () => {
   }
 });
 
+eventBus.on('APP_IS_READY', ejectMacOSInstallerDisks);
```
PixoDev, line 64: Worth to add a .catch to the promise, so if this fails we:

A) Handle the error correctly
B) Avoid unhandled exceptions in case it fails, exceptions thrown from an eventEmitter listener are hard to debug
C) Can report the error + context to Sentry if we want

Also I think we should register this listener on MacOS only, so we don't have a NOOP listening on Windows + Linux

PixoDev (changes requested):

src/workers/webdav/VirtualDrive.ts
```diff
@@ -328,10 +327,10 @@ const getMacOSMountedInstallerDisks = (): Promise<string[]> => {
   });
 };
 
-const ejectMacOSInstallerDisks = async (): Promise<void> => {
+export const ejectMacOSInstallerDisks = async (): Promise<void> => {
   const installerDisks = await getMacOSMountedInstallerDisks();
   installerDisks.forEach((installerDisk) => {
-    ejectMacOSDisk(installerDisk);
+    ejectMacOSDisk(installerDisk).catch();
```
PixoDev, line 333: Don't do this please, we are telling the error to shut up, if this fails, it will fail silently, at least log the error. Anyways, isn't better if we fail this operation in case some ejection fails? something like this:
```

for(const installerDisk of installerDisks) {
   await ejectMacOSDisk(installerDisk)
}
```

This way if something crash here, we propagate the error, and we stop the sequence

PixoDev (changes requested):

src/workers/webdav/VirtualDrive.ts
larry-internxt: hey, I'm doing it like this because I think it's better to not stop the execution, and keep trying to eject the other disks. Even if none of them are ejected, I think its not something critical enough to propagate the error or stop the execution.

PixoDev (approved):

PR #316, opened 2023-07-20
[_] feat: improve traverser performance

JoanVicens:
Changed the raw tree arrays to Sets, Sets have O(1) time complexity for adding and deleting elements, so we can delete items already added to the collection without cost. Since we loop through the raw files in every folder, we are reducing the length of files being looped over and over. 

https://stackoverflow.com/a/75514478
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

Also changed the file and folder data "verification" in one single loop


sg-gs:
Do you have a benchmark to establish a comparison between the current version and the proposed update? @JoanVicens

JoanVicens:
@sg-gs I did not benchmark it, on the first link there are comments with benchmarks for Set vs Array operations.

I did not describe the main point of this pr clearly. The main chain is not the change for the Set but the refactor to use one single loop for traversing the files and one for the folders. This way we avoid looping multiple times for all the items

sg-gs:
Okay, from the complexity standpoint seems good. Although, it would be beneficial to see the impact of this as things under the hood could be making a big difference (good or bad, also)

sg-gs:
Could we at least run this against a bigger test case enough to see if there is any difference? Just a test calling these functions decoupled from everything else. I think that test could be pretty fast to set up just faking files/folders with subtrees in an array and passing it to this function. 

If this is not the case, let's consider this as a rule of thumb for future optimizations. These kinds of things are the reasons why providers like Dropbox set max sync item [limits](https://help.dropbox.com/storage-space/file-storage-limit) on their products.

PR #317, opened 2023-07-20
[PB-635]: In memory items for faster feedback

PixoDev:
This system allow us to provide a faster feedback in the UI

**How**

When we perform an operation such a file rename, we first create an in memory renamed file, then we perform the actual HTTP operation, once we receive the HTTP operation result, we either rollback to the previous state if it failed, or we remove the in memory item and replace it with the local db one

**Why**

If we are going to aim to provide an almost native experience, we need to provide fast feedback to the user, otherwise this won't feel like the user is navigating though the OS filesystem. If we depend on the server latency to provide feedback, and the server is facing some slowness, we are going to pass that slowness to the UI feedback which is not a nice UX

**Corner cases**

Could be possible that if we perform an optimistic operation such folder creation, the user tries to perform an action under that optimistic folder, since the folder doesn't exist yet, the operation will fail.

### DIFF
--- a/src/workers/webdav/modules/files/application/WebdavFileCreator.ts
+++ b/src/workers/webdav/modules/files/application/WebdavFileCreator.ts
@@ -12,7 +12,7 @@ import { WebdavServerEventBus } from '../../shared/domain/WebdavServerEventBus';
 import { ContentFileUploader } from '../domain/ContentFileUploader';
 import { WebdavIpc } from '../../../ipc';
 import { Stopwatch } from '../../../../../shared/types/Stopwatch';
-
+import Logger from 'electron-log';
 export class WebdavFileCreator {
   constructor(
     private readonly repository: WebdavFileRepository,
@@ -35,7 +35,9 @@ export class WebdavFileCreator {
       this.ipc.send('WEBDAV_FILE_UPLOADING', {
         name: metadata.name,
         extension: metadata.extension,
-        nameWithExtension: metadata.name + (metadata.extension.length >= 0 ? '.' + metadata.extension : ''),
+        nameWithExtension:
+          metadata.name +
+          (metadata.extension.length >= 0 ? '.' + metadata.extension : ''),
         size: metadata.size,
         processInfo: { elapsedTime: stopwatch.elapsedTime() },
       });
@@ -45,7 +47,9 @@ export class WebdavFileCreator {
       this.ipc.send('WEBDAV_FILE_UPLOADING', {
         name: metadata.name,
         extension: metadata.extension,
-        nameWithExtension: metadata.name + (metadata.extension.length >= 0 ? '.' + metadata.extension : ''),
+        nameWithExtension:
+          metadata.name +
+          (metadata.extension.length >= 0 ? '.' + metadata.extension : ''),
         size: metadata.size,
         processInfo: { elapsedTime: stopwatch.elapsedTime(), progress },
       });
@@ -55,7 +59,9 @@ export class WebdavFileCreator {
       this.ipc.send('WEBDAV_FILE_UPLOAD_ERROR', {
         name: metadata.name,
         extension: metadata.extension,
-        nameWithExtension: metadata.name + (metadata.extension.length >= 0 ? '.' + metadata.extension : ''),
+        nameWithExtension:
+          metadata.name +
+          (metadata.extension.length >= 0 ? '.' + metadata.extension : ''),
         error: error.message,
       });
     });
@@ -66,7 +72,9 @@ export class WebdavFileCreator {
       this.ipc.send('WEBDAV_FILE_UPLOADED', {
         name: metadata.name,
         extension: metadata.extension,
-        nameWithExtension: metadata.name + (metadata.extension.length >= 0 ? '.' + metadata.extension : ''),
+        nameWithExtension:
+          metadata.name +
+          (metadata.extension.length >= 0 ? '.' + metadata.extension : ''),
         size: metadata.size,
         processInfo: { elapsedTime: stopwatch.elapsedTime() },
       });
@@ -83,11 +91,28 @@ export class WebdavFileCreator {
 
     await this.repository.add(file);
 
-    this.temporalFileCollection.remove(filePath.value);
-
     await this.eventBus.publish(file.pullDomainEvents());
 
     return file;
+  }
+
+  async createInMemoryTemporaryFile(path: string, size: number) {
+    const filePath = new FilePath(path);
+    const folder = this.folderFinder.run(filePath.dirname());
+    const metadata = ItemMetadata.from({
+      createdAt: Date.now(),
+      updatedAt: Date.now(),
+      name: filePath.name(),
+      size,
+      extension: filePath.extension(),
+      type: 'FILE',
+      visible: true,
+      externalMetadata: {
+        folderId: folder.id,
+      },
+    });
+
+    this.temporalFileCollection.add(filePath.value, metadata);
   }
 
   async run(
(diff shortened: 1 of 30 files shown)


src/workers/webdav/modules/files/application/WebdavFileRenamer.ts
```diff
@@ -40,39 +45,45 @@ export class WebdavFileRenamer {
...
+      }
+
+      if (path.extensionMatch(file.type)) {
+        await this.rename(file, path);
+        return;
+      }
+
+      await this.reupload(file, path);
+    } catch (error) {
+      this.inMemoryItems.remove(path.value);
```
JoanVicens, line 85: Shoulden't be the "rollback" action "add" if the first action was to remove it?

JoanVicens, line 85: Shoulden't be the "rollback" action "add" if the first action was to remove it?

JoanVicens, line 85: Shouldn't the "rollback" action "add" also file removed at the beginning?

src/workers/webdav/modules/files/domain/WebdavFileRepository.ts
```diff
@@ -17,4 +17,6 @@ export interface WebdavFileRepository {
   searchOnFolder(
     folderId: WebdavFolderAttributes['id']
   ): Promise<Array<WebdavFile>>;
+
+  runRemoteSync(): Promise<void>;
```
JoanVicens, line 21: The repository interface should not know about the sync, that is part of the implementation

src/workers/webdav/modules/files/infrastructure/persistance/InMemoryTemporalFileMetadataCollection.ts
```diff
@@ -7,22 +7,99 @@ export class InMemoryTemporalFileMetadataCollection
...
-    );
+  exists(pathToCheck: string): boolean {
+    return this.get(pathToCheck) ? true : false;
   }
 
-  get(path: string): ItemMetadata | undefined {
-    return this.collection[path];
+  existsByLastPath(pathToCheck: string): boolean {
+    let exists = false;
+    Object.keys(this.collection).forEach((path) => {
```
JoanVicens, line 55: exists the find method from array that would simplify this method

src/workers/webdav/modules/files/infrastructure/persistance/InMemoryTemporalFileMetadataCollection.ts
```diff
@@ -7,22 +7,99 @@ export class InMemoryTemporalFileMetadataCollection
...
+      if (item.lastPath && item.lastPath === pathToCheck) {
+        exists = true;
+      }
+    });
+    return exists;
+  }
+
+  getByLastPath(pathToCheck: string): ItemMetadata | undefined {
+    let match: ItemMetadata | undefined = undefined;
+    Object.keys(this.collection).forEach((path) => {
```
JoanVicens, line 72: the find method from array would be a good match here too

src/workers/webdav/modules/files/infrastructure/persistance/InMemoryTemporalFileMetadataCollection.ts
```diff
@@ -7,22 +7,99 @@ export class InMemoryTemporalFileMetadataCollection
...
+  }
+
+  getAll() {
+    return this.collection;
+  }
+
+  getAllByType(type: ItemMetadata['type']) {
+    const all: Record<string, ItemMetadata> = {};
+
+    Object.keys(this.collection).forEach((path) => {
```
JoanVicens, line 95: exists the filter method from array that would simplify this method

src/workers/webdav/modules/files/test/infrastructure/persistance/HttpWebdavFileRepository.test.ts
```diff
@@ -63,11 +66,11 @@ describe('Http File Repository', () => {
 
       fileSearchedBeforeRename.rename(new FilePath('/aa'));
 
+      Logger.info('File after', fileSearchedBeforeRename);
```
JoanVicens, line 69: We can debug the tests, so no need to add logs to it

src/workers/webdav/modules/folders/application/WebdavFolderDeleter.ts
```diff
@@ -1,11 +1,30 @@
...
+      this.inMemoryItems.add(
+        folder.path.value,
+        ItemMetadata.extractFromFolder(folder)
+      );
+
+      folder.trash();
+      this.inMemoryItems.update(folder.path.value, {
+        visible: false,
+      });
+      await this.repository.trash(folder);
```
JoanVicens, line 23: Same as before with the await in repository

src/workers/webdav/modules/folders/application/WebdavFolderRenamer.ts
```diff
@@ -1,15 +1,31 @@
...
+      this.inMemoryItems.add(
+        path.value,
+        ItemMetadata.extractFromFolder(folder)
+      );
 
-    await this.repository.updateName(folder);
+      await this.repository.updateName(folder);
+      await this.repository.runRemoteSync();
+    } catch (error) {
+      this.inMemoryItems.remove(path.value);
```
JoanVicens, line 27: If in line 15 the folder is removed shouldn't add it back here?

src/workers/webdav/modules/folders/infrastructure/HttpWebdavFolderRepository.ts
```diff
@@ -69,36 +66,45 @@ export class HttpWebdavFolderRepository implements WebdavFolderRepository {
...
-        }
+      const filteredInMemoryFolders: Record<string, WebdavFolder> = {};
+
+      Object.keys(inMemoryFolders).forEach((path) => {
+        const inMemoryFolder = inMemoryFolders[path];
+        if (!inMemoryFolder.visible) return;
+
+        filteredInMemoryFolders[path] = WebdavFolder.from({
+          // -1 means this folder is being created and doesn't exist yet in the server
+          id: inMemoryFolder.externalMetadata?.id || -1,
```
JoanVicens, line 96: I'm against this, WebdavFolder, Files, etc are the core models and should always be valid ones. Not being able to trust everywhere else if a folder is valid or not will be a problem. For example, on every file action that requires its folder, we will need to check if the id is valid or not

larry-internxt (approved):

src/workers/webdav/modules/folders/infrastructure/HttpWebdavFolderRepository.ts
PixoDev: Do you have any suggestion on how to solve this?

src/workers/webdav/modules/files/application/WebdavFileMover.ts
```diff
@@ -61,33 +65,72 @@ export class WebdavFileMover {
...
-    if (hasToBeOverwritten && !overwrite) {
-      throw new FileAlreadyExistsError(to);
-    }
-
-    const destinationFolder = this.folderFinder.run(desiredPath.dirname());
-
-    if (file.hasParent(destinationFolder.id)) {
-      await this.fileRenamer.run(file, to);
+    try {
+      this.inMemoryItems.add(
```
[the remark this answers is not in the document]
PixoDev: we are not adding the file that already exists, we are adding the file that will exists, which will be at `desiredPath.value`

src/workers/webdav/modules/files/application/WebdavFileCreator.ts
```diff
@@ -83,13 +91,30 @@ export class WebdavFileCreator {
 
     await this.repository.add(file);
```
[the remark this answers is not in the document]
PixoDev: The point of adding this to the memory is so if the client asks us for the tree, we can provide it with the server items + in memory items, we don't need to unawait things, they can take the same time as before, but in the meanwhile, we store the item in memory

PR #318, opened 2023-07-21
[PB-705]: Refresh folder after notification event

larry-internxt:
- Connect to notifications server with newToken (drive-server is now fixed)

### DIFF
--- a/src/main/realtime.ts
+++ b/src/main/realtime.ts
@@ -17,7 +17,7 @@ function cleanAndStartRemoteNotifications() {
   socket = io(process.env.NOTIFICATIONS_URL, {
     transports: ['websocket'],
     auth: {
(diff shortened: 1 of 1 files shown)


src/main/realtime.ts
```diff
@@ -17,7 +17,7 @@ function cleanAndStartRemoteNotifications() {
   socket = io(process.env.NOTIFICATIONS_URL, {
     transports: ['websocket'],
     auth: {
-      token: obtainToken('bearerToken'),
+      token: obtainToken('newToken'),
```
PixoDev, line 20: mm you sure about this?

larry-internxt: yeah

PixoDev (approved):

PR #322, opened 2023-07-26
[PB-604]: Feat/Show "Virtual Drive" instead of localhost

larry-internxt:
- Server is now created with the hostname 'Virtual.Drive.Internxt.com' which points to localhost ip. It's a temporary solution to the Finder name problem while we find another better solution

### DIFF
--- a/src/workers/webdav/worker/server.ts
+++ b/src/workers/webdav/worker/server.ts
@@ -19,7 +19,7 @@ export class InternxtWebdavServer {
     storageManager?: IStorageManager
   ) {
     this.server = new WebDAVServer({
-      hostname: 'localhost',
+      hostname: 'virtual.drive.internxt.com',
       port,
       requireAuthentification: false,
       storageManager: storageManager,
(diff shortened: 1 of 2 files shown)


src/workers/webdav/worker/server.ts
```diff
@@ -19,7 +19,7 @@ export class InternxtWebdavServer {
     storageManager?: IStorageManager
   ) {
     this.server = new WebDAVServer({
-      hostname: 'localhost',
+      hostname: 'virtual.drive.internxt.com',
```
PixoDev, line 22: Why is lowercase here and uppercase in the other file?

larry-internxt: get the IP from a host its not case sensitive (thats why here its not needed), but showing it on Finder it is

larry-internxt: Changed it for better consistency

PixoDev (approved):

PR #325, opened 2023-07-27
[PB-92] feat: upload empty files

JoanVicens:
Allows to create empty files by creating only the entry on drive server


src/workers/webdav/modules/files/test/application/WebdavFileCreator.test.ts
```diff
@@ -28,7 +28,7 @@ describe('Webdav File Creator', () => {
     eventBus = new EventBusMock();
     ipc = new WebdavIpcMock();
 
-    SUT = new WebdavFileCreator(
+    SUT = new WebdavFileUploader(
       fileReposiotry,
```
SimonGodefroid, line 32:
```suggestion
      fileRepository,
```

PR #326, opened 2023-07-27
[PB-743]:(feature) Update tray icon with webdav operations state

PixoDev:
Adds 4 icon states for the tray icon:

LOADING:
When the app is being initialized

STANDBY:
When the app is not processing

SYNCING:
When the app is processing some operation (move, create folder, delete, download, upload)

ALERT:
When something failed during an operation, or webdav failed to start


src/main/analytics/webdav-handlers.ts
```diff
@@ -5,11 +5,32 @@ import { WebdavErrorContext } from '../../shared/IPC/events/webdav';
 import { broadcastToWindows } from '../windows';
 import { VirtualDriveStatus } from '../../workers/webdav/VirtualDrive';
 import { ipcMain } from 'electron';
+import { setTrayStatus } from '../tray';
 
 function subscribeToFlowEvents(ipc: IpcWebdavFlow) {
+  ipc.on('WEBDAV_FOLDER_CREATING', () => {
```
JoanVicens, line 11: This could have its own place and keep tray and analytics separated

JoanVicens (approved):

PR #327, opened 2023-07-28
[PB-708]:(fix) Store files in tmp directory before upload

PixoDev:
On MacOS we don't receive the file size when opening a write stream since the Finder doesn't send it in the headers. This PR solves that problem by creating a tmp file which we later stat to get the real size

### DIFF
--- a/src/workers/webdav/modules/files/application/WebdavFileCreator.ts
+++ b/src/workers/webdav/modules/files/application/WebdavFileCreator.ts
@@ -1,4 +1,4 @@
-import { PassThrough, Writable } from 'stream';
+import { PassThrough, Readable, Writable } from 'stream';
 import { WebdavFolderFinder } from '../../folders/application/WebdavFolderFinder';
 import { FilePath } from '../domain/FilePath';
 import { RemoteFileContentsManagersFactory } from '../domain/RemoteFileContentsManagersFactory';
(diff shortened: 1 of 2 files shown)


src/workers/webdav/modules/files/application/WebdavFileCreator.ts
```diff
@@ -91,55 +96,84 @@ export class WebdavFileCreator {
...
-      extension: filePath.extension(),
-      type: 'FILE',
+    const fileValidator = new WebdavFileValidator();
+
+    const isValid = fileValidator.validateName(filePath.nameWithExtension());
+
+    // Invalid files goes nowhere
+    if (!isValid) return { stream: new PassThrough() };
+
+    const userDataPath = await ipcRenderer.invoke('get-path', 'userData');
```
JoanVicens, line 109: for temporal files, we can use the temporal file folder of the system ('temp')

src/workers/webdav/modules/files/application/WebdavFileCreator.ts
```diff
@@ -91,55 +96,84 @@ export class WebdavFileCreator {
...
-      name: filePath.name(),
-      size,
-      extension: filePath.extension(),
-      type: 'FILE',
+    const fileValidator = new WebdavFileValidator();
+
+    const isValid = fileValidator.validateName(filePath.nameWithExtension());
+
+    // Invalid files goes nowhere
+    if (!isValid) return { stream: new PassThrough() };
```
JoanVicens, line 107: Why don't we throw an error, if we return a stream, we will return "good" HTTP status, and the clients will think they are uploading a file to that writable

JoanVicens (approved):

JoanVicens:
We can leave it as an enchantment. In order to create a file, first we receive a PUT (I believe), to create a resource, and then we request to get the writable for that resource. The temporal file creation could be handled there.

src/workers/webdav/modules/files/application/WebdavFileCreator.ts
PixoDev: yeah I think it will be better

JoanVicens:
Closing this as the macOS app is a separated repo now

PR #331, opened 2023-08-07
Feature/DE-197-Update-Design-System-And-Adapt-To-Dark-Mode

aboredvaro:


sg-gs:
Hey @aboredvaro be careful to talk with @PixoDev so any iOS required design is not done here but on the native implementation we are working on.

src/main/preload.js
```diff
@@ -130,6 +130,15 @@ contextBridge.exposeInMainWorld('electron', {
   toggleAutoLaunch() {
     return ipcRenderer.invoke('toggle-auto-launch');
   },
+  toggleDarkMode(mode) {
```
JoanVicens, line 133: toggleDarkMode as defined on preload.d.ts has to return a promise

src/main/windows/settings.ts
```diff
@@ -49,7 +49,22 @@ ipcMain.on(
...
+          height: Math.trunc(height),
+        },
+        true
+      );
     }
   }
 );
+
+ipcMain.handle('dark-mode:light', () => {
+  nativeTheme.themeSource = 'light';
```
JoanVicens, line 63: I believe we also need to set the theme when the app starts, not found any clues in the documentation but after setting the dark theme and restarting the app (on dev) the theme was back to light even when the config correctly loaded the dark one.
image

aboredvaro: Good catch! I'll look into this and make the theme update on start 👍🏻

src/main/preload.js
aboredvaro: Right, fixing it, thanks!

JoanVicens (approved):

PR #336, opened 2023-08-09
[PB-842] feat: resuse webdav events

JoanVicens:
Changed listeners that were listening for `WEBDAV_*` to only the action:
Example: `WEBDAV_FILE_DELETED` is now `FILE_DELETED`.

I grouped all those events as "drive" events. We can talk about the name, this was the best I could think of at the moment.

I also cleaned up the analytics listeners to only track the events and moved the other actions to more fitting files


larry-internxt (approved):
pending fix conflicting files but LGTM

PR #343, opened 2023-08-22
[PB-683] fix: hide backups UI

JoanVicens:


aboredvaro (changes requested):
I uploaded some changes to the UI to be more accurate to the Figma description.

On the other hand, when the user removes from the list a folder with an error, it stills shows the error even if the folder no longer needs to be located (because it has been removed from the "folders to back up" list). This issue is gone when the user clicks "Backup now", but not when there are changes in the "folders to back up" list (List inside "Select folders to back up" button).

aboredvaro (approved):
Now it works flawlessly, thank you Joan! 🤗

PR #351, opened 2023-09-06
[_] feat: block deletion until is finished

JoanVicens:
Integrates the changes on https://github.com/internxt/node-win/pull/4

Uses the `notifyDeleteCallback` instead of the `notifyDeleteCompletionCallback` so until `true `is returned Windows doesn´t remove the file from the explorer.

Also adds FileFinderByContentsId to encapsulate find file logic as will be needed in more places


JoanVicens:
When the `notifyDeleteCallback` executes I get 
```
status: 0
Error in CfExecute().
Error in CfExecute(), HRESULT: 8007018e
status: 0
Error in CfExecute().
```

PR #352, opened 2023-09-06
[PB-944] Feat: include file size on placeholder

JoanVicens:
Integrates the changes on https://github.com/internxt/node-win/pull/5

Creates the file placeholder with the file size

Uses TreeBuilder service that does not use the FileRepository. This way in the future we be able to change the HttpFileRepository to not store the Record of path File and simply query the SQLite database.


JoanVicens:
This PR is in draft until https://github.com/internxt/node-win/pull/5 is merged to master

PR #386, opened 2023-10-02
[PB-1031] fix: return only base if path comes from a folder

apsantiso:
- Prevent extract last letter from a folderPath that contains multiple dots.

### DIFF
--- a/src/workers/sync-engine/modules/folders/domain/FolderPath.ts
+++ b/src/workers/sync-engine/modules/folders/domain/FolderPath.ts
@@ -17,7 +17,7 @@ export class FolderPath extends Path {
       return 'Internxt Drive';
     }
 
-    return super.name();
+    return super.name(true);
   }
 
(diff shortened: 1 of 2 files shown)


src/workers/sync-engine/modules/folders/domain/FolderPath.ts
```diff
@@ -17,7 +17,7 @@ export class FolderPath extends Path {
       return 'Internxt Drive';
     }
 
-    return super.name();
+    return super.name(true);
```
JoanVicens, line 20: If the logic is not the same for all Paths (files and folders) I would rather not call the parent class and return here `path.basename(this.value)` as we already know it's folder because we are in a FolderPath method. Nevertheless, looks good to me to merge

JoanVicens (approved):

JoanVicens:
Tried the changes with the app, but seems like that was not the issue

PR #424, opened 2023-11-20
[_] fix/wait-busy-files-updated

miguelsw:
Update: modified the wait function to try to load when it is finished being occupied by the windows process


miguelsw:
@JoanVicens it's ready to review

JoanVicens (approved):
The retries left are not used, but I think it's better this way so slow machines don't always fail to upload big files

PR #432, opened 2023-12-05
[PB-1268]: Clear temp folder on user log in and out

larry-internxt:
- Added helpers to clear the user temp folder when login in and out
- Fixed virtual root naming typo


miguelsw (approved):
LGTM

PR #436, opened 2023-12-22
[_]: fix/use basename insted of name to create folders

miguelsw:
Context:
In the creation of folders, we have been using the name provided by` path.posix.parse(folderPath`). However, for folders, it seems better to use basename since it provides the base name of the folders. Only use name in the case of files, as we want it without an extension

### DIFF
--- a/src/context/virtual-drive/folders/domain/FolderPath.ts
+++ b/src/context/virtual-drive/folders/domain/FolderPath.ts
@@ -20,6 +20,10 @@ export class FolderPath extends Path {
     return super.name();
   }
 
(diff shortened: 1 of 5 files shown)


src/context/virtual-drive/folders/domain/FolderPath.ts
```diff
@@ -20,6 +20,10 @@ export class FolderPath extends Path {
     return super.name();
   }
 
+  basename(): string {
```
JoanVicens, line 23: use get

JoanVicens (approved):

PR #438, opened 2024-01-10
[PB-1489] Fix: ignore nodes with same name

JoanVicens:
**Why**
On web, we can create files and folders with the same name but on desktop, we cannot due how the OS File System does not allow it.

**Proposed solution**
Ignore one of the "duplicated" elements and notify the user that we cannot sync both of them.


JoanVicens:
I've updated the message as requested by the product team

miguelsw:
I've updated the PR to point the changes on windows branch.

PR #442, opened 2024-01-18
[PB-1513]: Fix/loop-when-delete-files

miguelsw:
Context:
- Emited events from node-win to delete files, It is verified that there are no folders pending deletion in the queue so a loop was triggered.

Update:
- Retry mechanism to avoid error on deleting folders
- Remove folders from queue when a folder father is received
- Remove always files from queue if its folder was added in folder queue
- Always remove folder from  queue when is executed to delete from remote
- We should create the tree of folder with error during deleting.

### DIFF
--- a/src/context/virtual-drive/files/application/FileFolderContainerDetector.ts
+++ b/src/context/virtual-drive/files/application/FileFolderContainerDetector.ts
@@ -2,6 +2,7 @@ import { FolderFinder } from '../../folders/application/FolderFinder';
 import { File } from '../../files/domain/File';
 import { Folder } from '../../folders/domain/Folder';
 import { FileRepository } from '../domain/FileRepository';
+import { FileNotFoundError } from '../domain/errors/FileNotFoundError';
 
 export class FileFolderContainerDetector {
   constructor(
@@ -10,9 +11,13 @@ export class FileFolderContainerDetector {
   ) {}
 
   run(contentId: File['contentsId'], folderContentId: Folder['uuid']): boolean {
-    const file = this.repository.searchByPartial({ contentsId: contentId });
-    const folder = this.folderFinder.findFromId(file?.folderId);
-    const [_, folderUuid] = folder.placeholderId.split(':');
-    return folderUuid === folderContentId;
+    const file = this.repository.searchByPartial({
+      contentsId: contentId,
+    });
+    if (!file) {
+      throw new FileNotFoundError(contentId);
+    }
+    const folder = this.folderFinder.findFromId(file.folderId);
+    return folder.uuid === folderContentId;
   }
 }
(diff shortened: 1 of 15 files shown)


src/context/virtual-drive/files/application/FileFolderContainerDetector.ts
```diff
@@ -0,0 +1,18 @@
...
+export class FileFolderContainerDetector {
+  constructor(
+    private readonly repository: FileRepository,
+    private readonly folderFinder: FolderFinder
+  ) {}
+
+  run(contentId: File['contentsId'], folderContentId: Folder['uuid']): boolean {
+    const file = this.repository.searchByPartial({ contentsId: contentId });
+    const folder = this.folderFinder.findFromId(file?.folderId);
+    const [_, folderUuid] = folder.placeholderId.split(':');
```
JoanVicens, line 15: folder has the UUID stored, so you can simply do `folder.uuid`

src/context/virtual-drive/files/application/FileFolderContainerDetector.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class FileFolderContainerDetector {
+  constructor(
+    private readonly repository: FileRepository,
+    private readonly folderFinder: FolderFinder
+  ) {}
+
+  run(contentId: File['contentsId'], folderContentId: Folder['uuid']): boolean {
+    const file = this.repository.searchByPartial({ contentsId: contentId });
+    const folder = this.folderFinder.findFromId(file?.folderId);
```
JoanVicens, line 14: I would avoid having findFromId with the id optional. If there is no ID we already know that it will fail to find it. So I would add a guard clause, this way if the file is not found, then the error thrown is from the file not being in the repo and not that a folder

src/context/virtual-drive/shared/domain/DelayQueue.ts
```diff
@@ -20,26 +20,29 @@ export class DelayQueue {
...
       }
 
       Logger.debug(this.name, 'delay queue blocked');
       this.setTimeout();
-    }, DelayQueue.DELAY);
+    }, delay);
+  }
+
+  private async executeFn(item: string) {
+    this.queue.delete(item);
```
JoanVicens, line 44: Shouldn't be the other way around, first execute the function and delete it from the queue? I'm worried that if the deletion takes too long it can allow to delete files that does not need to be deleted

JoanVicens (commented):
Awesome work, I only have a few nitpicks

JoanVicens (approved):

c27gc (approved):

cg27shokworks (approved):

PR #450, opened 2024-02-02
[PB-1519] feat: faster reads from the file system

JoanVicens:


src/apps/fuse/callbacks/StatFsCallback.ts
```diff
@@ -16,15 +15,23 @@ type StatFsCallbackData = {
...
+  constructor() {
     super('Stat FS');
   }
 
   async execute(path: string) {
-    Logger.debug('Stats FS', path);
-
     return this.right({
-      ...this.localStats,
+      bsize: 1000000,
```
SimonGodefroid, line 24: This could be a constant, it'd also avoid repeating a magic number

PR #465, opened 2024-03-01
[PB-1420] feat: notify user of drive actions

JoanVicens:


JoanVicens:
The commits are not getting updated after I pushed a few more. I will close this one in favor of https://github.com/internxt/drive-desktop/pull/466

PR #492, opened 2024-04-15
[_] fix: ensure newer files are included

JoanVicens:
To avoid prs that depend on other prs. This one removes the StorageApp for the `RemoteSyncManager` and changes the checkpoint of files and folders to be 6 hours before the last updated item.

The StorageApp has been deleted and instead of managing the `lastFilesSyncAt` and `lastFoldersSyncAt`, we take advantage of the local database to query those values. This avoids showing the onboarding window on every login, the reset of the selected language, and the creation of new devices after logging out and back in.

When fetching the items we check for items 6 hours previous to the latest one saved to avoid missing some items when the backend does not return them on the synchronization.

### DIFF
--- a/src/apps/main/analytics/user-handlers.ts
+++ b/src/apps/main/analytics/user-handlers.ts
@@ -6,7 +6,6 @@ import {
   userSignin,
   userSigninFailed,
 } from './service';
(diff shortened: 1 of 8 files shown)


miguelsw (approved):
💯

PR #493, opened 2024-04-15
[_] feat/debounce remote manager sync

JoanVicens:
Instead of delaying the synchronization after the first notification, it's delayed if more notifications arrive in a given threshold


miguelsw (approved):
👍

PR #504, opened 2024-06-17
Change: Progressive file retrieval implemented

ArceDanielShok:


src/apps/main/remote-sync/RemoteSyncManager.ts
```diff
@@ -109,16 +108,19 @@ export class RemoteSyncManager {
...
    * Throws an error if there's a sync in progress for this class instance
    */
-  async startRemoteSync() {
+  async startRemoteSync(folderId?: number) {
     // const start = Date.now();
     Logger.info('Starting remote to local sync');
     Logger.info('Checking if we are in a valid state to start the sync');
 
-    const testPassed = this.smokeTest();
+    // const testPassed = this.smokeTest();
```
sg-gs, line 116: Using a smoke test is a good practise, so you can ensure everything is in the situation you expect to find before doing any step. If this does not fit your current implementation, please change the smoke test implementation to adapt to the circumstances and expectations you would like to find to run a successful sync.

PR #535, opened 2025-01-23
[_]: doc/new release instructions

ArceDanielShok:
These are the new instructions to release the app manually meanwhile we prepare the automation

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,37 @@ This will sign the application.
 
 **IMPORTANT:** Signing the application will modify its SHA512 hash.
 
+### Why Updating the Hash in `latest.yml` is Important?
+
+The `latest.yml` file is critical for the auto-updater to function properly. It contains metadata about the latest release, including the SHA512 hash of the installer file. This hash ensures the integrity and authenticity of the application during the update process.
+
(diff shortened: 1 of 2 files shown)


sg-gs:
Hey @ArceDanielShok, no need to ping someone to ask for a review. In GitHub, you can just add the reviewers to the list, they will get notified.

Also, consider providing a good description of the PR as I did (although maybe too concise) so that in case we need to go back to this Pull Request in the future, we can get useful information.

README.md
```diff
@@ -61,3 +61,109 @@ internxt
...
+```
+
+This will output the new hash.
+You can compare the hash before and after signing to confirm the modification.
+
+---
+
+## Step 4: Update the `latest.yml` File
+
+The `latest.yml` file contains the SHA512 hash in Base64 format.
```
sg-gs, line 137: Explain why this step needs to be done according to how electron-updater behaves, so we do not need to guess it

sg-gs (changes requested):

dajimenezriv-internxt:
I'm trying to follow the process, but I got stuck because I don't have Internxt Universal Technologies SL certificate. I'm not seeing it in Confluence, so I guess it's not meant to be stored there.

dajimenezriv-internxt (approved):

sg-gs (approved):
LGTM @ArceDanielShok

PR #579, opened 2025-02-17
Refactor SyncManager to Support Workspaces

ArceDanielShok:
### Refactor SyncManager to Support Workspaces

The management of files and folders within a Workspace has been implemented, separating the logic between the normal drive and workspace drives.

- Sync Files and Sync Folder now correctly handle files within a workspace.
- When initializing RemoteSyncManager, a workspaceID must be provided:
  - If null, it will operate as the normal drive.
  - If an workspaceID is passed, it will work within the specified workspace.
- A Map has been created to manage multiple RemoteSyncManager instances, allowing them to be referenced by their ID.


cg27shokworks (approved):

dajimenezriv-internxt:
I think this PR changes so many important things to be reviewed and should be split in more PRs. I also see that @cg27shokworks approved it, but as talked with @ArceDanielShok on friday, there were some things that should be checked before merging this PR, like for example the types of query endpoints.

src/apps/main/realtime.ts
```diff
@@ -60,7 +60,7 @@ function cleanAndStartRemoteNotifications() {
   });
 
   socket.on('connect_error', (error) => {
-    Logger.error('❌ Remote notifications connect error: ', error);
+    // Logger.error('❌ Remote notifications connect error: ', error);
```
dajimenezriv-internxt, line 63: Should we uncomment this?

src/context/virtual-drive/files/infrastructure/dtos/client.dto.ts
```diff
@@ -0,0 +1,4 @@
+import { components } from './../../../../../apps/shared/HttpClient/schema.d';
+export type persistFileDto = components['schemas']['CreateFileDto'];
```
dajimenezriv-internxt, line 2: Types should be in CamelCase.

dajimenezriv-internxt (changes requested):

src/apps/main/quit.ts
```diff
@@ -7,8 +7,6 @@ export function setCleanUpFunction(fn: () => Promise<void>) {
 }
 
 export async function quitApp() {
-  // await stopVirtualDrive();
```
dajimenezriv-internxt, line 10: Why this is commented and why this is related with b2b?

ArceDanielShok: That I was commenting on I just deleted it

dajimenezriv-internxt (approved):

PR #586, opened 2025-02-24
release: 2.4.3

cg27shokworks:
- Feature: Added windows toast notifications when the storage limit is reached.
- Bugfix: Separated execution contexts from rename callbacks

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "internxt-drive",
-  "version": "2.4.1",
+  "version": "2.4.3-qa",
   "author": "Internxt <hello@internxt.com>",
   "description": "Internxt client UI",
   "license": "AGPL-3.0",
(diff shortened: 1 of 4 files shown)


src/apps/main/background-processes/process-issues.ts
```diff
@@ -7,6 +8,31 @@ import {
...
+
+  if (now - lastDialogTime < TWO_MINUTES) {
+    return;
+  }
+  lastDialogTime = now;
+
+  const notification = new Notification({
+    title: 'Internxt',
+    body: 'Your account storage limit has been reached, for more details go to Settings -> Issues',
+    icon: path.join(__dirname, 'assets', 'icon.ico')
```
dajimenezriv-internxt, line 27: Isn't __dirname giving the path apps/main/backgroup-processes? Shouldn't we use process.cwd() to access from root?

dajimenezriv-internxt (changes requested):

dajimenezriv-internxt (approved):

PR #595, opened 2025-02-25
Workspace Header Implementation, Token Refresh Job, and Client Adaptation

ArceDanielShok:
- Logic to insert Workspace headers: The necessary logic has been added to insert Workspace-specific headers into HTTP requests. This allows for proper identification and authorization of requests within each workspace.
- Creation of a token refresh job (23 hours): A job has been implemented that runs periodically (every 23 hours) to refresh access tokens. This ensures continuous authentication and prevents token expiration.
- Client modification for dual functionality (Drive and Main): The client has been adapted to function correctly in both the Drive context and the main application context (Main). This provides flexibility and allows for client reuse in different parts of the system.


src/apps/main/database/entities/DriveFile.ts
```diff
@@ -30,7 +30,7 @@ export class DriveFile {
   folderId!: number;
 
   @Column({ nullable: true })
-  folderUuid?: string;
+  folderUuid!: string;
```
dajimenezriv-internxt, line 33: Now typeorm doesn't match the type of typescript. See that nullable is still true

src/apps/main/remote-sync/handlers.ts
```diff
@@ -234,7 +237,7 @@ remoteSyncManagers.forEach((manager) => {
   manager.onStatusChange((newStatus) => {
     if (!initialSyncReady && newStatus === 'SYNCED') {
       initialSyncReady = true;
-      eventBus.emit('INITIAL_SYNC_READY');
+      // eventBus.emit('INITIAL_SYNC_READY');
```
dajimenezriv-internxt, line 240: Why is this uncommented?

src/apps/shared/HttpClient/client.ts
```diff
@@ -1,12 +1,27 @@
...
+  if (providerId) {
+    return await ipcRendererSyncEngine.invoke('GET_HEADERS');
+  }
+  return getNewApiHeaders();
+};
 
 const middleware: Middleware = {
   async onRequest({ request }) {
-    const headers = getNewApiHeaders();
+    const headers = await getHeaders();
```
dajimenezriv-internxt, line 18: Why now we retrieve the headers from the ipc instead of calling the function getNewApiHeaders?

src/apps/main/remote-sync/handlers.ts
```diff
@@ -166,6 +167,11 @@ ipcMain.handle('GET_UPDATED_REMOTE_ITEMS', async (_, workspaceId = '') => {
   Logger.debug('[MAIN] Getting updated remote file items ' + workspaceId);
   return getUpdatedRemoteItems(workspaceId);
 });
+ipcMain.handle('REFRESH_WORKSPACE_TOKEN', async (_, workspaceId: string): Promise<string> => {
```
dajimenezriv-internxt, line 170: Isn't it possible to call directly the function instead of using events? Everything done with events loses it's type safety.

src/apps/main/remote-sync/handlers.ts
ArceDanielShok: I commented out the code because the INITIAL_SYNC_READY event is responsible for starting the syncEngine, and I believe it's better to initialize them after all files and folders have been fully synchronized. However, I left it commented in case, in the near future, we need to start each syncEngine separately. For now, all synchronizations start at the same time.

src/apps/shared/HttpClient/client.ts
ArceDanielShok: I don't call getNewApiHeaders directly because if any syncEngine process tries to use the client, it breaks when calling that function. To prevent this issue, I now validate the following:

- If the execution comes from the main process, I call getNewApiHeaders directly.
- Otherwise, I retrieve the headers via IPC events.

We might need to explore a better way to handle this logic in the future.

src/apps/main/remote-sync/handlers.ts
ArceDanielShok: Currently, workspaceSyncManager runs on the main thread, while workspace refresh token management is handled within the same thread as the syncEngine, and I am storing the values in my Config.

dajimenezriv-internxt (approved):

PR #597, opened 2025-02-27
Enhance user key management and add mnemonic support

ArceDanielShok:
### This Pull Request introduces significant improvements in user key management and system security:

- **Updated authentication endpoints:** Authentication endpoints have been updated to receive new user keys. This allows for more secure and flexible key management.
- **Implemented key decryption:** Logic has been implemented to decrypt user keys using the user's password. This ensures that keys are stored and used securely.
- **Stored keys in config storage**: Decrypted keys are stored in config storage for later use. This allows for efficient and secure key access.
- **Workspace key decryption:** Logic has been implemented to decrypt workspace keys using the user's private key. This ensures workspace security.
- **Stored mnemonics in global config:** Mnemonics are stored in the global config for each drive. This facilitates mnemonic management and access.
- **Changed sync engine configuration:** The sync engine configuration has been modified to set default values. This improves system stability and predictability.


package.json
```diff
@@ -252,6 +253,7 @@
     "multipipe": "^4.0.0",
     "node-schedule": "^2.1.0",
     "openapi-fetch": "^0.13.4",
+    "openpgp": "5.9.0",
```
dajimenezriv-internxt, line 256: This package (5.9.0) is wrote in typescript, I think we don't need @types

src/apps/main/auth/service.ts
```diff
@@ -56,8 +56,16 @@ function ecnryptToken(token: string): string {
   return buffer.toString(TOKEN_ENCODING);
 }
 
-export function setCredentials(userData: User, mnemonic: string, bearerToken: string, newToken: string) {
-  ConfigStore.set('mnemonic', mnemonic);
+export function setCredentials(userData: User, mnemonic: string, bearerToken: string, newToken: string, password: string) {
```
dajimenezriv-internxt, line 59: Also, this kind of functions we should start providing params as object since all props are strings and it's easy to pass them wrong

src/apps/main/background-processes/sync-engine.ts
```diff
@@ -224,14 +228,29 @@ const spawnAllSyncEngineWorker = async () => {
   await Promise.all(
     workspaces.map(async (workspace) => {
       const workspaceCredential = await FetchWorkspacesService.getCredencials(workspace.id);
+
+      // obtener la key
```
dajimenezriv-internxt, line 232: Don't write comments in spanish

src/apps/shared/crypto/service.ts
```diff
@@ -0,0 +1,51 @@
+import { Buffer } from 'buffer';
+import { Data, MaybeStream, WebStream } from 'openpgp';
+const WORDS_HYBRID_MODE_IN_BASE64 = 'SHlicmlkTW9kZQ=='; // 'HybridMode' in BASE64 format
+
+export async function getOpenpgp(): Promise<typeof import('openpgp')> {
+  return import('openpgp');
+}
+
+export const encryptMessageWithPublicKey = async ({
```
dajimenezriv-internxt, line 9: Is this function used anywhere?

package.json
ArceDanielShok: Why wouldn't we install the types? I believe it's always good to have library types

dajimenezriv-internxt: Types are only necessary when the library doesn't have it's own declarations. In this case, openpgp is wrote in typescript, so it already provides types definitions.
See the library:
https://www.npmjs.com/package/@types/openpgp
is deprecated

src/apps/renderer/pages/Login/service.ts
```diff
@@ -25,14 +25,16 @@ export type AccessResponse = {
...
 };
 
 export async function accessRequest(email: string, password: string, hashedPassword: string, tfa?: string): Promise<AccessResponse> {
   const fallbackErrorMessage = 'Error while logging in';
 
   let accessRes;
   try {
-    accessRes = await fetch(`${process.env.API_URL}/access`, {
+    window.electron.logger.info('accessRequest', `${process.env.NEW_DRIVE_URL}/login/access`);
+    accessRes = await fetch(`${process.env.NEW_DRIVE_URL}/drive/auth/login/access`, {
```
dajimenezriv-internxt, line 37: Is this right? You changed the login url

ArceDanielShok: Yes, that's correct, I switched to the new API.

dajimenezriv-internxt (approved):

PR #598, opened 2025-02-27
Migrate sync endpoints

dajimenezriv-internxt:
## What

Migrate some sync endpoints to the new API. These endpoints use uuid instead of id.
Also, with the new endpoints we have to stop using the name field since it's a legacy field and start using plainName.


src/apps/backups/BackupInfo.ts
```diff
@@ -4,5 +4,5 @@ export type BackupInfo = {
   tmpPath: string;
   backupsBucket: string;
   pathname: string;
-  name: string;
+  plainName: string;
```
ArceDanielShok, line 7: What's the reason for changing 'name' to 'plainName'?

src/apps/main/device/service.ts
```diff
@@ -195,10 +197,8 @@ function decryptDeviceName({ name, ...rest }: Device): Device {
   };
 }
 
-export type Backup = { id: number; name: string; uuid: string };
```
ArceDanielShok, line 198: What's the reason for removing Backup type?

src/apps/main/device/service.ts
```diff
@@ -314,27 +310,24 @@ export async function addBackup(): Promise<void> {
...
-    logger.error(error);
+    Logger.error(error);
   }
 }
 
-async function fetchFolder(folderId: number) {
-  const res = await fetch(`${process.env.API_URL}/storage/v2/folder/${folderId}`, {
-    method: 'GET',
-    headers: getHeaders(true),
+async function fetchFolder({ folderUuid }: { folderUuid: string }) {
```
ArceDanielShok, line 317: Is this 'Get Folder' the same one we need for 'FetchRemoteFoldersService'?

src/apps/backups/BackupInfo.ts
dajimenezriv-internxt: Currently in the our local database we are storing plainName and name (which is plainName encrypted).
So I asked @Ap1402 which one should I use:
image
It's like with previous endpoints we were sending the mnemonic in the headers so when we were fetching from the backend the name was decrypted, now it's not necessary.

src/apps/main/device/service.ts
dajimenezriv-internxt: Backup type has so many fields that may be useful, not only these three. So if we need to use any other type it would be difficult to find why backup only has those three in any other place in the code. Because this type is limiting the fields that the real object has.

src/apps/main/device/service.ts
dajimenezriv-internxt: Probably yes, it was declared in both places. In FetchRemoteFoldersService we still need to migrate to use folderUuid, but yes, it's the same endpoint.

dajimenezriv-internxt: No, it's not the same, this one is just for retrieve the information of a folder, the other one is for a group of folders.

ArceDanielShok (approved):

PR #614, opened 2025-03-06
Remove non used libraries

dajimenezriv-internxt:
## What

1. Remove non used libraries from node_modules.
2. Replace lodash with lodash.chunk and lodash.debounce.
3. Simplify clean.js script since we are only calling it with `dist` argv.
4. Remove beforeBuild.js script since we are using electron-builder and automatically performs the electron-rebuild under the hood.
5. Remove notarize.js script since it's only needed on MacOS.
6. Remove linux, dmg, mac, beforeBuild, afterSign and publich since we are not using those options on our build script.

## Results

1. Reduced size from the .exe file from 141MB to 101MB.
2. Reduced size from the installed product from 623MB to 401MB.
3. Reduced RAM usage from ~330MB to ~310MB (measured in the first minute since the installation).


.erb/scripts/clean.js
```diff
@@ -1,17 +1,4 @@
...
-  dll: webpackPaths.dllPath,
-};
-
-args.forEach((x) => {
-  const pathToRemove = commandMap[x];
-  if (pathToRemove !== undefined) {
-    rimraf.sync(pathToRemove);
-  }
-});
+rimraf.sync(webpackPaths.distPath);
```
ArceDanielShok, line 4: don't you need to delete the other 3 paths?

ArceDanielShok (approved):

PR #616, opened 2025-03-10
Refactor tests

ArceDanielShok:


.eslintignore
```diff
@@ -31,4 +31,6 @@ npm-debug.log.*
 src/workers/backups/process.ts
 src/workers/backups/*
 src/workers/filesystems/*
-src/test/*
\ No newline at end of file
+src/test/*
+
+*.d.ts
```
dajimenezriv, line 36: Why do not include these files?

src/apps/main/token-scheduler/token-scheduler.test.ts
```diff
@@ -3,16 +3,16 @@ import ms from 'ms';
 
 import { TokenScheduler } from './TokenScheduler';
 
-function calculateDayFromToday(t: string): Date {
+function calculateDayFromToday(t: number): Date {
   const today = new Date();
 
-  return new Date(today.getTime() + ms(t));
+  return new Date(today.getTime() + t);
```
dajimenezriv-internxt, line 9: Why not apply ms just one time? If it's because of the type, you can replace string with StringValue from 'ms'

.eslintignore
ArceDanielShok: to ignore files such as schema.d.ts

dajimenezriv-internxt: And why we should ignore files like schema.d.ts? They are part of the project and also are needed in pipelines validations when we start checking types again. Aren't we importing types from schema.d.ts into our code?

dajimenezriv-internxt: Ignore what I said, I though it was .eslintignore. In case of .eslintignore I think we should include just schema.d.ts, there are types like env.d.ts that I don't see why they shouldn't be eslinted.

src/context/virtual-drive/files/application/FilePathUpdater.ts
```diff
@@ -40,8 +41,12 @@ export class FilePathUpdater {
     try {
       const trackerId = await this.local.getFileIdentity(file.path);
       file.moveTo(destinationFolder, trackerId);
-    } catch (error: any) {
-      Logger.warn(`Error in FilePathUpdater.move: ${error?.message}`);
+    } catch (exc: unknown) {
+      logger.error({
```
dajimenezriv-internxt, line 45: Look that logger.error returns an error, so you can do throw logger.error in just one line and we also just throw something that is user friendly, because is the msg error, and the technical error is just logged

dajimenezriv-internxt (approved):

PR #617, opened 2025-03-11
Fix trypescripts errors

ArceDanielShok:


tsconfig.json
```diff
@@ -32,5 +32,5 @@
       "virtual-drive/dist": ["../node-win/dist"]
     }
   },
-  "include": ["env.d.ts", "src", "tests", ".erb/configs", ".erb/scripts"]
+  "include": ["env.d.ts", "src", "tests", ".erb/configs", ".erb/scripts", "**/**.d.ts"]
```
dajimenezriv-internxt, line 35: Why these addition?

ArceDanielShok: I included it so that it can recognize files like the assets.d.ts

src/apps/main/network/download.ts
```diff
@@ -247,7 +247,7 @@ async function _downloadFile(params: IDownloadParams): Promise<ReadableStream<Ui
   );
 
   return buildProgressStream(downloadStream, (readBytes) => {
-    params.options?.notifyProgress(metadata.fileMeta.size, readBytes);
+    params.options?.notifyProgress && params.options.notifyProgress(fileMeta.size, readBytes);
```
dajimenezriv-internxt, line 250: You can do "?."

src/apps/main/remote-sync/files/fetch-files.service.interface.ts
```diff
@@ -1,13 +1,18 @@
 import { RemoteSyncManager } from '../RemoteSyncManager';
 import { RemoteSyncedFile } from '../helpers';
+import { paths } from '@/apps/shared/HttpClient/schema';
+
+export type Query = paths['/files']['get']['parameters']['query'] & paths['/folders/{id}/files']['get']['parameters']['query'];
```
dajimenezriv-internxt, line 5: I wouldn't do this. Let's say that query params change for just the /files endpoint, now, we won't be able to use those types well since we are mixing them, when they are two different endpoints.

ArceDanielShok: Now separate the types

dajimenezriv-internxt (approved):

PR #631, opened 2025-03-14
Backup Availability Validation and Interface Adjustments

ArceDanielShok:
**Description:**
### 
This Pull Request introduces validations and interface adjustments based on the user's backup availability. Three distinct use cases have been implemented:
1. User with no backups and backups disabled: The backup view is hidden completely.
2. User with existing backups and backups disabled: Only backup download is allowed.
3. User with backups enabled: All backup-related functionalities are enabled.

Specific Changes:

- Implementation of validation logic to determine backup availability.
- User interface adjustments to reflect the three use cases mentioned.

### DIFF
--- a/src/apps/main/background-processes/backups/launchBackupProcesses.ts
+++ b/src/apps/main/background-processes/backups/launchBackupProcesses.ts
@@ -8,6 +8,8 @@ import { BackupsProcessStatus } from './BackupsProcessStatus/BackupsProcessStatu
 import { BackupsProcessTracker } from './BackupsProcessTracker/BackupsProcessTracker';
 import { BackupsStopController } from './BackupsStopController/BackupsStopController';
 import { isSyncError } from '../../../shared/issues/SyncErrorCause';
+import { isAvailableBackups } from '../../ipcs/ipcMainAntivirus';
+import { logger } from '@/apps/shared/logger/logger';
 
 function backupsCanRun(status: BackupsProcessStatus) {
   return status.isIn('STANDBY') && backupsConfig.enabled;
@@ -22,6 +24,15 @@ export async function launchBackupProcesses(
 ): Promise<void> {
   if (!backupsCanRun(status)) {
     Logger.debug('[BACKUPS] Already running');
+    return;
+  }
+
+  const isAvailable = await isAvailableBackups();
+
+  if (!isAvailable) {
+    logger.info({
+      msg: 'Backups are not available',
+    });
     return;
   }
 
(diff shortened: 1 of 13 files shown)


src/apps/main/background-processes/backups/launchBackupProcesses.ts
```diff
@@ -25,6 +27,15 @@ export async function launchBackupProcesses(
     return;
   }
 
+  const isAvailable = await isAvailableBackups();
+
+  if (!isAvailable) {
+    logger.info({
```
dajimenezriv-internxt, line 33: Should this log be sent to opentelemetry or should we replace it with debug? Remember that info is going to be only for important logs

src/apps/main/ipcs/ipcMainAntivirus.ts
```diff
@@ -34,12 +34,14 @@ ipcMain.handle('antivirus:is-available', async (): Promise<boolean> => {
 
     return availableProducts.antivirus;
   } catch (error) {
-    Logger.error('ERROR GETTING PRODUCTS: ', error);
-    throw error;
+    logger.info({
+      msg: 'ERROR GETTING PRODUCTS FOR ANTIVIRUS',
+      error: (error as Error).message,
```
dajimenezriv-internxt, line 39: Why not display the whole error which is going to include also the stack trace and more? And also we don't need to parse and if the error doesn't has the message field is doesn't throw any error. Also, since it's an error, we should use logger.warn if it's not too important or logger.error if it's more important. I would do:
logger.error({
msg: "wathever",
exc: error,
})

src/apps/main/ipcs/ipcMainAntivirus.ts
```diff
@@ -49,10 +51,15 @@ ipcMain.handle('backups:is-available', async (): Promise<boolean> => {
 
     return availableProducts.backups;
   } catch (error) {
-    Logger.error('ERROR GETTING PRODUCTS: ', error);
-    throw error;
+    logger.info({
```
dajimenezriv-internxt, line 54: Same here

src/apps/main/preload.js
```diff
@@ -22,7 +23,7 @@ contextBridge.exposeInMainWorld('electron', {
   },
 
   logger: {
-    info: (...message) => Logger.info(String(message)),
+    info: (...message) => Logger.info(inspect(message, { colors: true, depth: Infinity, breakLength: Infinity })),
```
dajimenezriv-internxt, line 26: Can we point to the new custom logger instead of this? Because we are going to use open telemetry in the other, so the plan is to stop using electron-log directly.

ArceDanielShok: image
for some reason the other logger breaks when calling from the preload, maybe in another PR we can fix it

PR #651, opened 2025-03-21
Refactoring Folder and File Endpoints: Migration from ID to UUID

ArceDanielShok:
This Pull Request focuses on technical debt removal and improving the implementation of folder and file endpoints. The following changes have been made:

Endpoint Migration: The get folder and get file endpoints have been migrated to use UUIDs instead of IDs. This simplifies the implementation and improves code clarity.

**Specific Changes:**
- Modified the get folder and get file endpoints to use UUIDs.
- Updated related logic to handle UUIDs instead of IDs.


src/apps/backups/index.ts
```diff
@@ -4,6 +4,8 @@ import { BackupInfo } from './BackupInfo';
 import { BackupsDependencyContainerFactory } from './dependency-injection/BackupsDependencyContainerFactory';
 import { DriveDesktopError } from '../../context/shared/domain/errors/DriveDesktopError';
 import { BackupsIPCRenderer } from './BackupsIPCRenderer';
+import { setDefaultConfig } from '../sync-engine/config';
+setDefaultConfig();
```
dajimenezriv-internxt, line 8: Can we call this function inside another function? This is problem I mentioned that you don't really know what's happening when you import a file and also in testing you are executing code that you don't want to test, so the best would be to encapsulate everything into a function. Is is possible here?

src/apps/main/remote-sync/RemoteSyncManager.ts
```diff
@@ -97,13 +97,15 @@ export class RemoteSyncManager {
     try {
       const syncFilesPromise = this.syncRemoteFiles.run({
         self: this,
-        folderId,
+        retry: 1,
```
dajimenezriv-internxt, line 100: We don't need to send the retry parameter

dajimenezriv-internxt (changes requested):

src/apps/main/device/service.ts
```diff
@@ -88,7 +88,7 @@ export async function getDevices(): Promise<Array<Device>> {
 
   const devices = ((await response.json()) as Array<DeviceDTO>) || [];
 
-  return devices.filter(({ removed, hasBackups }) => !removed && hasBackups).map((device) => decryptDeviceName(device));
+  return devices?.filter(({ removed, hasBackups }) => !removed && hasBackups).map((device) => decryptDeviceName(device));
```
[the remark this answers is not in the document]
ArceDanielShok: I don't really know why, but with new accounts where there are no devices it passes as undefined

src/apps/main/remote-sync/folders/fetch-remote-folders.service.ts
```diff
@@ -13,34 +12,44 @@ export class FetchRemoteFoldersService implements FetchFoldersService {
...
+        });
     const result = await promise;
 
     if (result.data) {
       const hasMore = result.data.length === self.config.fetchFilesLimitPerRequest;
       return { hasMore, result: result.data };
     }
 
-    throw logger.error({ msg: 'Fetch folders response not ok', query, error: result.error });
+    throw logger.error({ msg: 'Fetch folders response not ok', exc: result.error });
```
dajimenezriv-internxt, line 44: Wouldn't be better to have also the query in the log so we can check easily why the error happened or to replicate?

src/apps/main/remote-sync/files/fetch-workspace-files.service.ts
```diff
@@ -36,7 +41,7 @@ export class FetchWorkspaceFilesService implements FetchFilesService {
       return { hasMore, result: result.data };
     }
 
-    throw logger.error({ msg: 'Fetch workspace files response not ok', query, error: result.error });
+    throw logger.error({ msg: 'Fetch workspace files response not ok', error: result.error });
```
dajimenezriv-internxt, line 44: Same here about query

src/apps/main/remote-sync/files/fetch-remote-files.service.ts
```diff
@@ -20,15 +30,15 @@ export class FetchRemoteFilesService implements FetchFilesService {
       return { hasMore, result: result.data };
     }
 
-    throw logger.error({ msg: 'Fetch files response not ok', query, error: result.error });
```
dajimenezriv-internxt, line 23: And here

dajimenezriv-internxt (approved):

PR #654, opened 2025-03-23
Add open telemetry

dajimenezriv-internxt:
## What

Add open telemetry.

The docker image used for testing is https://hub.docker.com/r/grafana/otel-lgtm.

It appends the user id to any log (see prepareBody in logger.ts).

In the image we can see that we are filtering by a user.

image


dajimenezriv:
Not stale.

PR #658, opened 2025-03-25
Migrate backups endpoints

dajimenezriv-internxt:


src/infra/drive-server-wip/in/client-wrapper.service.ts
```diff
@@ -0,0 +1,34 @@
...
+
+type TProps<T> = {
+  loggerBody: TLoggerBody;
+  promise: Promise<{ data: T; error: undefined; response: Response } | { data: undefined; error: unknown; response: Response }>;
+};
+
+export class ClientWrapperService {
+  constructor(private readonly logger = loggerService) {}
+
+  async run<T>({ loggerBody, promise }: TProps<T>) {
```
c27gc, line 11: @dajimenezriv-internxt  According to the agreements we were going to try to keep explicit typing in the function output only for primitives, I see totally reasonable this kind of explicit typing in this case, but more than a request for change, I would like to know if for you in this use case it is necessary.

cg27shokworks (approved):

PR #680, opened 2025-04-01
feat: check remote files healthy

cg27shokworks:
What: 
Solves remote dangled files. This new service checks in the backup process the status of remote files and reloads them if necessary.


src/apps/backups/dangled-files/DangledFilesService.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    for (const [localFile, remoteFile] of danglingFiles.entries()) {
+      if (!remoteFile) continue;
+
+      let isDownloadable: boolean;
+      try {
+        isDownloadable = await this.isFileDownloadable(remoteFile);
+      } catch (error) {
+        logger.warn({
+          msg: '[BACKUPS] Error checking if file is downloadable in handleDangledFile',
```
dajimenezriv-internxt, line 88: Here are in some others that are similar, you can just use tag: 'BACKUPS' and it will print BACKUPS in the beginning and also add it to attributes, so you can remove it from attributes and from msg '[BACKUPS]'

PR #682, opened 2025-04-01
Logs Refactoring and Offline Folder Creation Improvement

ArceDanielShok:
This Pull Request includes two main improvements: the replacement of the Electron logger with a custom logger and the addition of extra parameters when creating offline folders before uploading them to the server.

- Replacement of the Electron logger with the custom logger:
  - The default Electron logger has been replaced with a custom logger. This allows greater control over the format and management of logs, facilitating debugging and issue tracking.
- Addition of parameters when creating offline folders:
  - The parameters uuid, status, updatedAt and createdAt have been added when creating offline folders before uploading them to the server. This provides more complete and accurate information about offline folders, facilitating their management and synchronization with the server.

### DIFF
--- a/src/context/virtual-drive/folders/infrastructure/HttpRemoteFolderSystem.ts
+++ b/src/context/virtual-drive/folders/infrastructure/HttpRemoteFolderSystem.ts
@@ -1,7 +1,6 @@
 import { Folder, FolderAttributes } from '../domain/Folder';
 import { Service } from 'diod';
 import { FolderStatuses } from '../domain/FolderStatus';
-import { OfflineFolder } from '../domain/OfflineFolder';
 import { PersistFolderDto, PersistFolderInWorkspaceDto, PersistFolderResponseDto, UpdateFolderMetaDto } from './dtos/client.dto';
 import { client } from '../../../../apps/shared/HttpClient/client';
 import { logger } from '@/apps/shared/logger/logger';
@@ -9,8 +8,8 @@ import { logger } from '@/apps/shared/logger/logger';
 export class HttpRemoteFolderSystem {
   constructor(private readonly workspaceId?: string) {}
 
-  async persist(offline: OfflineFolder): Promise<FolderAttributes> {
-    if (!offline.name || !offline.basename) {
+  async persist(offline: { basename: string; parentUuid: string; path: string }): Promise<FolderAttributes> {
+    if (!offline.basename) {
       throw new Error('Bad folder name');
     }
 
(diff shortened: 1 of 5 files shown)


src/context/virtual-drive/folders/infrastructure/HttpRemoteFolderSystem.ts
```diff
@@ -1,16 +1,15 @@
...
 import { client } from '../../../../apps/shared/HttpClient/client';
 import { logger } from '@/apps/shared/logger/logger';
 @Service()
 export class HttpRemoteFolderSystem {
   constructor(private readonly workspaceId?: string) {}
 
-  async persist(offline: OfflineFolder): Promise<FolderAttributes> {
-    if (!offline.name || !offline.basename) {
+  async persist(offline: { basename: string; parentUuid: string; path: string }): Promise<FolderAttributes> {
+    if (!offline.basename) {
```
dajimenezriv-internxt, line 12: This isn't necessary, baseline is a string always

src/context/virtual-drive/folders/infrastructure/HttpRemoteFolderSystem.ts
```diff
@@ -93,12 +92,12 @@ export class HttpRemoteFolderSystem {
...
   }
 
-  private async existFolder(offline: OfflineFolder): Promise<FolderAttributes> {
+  private async existFolder(offline: { parentUuid: string; basename: string; path: string }): Promise<FolderAttributes> {
     try {
       const response = await client.POST('/folders/content/{uuid}/folders/existence', {
         params: {
           path: {
-            uuid: offline.uuid,
+            uuid: offline.parentUuid,
```
dajimenezriv-internxt, line 100: I think this sould be uuid, and when calling this function just send parentUuid, but technically we are checking the existence of a uuid folder

src/context/virtual-drive/folders/infrastructure/HttpRemoteFolderSystem.ts
ArceDanielShok: Yes, it is always a string but it can be "", perhaps the validation would be that empty strings are not allowed

src/context/virtual-drive/folders/infrastructure/HttpRemoteFolderSystem.ts
ArceDanielShok: Here we are validating if a folder with that name exists within the parent folder

PR #705, opened 2025-04-10
Migration of IDs in Placeholders and Refactoring of File Logic

ArceDanielShok:
This Pull Request introduces several improvements and refactorings related to file identification and placeholder management.

- Migration from contentId to uuid as Placeholder ID:
  - The usage of contentId has been migrated to the uuid of the files when setting them as placeholderId within the placeholder. This unifies the use of UUID-based unique identifiers for files.
- Removal of Unused Code:
  - Code that was no longer used in the project has been identified and removed, simplifying the codebase and improving its maintainability.
- Removal of FileFinderByContentsId:
  - The FileFinderByContentsId class has been removed. File lookups are now performed directly through the repository, simplifying the data flow and eliminating a layer of indirection.
- Start of Placeholder ID Migration in BindingManager:
  - The migration of placeholderId has been initiated. When executing the update function in the BindingManager, the difference between the current placeholderID and the correct value (the file's uuid) is evaluated to update it.
- Creation of Tests for the Update Function:
  - Unit tests have been created for the update function of the BindingManager, ensuring the correct operation of the placeholderId update logic.

Specific Changes:

1. Replacement of contentId with uuid as placeholderId.
2. Removal of unused code.
3. Removal of the FileFinderByContentsId class.
4. Start of the placeholderId migration in the update function of the BindingManager.
5. Creation of tests for the update function of the BindingManager.


src/apps/sync-engine/callbacks/fetchData.service.ts
```diff
@@ -9,22 +9,22 @@ import { dirname } from 'path';
...
+  async run({ self, filePlaceholderId, callback, ipcRendererSyncEngine }: TProps) {
     try {
       Logger.debug('[Fetch Data Callback] Donwloading begins');
 
       const startTime = Date.now();
-      const path = await self.controllers.downloadFile.execute(contentsId, callback);
+      const path = await self.controllers.downloadFile.execute(filePlaceholderId, callback);
 
       // eslint-disable-next-line no-control-regex
-      const parsedContentsId = contentsId.replace(/[\x00-\x1F\x7F-\x9F]/g, '').split(':')[1];
```
dajimenezriv-internxt, line 26: This shouldn't be called parsedContentsId

src/context/virtual-drive/files/application/update/FilesPlaceholderUpdater.test.ts
```diff
@@ -0,0 +1,271 @@
...
+import { RelativePathToAbsoluteConverter } from '../../../shared/application/RelativePathToAbsoluteConverter';
+import { LocalFileIdProvider } from '../../../shared/application/LocalFileIdProvider';
+import { mockDeep } from 'vitest-mock-extended';
+import { FileMother } from 'tests/context/virtual-drive/files/domain/FileMother';
+import { FileStatuses } from '../../domain/FileStatus';
+import { v4 } from 'uuid';
+import fs from 'fs/promises';
+import { InMemoryEventRepository } from '@/context/virtual-drive/shared/infrastructure/InMemoryEventHistory';
+
+vi.mock('fs/promises');
```
dajimenezriv-internxt, line 14: It's better to wrap these imports betwen import, so it performs some kind of type check.
vi.mock(import("fs/promises"))

src/context/virtual-drive/files/application/update/FilesPlaceholderUpdater.test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+const mockRepository = mockDeep<InMemoryFileRepository>();
+
+const mockLocalFileSystem = mockDeep<NodeWinLocalFileSystem>();
+
+const mockPathConverter = mockDeep<RelativePathToAbsoluteConverter>({
+  run: vi.fn((relativePath: string) => relativePath),
```
dajimenezriv-internxt, line 21: This vi.fn is unnecesary, you are already mocking run later

src/context/virtual-drive/files/application/update/FilesPlaceholderUpdater.ts
```diff
@@ -25,6 +26,24 @@ export class FilesPlaceholderUpdater {
...
+  private async hasToBeUpdatedIdentity(local: File, remote: File): Promise<boolean> {
+    const localExists = local.status.is(FileStatuses.EXISTS);
+    const remoteExists = remote.status.is(FileStatuses.EXISTS);
+
+    const systemFileidentity = await this.localFileSystem.getFileIdentity(local.path);
+    const remoteIdentity = remote.placeholderId;
+
+    const isDifferentIdentity = systemFileidentity !== remoteIdentity;
+
+    logger.debug({
```
dajimenezriv-internxt, line 38: I would put only this log when we have to update the identity, otherwise this is going to print so many many logs, and they are just necessary when we change the identity.

src/context/virtual-drive/files/application/update/FilesPlaceholderUpdater.ts
```diff
@@ -54,6 +75,10 @@ export class FilesPlaceholderUpdater {
       return;
     }
 
+    if (await this.hasToBeUpdatedIdentity(local, remote)) {
```
dajimenezriv-internxt, line 78: Put a comment about why we are doing this. Explain that we were using contentsId, but we needed to change because it was dynamic and now we use the uuid. Something like this, otherwise we won't understand this method in the future.

src/apps/sync-engine/BindingManager.ts
```diff
@@ -103,10 +103,10 @@ export class BindingsManager {
         Logger.debug('Path received from callback', absolutePath);
         await this.controllers.addFile.execute(absolutePath);
       },
-      fetchDataCallback: (contentsId: FilePlaceholderId, callback: CallbackDownload) =>
+      fetchDataCallback: async (filePlaceholderId: FilePlaceholderId, callback: CallbackDownload) =>
```
dajimenezriv-internxt, line 106: This async is unnecesary

tests/context/virtual-drive/contents/application/ContentsDownloader.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable require-await */
```
dajimenezriv-internxt, line 1: In this file, instead of remove the await and disable eslint, you can just do Promise.resolve('C:/temp')

src/context/virtual-drive/folders/application/UpdatePlaceholderFolder.ts
```diff
@@ -48,7 +48,7 @@ export class FolderPlaceholderUpdater {
 
   private canWrite(win32AbsolutePath: string) {
     try {
-      fs.access(win32AbsolutePath, FsConstants.R_OK | FsConstants.W_OK);
+      void fs.access(win32AbsolutePath, FsConstants.R_OK | FsConstants.W_OK);
```
dajimenezriv-internxt, line 51: This is never going to go inside the catch, so it doesn't make sense. If the function cannot be async then use fs sync

src/context/virtual-drive/files/infrastructure/NodeWinLocalFileSystem.ts
```diff
@@ -35,8 +35,11 @@ export class NodeWinLocalFileSystem {
     });
   }
 
-  getFileIdentity(path: File['path']): string {
-    return this.virtualDrive.getFileIdentity({ path });
+  async getFileIdentity(path: File['path']): Promise<string> {
+    // v2.5.2 Jonathan Daniel
```
dajimenezriv-internxt, line 39: Why needs to be awaited and why not declare the function as async in virtual-drive which is the root of this function?

src/context/virtual-drive/files/application/update/FilesPlaceholderUpdater.ts
```diff
@@ -47,23 +65,31 @@ export class FilesPlaceholderUpdater {
...
       if (remote.status.is(FileStatuses.EXISTS)) {
-        await this.repository.add(remote);
-        await this.localFileSystem.createPlaceHolder(remote);
+        this.repository.add(remote);
+        this.localFileSystem.createPlaceHolder(remote);
       }
       return;
     }
 
+    // v2.5.2 Jonathan Daniel
```
dajimenezriv-internxt, line 74: To discuss, are we going to use multi line comments or single line comments? So everyone do the same

src/context/virtual-drive/files/application/update/FilesPlaceholderUpdater.test.ts
```diff
@@ -0,0 +1,270 @@
...
+
+const mockFileIdProvider = mockDeep<LocalFileIdProvider>();
+
+const mockEventHistory = mockDeep<InMemoryEventRepository>();
+
+describe('FilesPlaceholderUpdater', () => {
+  let updater: FilesPlaceholderUpdater;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
```
dajimenezriv-internxt, line 30: With vi.clearAllMocks, you are already clearing all mocks, you don't need to construct updater everytime.

src/context/virtual-drive/files/application/update/FilesPlaceholderUpdater.test.ts
```diff
@@ -0,0 +1,270 @@
...
+    updater = new FilesPlaceholderUpdater(mockRepository, mockLocalFileSystem, mockPathConverter, mockFileIdProvider, mockEventHistory);
+  });
+
+  afterAll(() => {
+    vi.restoreAllMocks();
```
dajimenezriv-internxt, line 35: Also, you just need to clear/restore mocks or before or after, not both times.

src/context/virtual-drive/files/infrastructure/NodeWinLocalFileSystem.ts
ArceDanielShok: Yes, that has to be corrected from node win

src/context/virtual-drive/folders/application/UpdatePlaceholderFolder.ts
```diff
@@ -127,10 +127,8 @@ export class FolderPlaceholderUpdater {
       return;
     }
 
-    if (await this.hasToBeCreated(remote)) {
```
dajimenezriv-internxt, line 130: Why do you have remove this if?

dajimenezriv-internxt (approved):

PR #706, opened 2025-04-10
Removal of Unnecessary Logs and Addition of Update Placeholder on Startup

ArceDanielShok:
This Pull Request introduces improvements to log clarity and placeholder initialization.

- Removal of Unnecessary Logs:
  - Logs that did not provide relevant information or were redundant have been identified and removed. This makes it easier to read and analyze important logs.
- Addition of Update Placeholder on Startup:
  - The functionality to update placeholders during the application startup process has been added. This ensures that placeholders are in a consistent and updated state from the beginning.


src/apps/main/logger.ts
```diff
@@ -4,7 +4,7 @@ import log from 'electron-log';
 
 log.initialize();
 
-log.transports.file.maxSize = 1048576 * 150; // 150MB
+log.transports.file.maxSize = 150 * 1024 * 1024; // 120MB
```
dajimenezriv-internxt, line 7: I think the comment should be 150MB

src/apps/sync-engine/BindingManager.ts
```diff
@@ -174,6 +174,14 @@ export class BindingsManager {
     await this.container.virtualDrive.connectSyncRoot();
 
     await this.load();
+    /**
+     * Jonathan Arce v2.5.1
+     * This is a quick fix to populate and update the placeholders
+     * of the files that already exist in the local DB as soon as the sync engine starts.
+     * This should be a continuous process, fetching and updating the placeholders.
+     */
```
dajimenezriv-internxt, line 182: I don't agree with the comment. I wouldn't say it's a quick fix. I think now, with update and polling, it should be the solution. It can happen that we had some files in our local db and we want to populate before starting fetching everything new and wait for that. I think, we should have this, then creating/updating/deleting placeholders as we fetch and then the last update and polling when that finish.

PR #719, opened 2025-04-14
Implementation of Retry Logic in Usage Service and Refactoring

ArceDanielShok:
This Pull Request introduces improvements to the resilience of the usage service and performs code cleanup.

- Retry Logic in Usage Service:
  - A retry logic has been implemented in the usage service. In case of failures, the service will attempt the operation again after a short sleep period, thus avoiding overloading the system with immediate requests after an error.
- Test Addition:
  - Unit tests have been added for the new retry logic in the usage service, ensuring its correct functionality and robustness.
- Removal of Unused Code:
  - Code that was no longer used in the project has been identified and removed, simplifying the codebase and improving its maintainability.


tests/issues.e2e.ts
```diff
@@ -19,7 +19,7 @@ test.describe('process issues', () => {
   const addSyncErrors = async (errors: Array<ProcessErrorName>) => {
     const emitEvents = errors.map(createSyncError).map((error) => ipcMainEmit(electronApp, 'SYNC_INFO_UPDATE', error));
 
-    return Promise.all(emitEvents);
+    return await Promise.all(emitEvents);
```
dajimenezriv-internxt, line 22: Here instead of doing an wait in the Promise.all, we can just remove the async in the function, otherwise it's like awaiting two promises.

src/infra/drive-server-wip/services/workspaces.service.ts
```diff
@@ -4,7 +4,7 @@ import { clientWrapper } from '../in/client-wrapper.service';
 export async function getWorkspaces() {
   const promise = client.GET('/workspaces');
 
-  return clientWrapper({
+  return await clientWrapper({
```
dajimenezriv-internxt, line 7: And for all client wrappers it's the same, it's better to remove the async instead of adding await.

src/apps/main/usage/service.ts
```diff
@@ -1,20 +1,41 @@
+import { sleep } from '../util';
 import { Usage } from './Usage';
-import { driveServerWipModule } from '@/infra/drive-server-wip/drive-server-wip.module';
+import { DriveServerWipModule } from '@/infra/drive-server-wip/drive-server-wip.module';
 
 const INFINITE_SPACE_TRHESHOLD = 108851651149824;
 const OFFER_UPGRADE_TRHESHOLD = 2199023255552;
 
 export class UserUsageService {
+  constructor(private driveServerWipModule = new DriveServerWipModule()) {}
```
dajimenezriv-internxt, line 9: Do you think here is better to instance the class again? I was usually doing:
private readonly driveServerWip = driveServerWipModule, so then we have like just one singleton

ArceDanielShok: I think it's irrelevant, but I can make the change

dajimenezriv-internxt (approved):

PR #723, opened 2025-04-15
Update node-win version, Refactor nde-win Method Parameters, and Fix PersistQueue Path Issue

ArceDanielShok:
This Pull Request introduces improvements to Windows compatibility, method structure, and internal path management.

- Update node-win version
- Modification of Parameters in node win Methods
- Resolution of Issue in Obtaining PersistQueue Path:
  - A minor error affecting how the persistQueue path was obtained has been resolved. This fix ensures that the application accesses the correct location of the persistence queue, which is crucial for task management and data persistence.

### DIFF
--- a/src/apps/sync-engine/BindingManager.ts
+++ b/src/apps/sync-engine/BindingManager.ts
@@ -265,7 +265,15 @@ export class BindingsManager {
 
       const placeholders = this.container.virtualDrive.getPlaceholderWithStatePending();
 
-      logger.debug({ msg: '[SYNC ENGINE] Update and check placeholders', workspaceId, total: placeholders.length, placeholders });
+      logger.debug({
+        msg: 'Update and check placeholders',
+        workspaceId,
+        total: placeholders.length,
+        placeholders,
+        attributes: {
+          tag: 'SYNC-ENGINE',
+        },
+      });
 
       if (placeholders.length === 0) {
         ipcRendererSyncEngine.send('CHANGE_SYNC_STATUS', workspaceId, 'SYNCED');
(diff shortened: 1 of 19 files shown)


src/apps/sync-engine/BindingManager.ts
```diff
@@ -190,23 +194,16 @@ export class BindingsManager {
...
-      onTaskSuccess: async () => {
-        //ipcRendererSyncEngine.send('SYNCED', getConfig().workspaceId)
-      },
-      onTaskProcessing: async () => {
-        // ipcRendererSyncEngine.send('SYNCING', getConfig().workspaceId);
-      },
-    };
-
-    const persistQueueManager: string = configStore.get('persistQueueManagerPath');
+    const PATHS = await ipcRenderer.invoke('get-paths', this.PROVIDER_NAME);
```
dajimenezriv-internxt, line 197: Why you provide an argument?

src/apps/sync-engine/BindingManager.ts
```diff
@@ -268,6 +265,8 @@ export class BindingsManager {
 
       const placeholders = this.container.virtualDrive.getPlaceholderWithStatePending();
 
+      logger.debug({ msg: '[SYNC ENGINE] Update and check placeholders', workspaceId, total: placeholders.length, placeholders });
```
dajimenezriv-internxt, line 268: Here, I've added the tag field, so we can use:
logger.debug({ tag: "SYNC ENGINE" msg: "Update and check placeholders" }).
The proposal is to print it but also add it to the attributes in the logger for later process in grafana.
What do you think about the tag? Or do you prefer to use [SYNC ENGINE] inside the msg?

PR #734, opened 2025-04-21
Added Debounce to updateUsage to Prevent Multiple Calls

ArceDanielShok:
This Pull Request introduces an efficiency improvement to the updateUsage method.

- Implementation of Debounce in updateUsage:
  - A debounce has been added to the updateUsage method. This means that if calls are made to this method quickly and consecutively, only the last call within a specified time period will be executed. This helps prevent multiple unnecessary calls to the service, optimizing performance and reducing the load on the system.

### DIFF
--- a/src/apps/renderer/hooks/useUsage.tsx
+++ b/src/apps/renderer/hooks/useUsage.tsx
@@ -1,10 +1,12 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useState, useCallback } from 'react';
 import { Usage } from '../../main/usage/Usage';
+import { debounce } from 'lodash';
 
 export default function useUsage() {
   const [usage, setUsage] = useState<Usage>();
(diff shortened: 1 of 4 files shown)


src/apps/renderer/hooks/useUsage.tsx
```diff
@@ -18,14 +20,18 @@ export default function useUsage() {
...
-  }
+  }, []);
+
+  const debouncedUpdateUsage = useCallback(debounce(updateUsage, 500), []);
 
   useEffect(() => {
     setStatus('loading');
-    updateUsage();
-    const listener = window.electron.onRemoteChanges(updateUsage);
+    // The 'void' operator explicitly ignores the Promise returned by this async function,
```
dajimenezriv-internxt, line 29: I would remove this comment, because probably we are going to use void in many places. Also, is someone doesn't understand what it does it's easy to find the solution in google or chatgpt.

dajimenezriv-internxt (approved):

PR #735, opened 2025-04-23
Create placeholders as we fetch

dajimenezriv-internxt:
## What

1. Improve logger so it only prints workspaceId or tag if they are defined (we now exclude empty workspaceId, so that the default drive doesn't include workspaceId: '' always).
2. Create placeholders as we fetch. We have created an in memory store, that saves all folders so we can retrieve the path recursively. Also, as we fetch we send an IPC signal to the worker to create the placeholder.
3. Add to the CONTRIBUTING file how we are going to make comments.

### DIFF
--- a/src/apps/main/remote-sync/handlers.ts
+++ b/src/apps/main/remote-sync/handlers.ts
@@ -16,6 +16,8 @@ import Queue from '@/apps/shared/Queue/Queue';
 import { driveFilesCollection, driveFoldersCollection, getRemoteSyncManager, remoteSyncManagers } from './store';
 import { TWorkerConfig } from '../background-processes/sync-engine/store';
 import { getSyncStatus } from './services/broadcast-sync-status';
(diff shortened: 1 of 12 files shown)


src/apps/main/remote-sync/handlers.ts
```diff
@@ -16,6 +16,8 @@ import Queue from '@/apps/shared/Queue/Queue';
 import { driveFilesCollection, driveFoldersCollection, getRemoteSyncManager, remoteSyncManagers } from './store';
 import { TWorkerConfig } from '../background-processes/sync-engine/store';
 import { getSyncStatus } from './services/broadcast-sync-status';
+import { FolderStore, folderStore } from './folders/folder-store';
+import { Folder } from '@/context/virtual-drive/folders/domain/Folder';
```
ArceDanielShok, line 20: Folder and folderStore are not being used, we could delete them

ArceDanielShok (approved):

PR #750, opened 2025-05-07
Validate windows name

dajimenezriv-internxt:
## What

First step to introduce validation for windows names.

1. Remove ipc SYNC_PROBLEM and BACKUP_ISSUE that were not used.
2. Remove callbacks notifyFileAddedCallback and notifyMessageCallback, since they were not used in node-win.
3. Validate windows name on create file and folder placeholder and create an issue.


src/context/virtual-drive/items/validate-windows-name.ts
```diff
@@ -0,0 +1,30 @@
...
+   * v2.5.3 Daniel Jiménez
+   * These characters are invalid in windows paths.
+   */
+  const forbiddenPattern = /[<>:"/\\|?*]/;
+  const isValid = !forbiddenPattern.test(name);
+
+  if (!isValid) {
+    logger.debug({ msg: 'Invalid windows name', name, type });
+
+    ipcRendererSyncEngine.send('SYNC_INFO_UPDATE', {
```
dajimenezriv-internxt, line 20: This is temporal, I'm going to change a bit the structure of the issue since now we are asking for many things and then only printing the name.

src/apps/sync-engine/BindingManager.ts
```diff
@@ -91,37 +90,13 @@ export class BindingsManager {
           Logger.error('Error during rename or move operation', error);
         }
       },
-      notifyFileAddedCallback: async (absolutePath: string) => {
```
dajimenezriv-internxt, line 94: This callback and notifyMessageCallback were defined in node-win as extra callbacks, but they were never used.

ArceDanielShok (approved):

PR #754, opened 2025-05-09
Simplify remote items generator

dajimenezriv-internxt:
## What

1. Simplify RemoteItemsGenerator so we do not need services, it becomes just one function and also move the traverser and remote tree domain to only backups (they are no longer used in sync engine).
2. Change sync manually to perform the sync directly instead of doing the debounce.

The objective is to simplify backups so we can move it ouside a renderer process.


src/apps/main/remote-sync/handlers.ts
```diff
@@ -167,7 +167,7 @@ ipcMain.handle('get-remote-sync-status', () => {
 
 ipcMain.handle('SYNC_MANUALLY', async () => {
   Logger.info('[Manual Sync] Received manual sync event');
-  await debouncedSynchronization();
+  await updateAllRemoteSync();
```
dajimenezriv-internxt, line 170: This is because when we are doing the sync manually we need to wait 5s until it starts because of the debounce. I think that in manually we can leave it normal, because the debounce is more useful for notifications.

ArceDanielShok (approved):

PR #755, opened 2025-05-09
 Download Refinement, Code Cleanup, and Items Component Improvement

ArceDanielShok:
This Pull Request introduces several improvements focused on the download experience and code organization.

- Removal of Decrypting Message in Download:
  - The message indicating the "decrypting" phase during download has been removed. This simplifies the information displayed to the user during this process.
- Items Component Improvement:
  - Improvements have been made to the Items component.


src/apps/renderer/pages/Widget/Item.tsx
```diff
@@ -15,36 +48,9 @@ export function Item({
...
-  } else if (action === 'DELETED') {
-    description = translate('widget.body.activity.operation.deleted');
-  } else if (action === 'RENAMING') {
-    description = translate('widget.body.activity.operation.renaming');
-  } else if (action === 'RENAMED') {
-    description = translate('widget.body.activity.operation.renamed');
-  } else if (errorName) {
-    description = shortMessages[errorName];
-  }
+  const description = getDescription({ action, errorName });
```
dajimenezriv-internxt, line 53: You are not passing here progress, but anyway, I'm not sure if this new form is more redeable. Action is always a string, so it's never going to enter into errorMessage in the getDescription function.

dajimenezriv-internxt: I mean, maybe is more redeable, but we are losing the type of action when sending it to getDescription for example.

PR #781, opened 2025-05-24
Remove server file and server folder

dajimenezriv-internxt:
## What

Remove the intermidate server file and server folder types.


src/apps/backups/remote-tree/traverser.ts
```diff
@@ -58,16 +61,20 @@ export class Traverser {
...
-      const name = createRelativePath(currentFolder.path, decryptedName);
+        const relativePath = createRelativePath(currentFolder.path, decryptedName);
 
-      try {
-        const folder = createFolderFromServerFolder(serverFolder, name);
+        const folder = Folder.from({
+          ...serverFolder,
+          parentUuid: serverFolder.parentUuid || null,
+          path: relativePath,
+        });
```
AlexisMora, line 77: this could be its own private function inside the traverser, this way the code is more legible

src/apps/backups/remote-tree/traverser.ts
```diff
@@ -39,17 +36,23 @@ export class Traverser {
...
+
+        const relativePath = createRelativePath(currentFolder.path, decryptedName);
+
+        const file = File.from({
+          ...serverFile,
+          path: relativePath,
+          contentsId: serverFile.fileId,
+          size: Number(serverFile.size),
+        });
+
```
AlexisMora, line 55: this as well

src/context/virtual-drive/items/application/Traverser.ts
```diff
@@ -1,16 +1,15 @@
-import { ServerFile, ServerFileStatus } from '../../../shared/domain/ServerFile';
-import { ServerFolder, ServerFolderStatus } from '../../../shared/domain/ServerFolder';
-import { createFileFromServerFile } from '../../files/application/FileCreatorFromServerFile';
 import { Folder } from '../../folders/domain/Folder';
```
AlexisMora, line 1: What would be the main difference from this traverser and the src/backups/remote-tree/traverser?

src/context/virtual-drive/items/application/Traverser.ts
```diff
@@ -69,18 +73,23 @@ export class Traverser {
...
+          ...serverFolder,
+          parentId: serverFolder.parentId || null,
+          parentUuid: serverFolder.parentUuid || null,
+          path: relativePath,
+        });
+
+        if (serverFolder.status === 'DELETED' || serverFolder.status === 'TRASHED') {
           tree.trashedFolders.push(folder);
         } else {
           tree.folders.push(folder);
```
AlexisMora, line 95: I have 2 questions: 
1.  i see in the other traverser that we are running some ` Sentry.captureException(error);` but not here, should we add it?
2. we are Catching the  exception when creating a folder but we are not stopping, should we send the error and stop the traverse function?

dajimenezriv-internxt: We shouldn't stop the traverser. I think we are not really using Sentry (we are already stopping using it in the rest of the application (we will need to configure loki and grafana)). We should add an error to backupIssues or syncIssues (work in progress), I'm going to add a comment.

src/apps/backups/remote-tree/traverser.ts
AlexisMora: thats perfect

AlexisMora (approved):

PR #784, opened 2025-05-25
Remove remote tree

dajimenezriv-internxt:


src/apps/main/auth/handlers.ts
```diff
@@ -93,5 +93,5 @@ export function setupAuthIpcHandlers() {
 async function emitUserLoggedIn() {
   eventBus.emit('USER_LOGGED_IN');
   cleanAndStartRemoteNotifications();
-  await initSyncEngine();
+  // await initSyncEngine();
```
AlexisMora: if you are not going to use it delete this commented code

src/infra/drive-server-wip/services/files.service.ts
```diff
@@ -85,22 +86,23 @@ export class FilesService {
...
+          msg: 'Replace file request was not successful',
+          context,
+          attributes: {
+            method: 'PUT',
+            endpoint: '/files/{uuid}',
+          },
         },
-      },
-    });
+      });
```
AlexisMora, line 103: why are you wrapping it like this?

dajimenezriv-internxt: We were doing the retryWrapper outside the infra, however it's boilerplate and it will lead to printing two logs for each error. Also, since there are required requests, it's better to have the retry wrapper always.

PR #791, opened 2025-05-27
[BR-1233] WebSocket error connection issue

AlexisMora:
## Improvements
- Now, when the websocket emits the event 'connect_error', we generate an issue regarding the connection error that will be displayed on Issues > General
- When the websocket emits the event connect, we will remove the issue if there is already one

## Type Migration
- Migrated from older `AppError` and `AppIssue` types on the frontend to the new `GeneralIssueError` and `GeneralIssue` types, those will be used both in the main and the renderer
- Updated related functions to work with the new types

## Others
Separated GeneralIssueAccordion into its own file for better readability

##Preview
Captura de pantalla 2025-05-27 164541

### DIFF
--- a/examples/handlers/handle-add.ts
+++ b/examples/handlers/handle-add.ts
@@ -4,7 +4,7 @@ import { v4 } from 'uuid';
 
 import { QueueItem } from '@/node-win/queue/queueManager';
 
-export const handleAdd = async (task: QueueItem) => {
+export const handleAdd = (task: QueueItem) => {
   try {
     logger.debug({ msg: 'handleAdd', task });
     const id = task.isFolder ? v4() : addInfoItem(task.path);
--- a/examples/handlers/handle-change-size.ts
+++ b/examples/handlers/handle-change-size.ts
@@ -3,7 +3,7 @@ import { v4 } from 'uuid';
 
 import { QueueItem } from '@/node-win/queue/queueManager';
 
-export const handleChangeSize = async (task: QueueItem) => {
+export const handleChangeSize = (task: QueueItem) => {
   try {
     logger.debug({ msg: 'handleChangeSize', path: task.path });
     const id = v4();
--- a/examples/handlers/handle-dehydrate.ts
+++ b/examples/handlers/handle-dehydrate.ts
@@ -2,7 +2,7 @@ import { drive, logger } from 'examples/drive';
 
 import { QueueItem } from '@/node-win/queue/queueManager';
 
-export const handleDehydrate = async (task: QueueItem) => {
+export const handleDehydrate = (task: QueueItem) => {
   try {
     logger.debug({ msg: 'handleDehydrate', path: task.path });
     drive.dehydrateFile({
--- a/src/apps/backups/Backups.ts
+++ b/src/apps/backups/Backups.ts
@@ -30,7 +30,7 @@ export class Backup {
 
   backed = 0;
 
-  async run(tracker: BackupsProcessTracker, context: BackupsContext): Promise<DriveDesktopError | undefined> {
+  async run(tracker: BackupsProcessTracker, context: BackupsContext) {
     const local = await LocalTreeBuilder.run(context.pathname as AbsolutePath);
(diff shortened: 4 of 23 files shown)


src/apps/main/background-processes/issues.ts
```diff
@@ -63,7 +80,37 @@ export function clearBackupsIssues() {
   onIssuesChanged();
 }
 
+export function getGeneralIssues(): Array<GeneralIssue> {
+  return issues.filter(isGeneralIssue);
+}
+
+export function onGeneralIssuesChanged() {
+  broadcastToWindows('general-issues-changed', getGeneralIssues());
```
AlexisMora, line 88: As we talked earlier @dajimenezriv ,this is a middle step between what we had and we are going to, the idea is to move all of the general issues into the same list and manage them there, also having one unique customHook on the frontend for managing the issues

src/apps/renderer/pages/Issues/GeneralIssuesByErrorAccordion.tsx
```diff
@@ -1,16 +1,22 @@
...
+type AppIssueElementProps = {
+  issues: Array<GeneralIssue>;
+};
+/*
+ * v2.5.3 Alexis Mora
+ * This function has to be here because it we add it to issues
+ * the app tries to import it from the main process
+ * which is not possible
+ * */
+export function groupGeneralIssuesByErrorName(issues: GeneralIssue[]) {
```
AlexisMora, line 16: I added the comment on the code but maybe its helpful as well to add context here:

I was getting Webpack errors when trying to import this method from issues.ts. The reason is that importing the method causes webpack to include the entire file. And because that file imports Node/Electron-specific modules (like ipcMain), it breaks the build in the renderer process.

We could mode this and other functions onto a separate file so that we can benefit of having everything centralized

dajimenezriv: I think this comment isn't necessary. Technically we don't add a comment in each function that we declare in the renderer because we cannot create it in the main.

src/apps/main/background-processes/issues.ts
```diff
@@ -63,7 +80,37 @@ export function clearBackupsIssues() {
   onIssuesChanged();
 }
 
+export function getGeneralIssues(): Array<GeneralIssue> {
```
dajimenezriv, line 83: Also, we agreed on not adding return types when, you can see it in CONTRIBUTION.md. But we can discuss it if you thing providing types is better. Basically the reason was:
- less boilerplate.
- you don't have to mantain it.
- you don't "hide" properties because the explicit type is more narrow.
- you don't have to name types that are innecesary.

src/apps/main/background-processes/issues.ts
```diff
@@ -22,27 +22,44 @@ export type BackupsIssue = {
...
+  name: string;
+  error: GeneralIssueError;
+};
 
-function onIssuesChanged() {
-  broadcastToWindows('issues-changed', issues);
+export type Issue = SyncIssue | BackupsIssue | GeneralIssue;
+
+export function isGeneralIssue(issue: Issue): issue is GeneralIssue {
+  return issue.tab === 'general';
```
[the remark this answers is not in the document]
AlexisMora: What complexity? This is just a typeguard to infer the type

src/apps/renderer/pages/Issues/GeneralIssueAccordion.tsx
```diff
@@ -0,0 +1,53 @@
...
+              </div>
+            ))}
+          </motion.div>
+        )}
+      </AnimatePresence>
+    </>
+  );
+};
+
+export default GeneralIssueAccordion;
```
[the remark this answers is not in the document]
AlexisMora: I just dont see the benefit right now in react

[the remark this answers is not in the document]
AlexisMora: Ah yes I understand it now, we can change it

PR #808, opened 2025-06-03
Move on add to function and type some ipc

dajimenezriv-internxt:
## What

1. Move onAdd to a function instead of being a class and separate the test. The point is to split the watcher.test file into different files since it's too big.
2. Add some typed ipcs.


src/node-win/watcher/events/on-add.service.ts
```diff
@@ -4,40 +4,44 @@ import { Watcher } from '../watcher';
 import { PinState, SyncState } from '@/node-win/types/placeholder.type';
```
dajimenezriv, line 4: It's exactly the same content, nothing changed.

src/node-win/watcher/events/watcher-on-add.test.ts
```diff
@@ -0,0 +1,102 @@
+import { mkdir, writeFile } from 'fs/promises';
```
dajimenezriv, line 1: Same test as before, just moved

PR #809, opened 2025-06-04
Add infra test create placeholder

dajimenezriv-internxt:
## What

Add first integration test mocking the upload and the creation of the file in drive-server-wip. This test starts the BindingManager and creates a file in a temporal drive.


src/context/virtual-drive/contents/application/ContentsUploader.ts
```diff
@@ -69,6 +68,7 @@ export class ContentsUploader {
       this.registerEvents(uploader, contents);
 
       const contentsId = await uploader.upload(contents.stream, contents.size);
+      console.log('🚀 ~ ContentsUploader ~ run ~ contentsId:', contentsId);
```
AlexisMora: Should this be here?

src/apps/sync-engine/tests/create-placeholder.infra.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    const bindingManager = new BindingsManager(container);
+
+    // When
+    await bindingManager.start();
+    await bindingManager.watch();
+
+    await sleep(100);
+    await writeFile(file, 'content');
+    await sleep(10000);
```
AlexisMora, line 105: Why do we need the sleep functions?

AlexisMora (approved):

vitest.config.infra.mts
```diff
@@ -0,0 +1,20 @@
+import tsconfigPaths from 'vite-tsconfig-paths';
```
dajimenezriv, line 1: Infra tests are going to have extension infra.test.ts, since they are slower and it's not needed to run at the same time as unit tests. In local we can run unit tests only, in the pipeline both will run.

AlexisMora: Thanks for the info, I forgot to ask about this!

src/context/virtual-drive/contents/application/ContentsUploader.ts
dajimenezriv: No, already deleted

src/apps/sync-engine/tests/create-placeholder.infra.test.ts
dajimenezriv: The first sleep is to wait until the watcher starts listening. When we write a file into the temporal drive, it generates an event in the watcher in another promise and then starts all the process of creating the placeholder, uploading the content to bucket (mocked) and creating the file in drive-server-wip (mocked). This takes some time, probably around 5s, but just in case, since this is an infra test an speed is not so necessary.

PR #811, opened 2025-06-04
Electron log max size to 1GB

dajimenezriv-internxt:


AlexisMora:
What would be the default behavior on the logger? Will it stop logging or will be deleting the oldest logs?

Edit: Looking at the [documentation ](https://github.com/megahertz/electron-log/blob/master/docs/transports/file.md#maxsize-number:~:text=Maximum%20size%20of%20a%20log%20file%20in%20bytes.%20When%20a%20log%20file%20exceeds%20this%20limit%2C%20it%20will%20be%20moved%20to%20%7Bfile%20name%7D.old.log%20file.%20You%20can%20set%20it%20to%200%20to%20disable%20log%20rotation.)i see the following:
>**maxSize {number}**
>Default: 1048576 (1mb)
>Maximum size of a log file in bytes. When a log file exceeds this limit, it will be moved to {file name}.old.log file. You can set it >to 0 to disable log rotation.

AlexisMora:
> What would be the default behavior on the logger? Will it stop logging or will be deleting the oldest logs?

AlexisMora:
I did some research and confirmed that electron-log handles file size limits by renaming the current log file to log.old.log once it exceeds the configured maxSize. A new log.log file is then created and logging continues there. This means that at most, two files will be kept: the current and the previous one, capping disk usage to around 2 GB.

From what I have seen, electron-log doesn’t support multi-file rotation or automatic cleanup beyond the .old file. If we ever need more advanced log rotation or retention strategies (e.g. keeping N rotated files, time-based rotation), we might consider switching to a more feature-rich solution like winston with winston-daily-rotate-file.

AlexisMora (approved):

PR #814, opened 2025-06-06
fix: add more error handling to the getOrCreateDevice and dont create a device when we recieve an error from the infra layer

AlexisMora:


src/apps/main/device/service.ts
```diff
@@ -49,78 +49,114 @@ export const addUnknownDeviceIssue = (error: Error) => {
...
+ * Creates a new device with a unique name
+ * @returns The an object with the created device or
+ * an object with the error if device creation fails after multiple attempts
+ */
+export async function createUniqueDevice() {
+  const baseName = os.hostname();
+  const nameVariants = [
+    baseName,
+    ...Array.from({ length: 10 }, (_, i) => `${baseName} (${i + 1})`),
+    `${baseName} (${new Date().valueOf() % 1000})`,
```
dajimenezriv: Isn't it better to just continue iterating until 1000 instead of using random numbers?

src/apps/shared/logger/logger.ts
```diff
@@ -3,7 +3,7 @@ import { getUser } from '@/apps/main/auth/service';
 import ElectronLog from 'electron-log';
 import { paths } from '../HttpClient/schema';
 
-type TTag = 'AUTH' | 'BACKUPS' | 'SYNC-ENGINE' | 'ANTIVIRUS' | 'NODE-WIN';
+type TTag = 'AUTH' | 'BACKUPS' | 'SYNC-ENGINE' | 'ANTIVIRUS' | 'NODE-WIN' | 'DEVICE';
```
dajimenezriv: Technically DEVICE is for the BACKUPS feature, so I'm not sure if we need another one. It would be "AUTH" | "BACKUPS" | "SYNC-ENGINE" | "ANTIVIRUS". (Because "NODE-WIN" it's "SYNC-ENGINE", but we still need to migrate the tag)

src/apps/shared/logger/logger.ts
```diff
@@ -63,11 +63,16 @@ export class LoggerService {
...
+   * v2.5.3
+   * Alexis Mora
+   * We could enhance the error handling by passing specific error types like NotFoundError, BadRequestError, etc.
+   * */
   error(rawBody: TLoggerBody) {
     const { body } = this.prepareBody('error', rawBody);
     ElectronLog.debug(body);
     ElectronLog.info(body);
-    return new Error(rawBody.msg);
+    return new Error(rawBody.msg, { cause: rawBody.exc });
```
dajimenezriv, line 72: I like the idea, I think that the comment is not necessary and also you have to include this in the fatal function

src/apps/main/device/service.ts
AlexisMora: Yes, even though this is the current functionality, I feel like iterating to a certain number is more predictable rather than just a random number

src/apps/shared/logger/logger.ts
AlexisMora: I always thought that device and backups should be a separate domain even though we always used device for the backups

src/apps/shared/logger/logger.ts
AlexisMora: You are right, done

src/apps/shared/logger/logger.ts
ArceDanielShok: I also think that backups and devices should be in a single section

PR #816, opened 2025-06-10
VirtualDriveStatus was never NOT_FOUND

dajimenezriv-internxt:
## What

1. Remove VirtualDriveStatus since the status VirtualDriveStatus.NOT_FOUND was never used so the code wasn't doing anything.
2. Add `setDefaultConfig({})` at the startup, otherwise if a backup was started in schedule it didn't have the mnemonic yet.


src/apps/main/main.ts
```diff
@@ -107,6 +107,8 @@ if (process.env.NODE_ENV === 'development') {
 app
   .whenReady()
   .then(async () => {
+    setDefaultConfig({});
```
AlexisMora, line 110: I  have seen that the setDefaultConfig is being set on 
`eventBus.on('USER_LOGGED_IN', async () => {
  try {
    setDefaultConfig({});
    ....
`
since you are already setting it here, is it really necessary?
Also, we could add `{}` as default value on the function since it seems like the initial value, what do you think?

src/apps/main/antivirus/Antivirus.ts
```diff
@@ -1,15 +1,16 @@
...
+import { cwd } from 'process';
 
 export interface SelectedItemToScanProps {
   path: string;
   itemName: string;
   isDirectory: boolean;
 }
 
-const RESOURCES_PATH = app.isPackaged ? path.join(process.resourcesPath, 'clamAV') : path.join(__dirname, '../../../../clamAV');
+const RESOURCES_PATH = app.isPackaged ? path.join(process.resourcesPath, 'clamAV') : path.join(cwd(), 'clamAV');
```
AlexisMora, line 13: What is the difference between using __dirname and cwd() ?

dajimenezriv: __dirname gives the path of the current file, in this case Antivirus.ts, however, if we move the file this stops working. cwd() instead gives the path from where we are executing the app, the root folder.

src/apps/main/main.ts
dajimenezriv: Yes, it's necessary, at least for now. In this case is for backups.
We have the function `setupBackups` in the beginning of the code which sets the listeners for the ipc for backups and also starts the scheduled backup if necessary. However, USER_LOGGED_IN happens after this, so we need to set the user config at the beginning to prevent this. Probably we should move the ipc handlers into a different function but it's not easy because there is dependent code between the handlers and scheduled backup.

src/apps/main/antivirus/Antivirus.ts
AlexisMora: Okay understood

src/apps/main/main.ts
AlexisMora: I meant to use  setDefaultConfig({}) on 'USER_LOGGED_IN', since we are already setting it on this change that you just introduced

PR #817, opened 2025-06-11
Join get files from directory

dajimenezriv-internxt:
## What

This is a more optimized way of getting all files and folders recursively, but the output should be the same as the file getFilesFromDirectory. However, previously there were two different functions for counting and for retrieving all files, now it's just one so we ensure that we do not count less files that we are scanning or viceversa.

### DIFF
--- a/src/apps/main/antivirus/utils/initializeAntivirus.ts
+++ b/src/apps/main/antivirus/utils/initializeAntivirus.ts
@@ -35,7 +35,7 @@ async function initializeClamAV() {
       await clamAVServer.waitForClamd();
 
       logger.debug({ msg: '[INITIALIZING CLAM AV] ClamAV is running. Scheduling daily scan...' });
-      // scheduleDailyScan();
+      scheduleDailyScan();
 
       isClamAVRunning = true;
       clamAVInitializationPromise = null;


src/apps/main/antivirus/utils/get-files-from-directory.ts
```diff
@@ -0,0 +1,47 @@
+import { readdir } from 'fs/promises';
```
dajimenezriv, line 1: This is a more optimized way of getting all files and folders recursively, but the output should be the same as the file `getFilesFromDirectory`. However, previously there were two different functions for counting and for retrieving all files, now it's just one so we ensure that we do not count less files that we are scanning or viceversa.

src/apps/main/antivirus/ManualSystemScan.ts
```diff
@@ -174,53 +175,51 @@ class ManualSystemScan {
...
+    logger.debug({
+      tag: 'ANTIVIRUS',
+      msg: 'Starting manual scan',
+      pathNames,
+    });
+
     console.time('manual-scan');
+
     try {
-      if (!pathNames) {
```
dajimenezriv, line 179: If no pathNames then pathNames is userSystemPath

src/apps/main/antivirus/ManualSystemScan.ts
```diff
@@ -174,53 +175,51 @@ class ManualSystemScan {
...
+      for (const p of pathNames) {
+        await getFilesFromDirectory({ filePaths, folder: p });
+      }
 
-        await filesPromise;
-      } else {
-        let pathsToScan: string[] = [];
-        if (pathNames && pathNames.length > 0) {
-          pathsToScan = pathNames;
-        } else {
```
dajimenezriv, line 197: If no pathNames then pathNames is userSystemPath

PR #821, opened 2025-06-11
Add file size too big issue

dajimenezriv-internxt:
## What

When a file size is too big add an issue. This works for the sync and backups tab.


src/apps/backups/remote-tree/traverser.ts
```diff
@@ -59,10 +62,14 @@ export class Traverser {
 
         tree.files[relativePath] = file;
       } catch (exc) {
-        /**
-         * v2.5.3 Daniel Jiménez
-         * TODO: Add issue to backups
-         */
+        if (relativePath) {
+          const name = relativePath;
```
dajimenezriv, line 66: We need to create this intermidiate variable, otherwise it throws a type error because it thinks that relativePath can still be undefined.

PR #824, opened 2025-06-12
Add get files from directory test

dajimenezriv-internxt:
## What

Add test to getFilesFromDirectory and move it to a non recursive function.


src/apps/shared/fs/PathTypeChecker.ts
```diff
@@ -0,0 +1,18 @@
+import { logger } from '../logger/logger';
```
dajimenezriv, line 1: Is the same as above, however now we use our version of fileSystem.stat and also if you check the previous name it was PathTypeChecker .ts, with an space in between.

src/apps/main/antivirus/utils/get-files-from-directory.ts
```diff
@@ -1,47 +1,48 @@
...
-      } else if (item.isDirectory()) {
-        await _getFilesFromDirectory({
-          filePaths,
-          folder: fullPath,
-        });
+      for (const item of items) {
+        const name = item.name.toLowerCase();
+
+        if (name.includes('temp')) continue;
+        if (name.includes('tmp')) continue;
```
AlexisMora: this conditional could be in the same line right?
```suggestion
        if (name.includes('temp') || name.includes('tmp')) continue;
```

PR #828, opened 2025-06-16
Improve logs and add tanstack query

dajimenezriv-internxt:
## What

1. Improve log system to be shorter and easier to read. Basically join the level, process and tag into the same key and don't print SERVER and NETWORK exceptions since they are known errors.
2. Add tanstack query to start using some cache in the renderer and stop making the same call to the backend many times (getUsage was being called like 8 times).


src/apps/renderer/api/get-usage.ts
```diff
@@ -0,0 +1,9 @@
+import { useQuery } from '@tanstack/react-query';
+import { queryKeys } from '../core/tanstack-query/query-keys';
+
+export const useApiUsage = () => {
```
AlexisMora: you could change the file to reflect the proper name of the hook, and move it inside the hooks folder

src/apps/renderer/hooks/useUsage.tsx
```diff
@@ -1,35 +1,14 @@
...
+  const { data: usage, isLoading, isError, refetch } = useApiUsage();
 
-  const updateUsage = useCallback(async () => {
-    try {
-      const userIsLoggedIn = await window.electron.isUserLoggedIn();
+  const status = useMemo(() => {
+    if (isLoading) return 'loading';
+    if (isError) return 'error';
+    return 'ready';
+  }, [usage, isLoading, isError]);
```
AlexisMora, line 11: since `useApiUsage` is a catched query, usage is not gonna change right?

src/apps/renderer/api/get-usage.ts
dajimenezriv: Do you mean calling it use-get-usage and the hook itself useGetUsage? About moving it inside hooks folders I think it doesn't make sense because not all hooks are related.

src/apps/renderer/hooks/useUsage.tsx
dajimenezriv: Well, it can, maybe we have later a policy of refetching every 5 seconds globally or just for this endpoint. Or when it's refetched with the function refetchUsage.

src/apps/renderer/api/get-usage.ts
dajimenezriv: The path of the file is following the arquitecture: core, pages, features and api/ipc. The name of api is just to differenciate from ipc calls between the main and sync-engine process.

AlexisMora: okay then i find it usefull to name it  use-get-usage

PR #830, opened 2025-06-17
Fix max file size in backups

dajimenezriv-internxt:
## What

Add max size to files added in backups. Max is 20GB.


src/context/local/localTree/application/LocalTreeBuilder.infra.test.ts
```diff
@@ -0,0 +1,46 @@
+import { mkdir, rm, writeFile } from 'fs/promises';
```
dajimenezriv, line 1: This is exactly the same as LocalTreeBuilder.test.ts, however, now it has been moved to infra since it was interacting with the file explorer. A new LocalTreeBuilder.test.ts has been created that mocks CLSFsLocalItemsGenerator.getAll, the method used to retrieve local files and folders.

PR #831, opened 2025-06-17
Add workspace token to delete file by uuid

dajimenezriv-internxt:
## What

Add x-workspace-token to deleteFileByUuid request since now it's executed from main process and it doesn't have access to getConfig().workspaceToken.


src/infra/drive-server-wip/services/storage.service.ts
```diff
@@ -43,24 +44,28 @@ export class StorageService {
     });
   }
 
-  deleteFileByUuid(context: { uuid: string }) {
-    const promise = noContentWrapper({
+  deleteFileByUuid(context: { uuid: string; workspaceToken: string }) {
+    const promise1 = noContentWrapper({
       request: client.POST('/storage/trash/add', {
+        headers: { 'x-internxt-workspace': context.workspaceToken },
```
dajimenezriv, line 50: Now this function is executed only in main process, so we need to add manually the workspaceToken.

src/context/virtual-drive/files/application/FileDeleter.ts
```diff
@@ -48,13 +48,9 @@ export class FileDeleter {
...
       file.trash();
 
-      const promise = () => ipcRendererDriveServerWip.invoke('storageDeleteFileByUuid', { uuid: file.uuid });
-      const { error } = await retryWrapper({
-        promise,
-        loggerBody: {
-          tag: 'SYNC-ENGINE',
-          msg: 'Retry deleting file',
-        },
+      const { error } = await ipcRendererDriveServerWip.invoke('storageDeleteFileByUuid', {
```
dajimenezriv, line 51: This is a fix since I forgot to add the workspaceToken.

dajimenezriv: Also, the function itself adds the retryWrapper so we don't need to include it everytime we call it.

src/apps/backups/Backups.ts
```diff
@@ -181,25 +182,6 @@ export class Backup {
     }
   }
 
-  private async deleteRemoteFiles(context: BackupsContext, deleted: Array<File>) {
```
dajimenezriv, line 184: This has been extracted into delete-remote-files and added a test. However we no longer throw an error if we cannot delete a file because it will stop the backup.

AlexisMora (approved):

PR #836, opened 2025-06-19
Extract scan file in antivirus

dajimenezriv-internxt:
## What

Extract scanFile function in antivirus and improve logs.


src/apps/main/antivirus/scan-file.ts
```diff
@@ -0,0 +1,51 @@
+import { logger } from '@/apps/shared/logger/logger';
```
dajimenezriv, line 1: It's the same code as before. We just extract it and change the throw in the end for a log.

dajimenezriv: Because if we throw it stops the backgroundQueue.drain and the antivirus never finish

PR #838, opened 2025-06-19
Fix/br 1235 unable to delete files

AlexisMora:


src/backend/features/remote-notifications/in/process-web-socket-event.ts
```diff
@@ -0,0 +1,18 @@
+import { broadcastToWindows } from '@/apps/main/windows';
+import { FOLDER_DELETED, NOTIFICATION_SCHEMA } from '@/apps/main/notification-schema';
+import { logger } from '@/apps/shared/logger/logger';
+import { debouncedSynchronization } from '@/apps/main/remote-sync/handlers';
+import { handleParsedNotificationEvent } from './handle-parsed-notification-event';
+
+export const processWebSocketEvent = async (data: unknown) => {
```
dajimenezriv: Use object prop

src/backend/features/remote-notifications/in/handle-parsed-notification-event.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const handleParsedNotificationEvent = async ({ event }: { event: NotificationSchema }) => {
+  if (event.clientId === 'drive-desktop') {
+    logger.debug({
+      msg: 'Notification received',
+      event: event.event,
+      clientId: event.clientId,
+      payload: event.payload,
+    });
+    return { data: true };
```
dajimenezriv: You are not in infra, so you shouldn't need to return {data: true or false}. Now, with your code also, if I'm not wrong you are printing two times the notification in case the cliend is drive-desktop. You print it here, you return true and then you print it again in processWebSocketEvent because you are comparing against false.

src/backend/features/remote-notifications/in/mark-items-as-trashed.test.ts
```diff
@@ -0,0 +1,52 @@
...
+describe('markItemsAsTrashed', () => {
+  const splitItemsIntoFilesAndFoldersMock = deepMocked(splitItemsIntoFilesAndFolders);
+  const updateDatabaseStatusToTrashedMock = deepMocked(updateDatabaseStatusToTrashed);
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('should update files and folders to TRASHED status', async () => {
+    const items = [
```
dajimenezriv: Here you can reuse const items.

src/backend/features/remote-notifications/in/mark-items-as-trashed.ts
```diff
@@ -0,0 +1,19 @@
+import { ItemsToTrashEvent } from '@/apps/main/notification-schema';
+import { logger } from '@/apps/shared/logger/logger';
+import { splitItemsIntoFilesAndFolders } from '@/backend/features/remote-notifications/in/split-items-into-files-and-folders';
+import { updateDatabaseStatusToTrashed } from '@/backend/features/remote-notifications/in/update-database-status-to-trashed';
+
+export const markItemsAsTrashed = async ({ items }: { items: ItemsToTrashEvent['payload'] }) => {
+  try {
+    const { files, folders } = splitItemsIntoFilesAndFolders({ items });
+    await updateDatabaseStatusToTrashed(files, folders);
+    return { data: true };
```
dajimenezriv: Same here, you are not in infra.

src/backend/features/remote-notifications/in/process-web-socket-event.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const processWebSocketEvent = async ({ data }: { data: unknown }) => {
+  if (FOLDER_DELETED.safeParse(data).success) {
+    broadcastToWindows('refresh-backup', undefined);
+  }
+  const parsedEventData = await NOTIFICATION_SCHEMA.safeParseAsync(data);
+  if (parsedEventData.success) {
+    const { data } = await handleParsedNotificationEvent({ event: parsedEventData.data });
+    if (data === false) return;
+  }
```
dajimenezriv, line 14: What I would do here is just wrap this:
```ts
 logger.info({ msg: 'Notification received', data });
  await debouncedSynchronization();
```
into a function and I call it here in case parsedEventData.success is false and inside handleParsedNotificationEvent in case client is not drive-desktop. With that you don't need to be handling all returns.

src/backend/features/remote-notifications/in/update-database-status-to-trashed.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  it('should update file and folder statuses to TRASHED', async () => {
+    await updateDatabaseStatusToTrashed([{ type: 'file', uuid: 'f-1' }], [{ type: 'folder', uuid: 'd-1' }]);
+
+    expect(driveFilesCollectionUpdateInBatchMock).toHaveBeenCalledWith(
+      expect.objectContaining({
+        payload: { status: 'TRASHED' },
+        where: {
+          uuid: expect.any(Object),
```
dajimenezriv: Can it be expect.any(String)?

AlexisMora: No, because inside the updateInBach has this:
```ts
where: { uuid: In(files.map((file) => file.uuid)) },
```

src/backend/features/remote-notifications/in/process-web-socket-event.ts
```diff
@@ -0,0 +1,24 @@
...
+
+async function logAndSync({ data }: { data: unknown }) {
+  logger.info({ msg: 'Notification received', data });
+  await debouncedSynchronization();
+}
+
+export async function processWebSocketEvent({ data }: { data: unknown }) {
+  if (FOLDER_DELETED.safeParse(data).success) {
+    broadcastToWindows('refresh-backup', undefined);
+    await logAndSync({ data });
```
dajimenezriv: In this case you are going to log the notifcation two times. One here, and the other at the end of the function

src/backend/features/remote-notifications/in/mark-items-as-trashed.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    { type: 'file' as const, uuid: 'file-uuid' },
+    { type: 'folder' as const, uuid: 'folder-uuid' },
+  ];
+
+  it('should update files and folders to TRASHED status', async () => {
+    splitItemsIntoFilesAndFoldersMock.mockReturnValue({
+      files: [{ type: 'file', uuid: 'file-uuid' }],
+      folders: [{ type: 'folder', uuid: 'folder-uuid' }],
+    });
+    updateDatabaseStatusToTrashedMock.mockResolvedValueOnce(undefined);
```
dajimenezriv: Since you are not using the response, you don't need mock the resolved value

src/backend/features/remote-notifications/in/handle-parsed-notification-event.ts
```diff
@@ -0,0 +1,16 @@
...
+export const handleParsedNotificationEvent = async ({ event }: { event: NotificationSchema }) => {
+  if (event.clientId === 'drive-desktop') {
+    logger.debug({
+      msg: 'Notification received',
+      event: event.event,
+      clientId: event.clientId,
+      payload: event.payload,
+    });
+  } else if (event.event === 'ITEMS_TO_TRASH') {
+    await markItemsAsTrashed({ items: event.payload });
```
dajimenezriv, line 20: In this case, I think you are never going to log the notification.

dajimenezriv-internxt (approved):

PR #839, opened 2025-06-19
Improve logs

dajimenezriv-internxt:
## What

1. Improve architecture in CONTRIBUTING.
2. Set a different queue-manager.json for each workspace (this needs improvement).
3. Improve logs.


tests/context/virtual-drive/contents/infrastructure/upload/EnvironmentContentFileUploader.test.ts
```diff
@@ -14,13 +14,18 @@ describe('Environment Content File Uploader', () => {
...
 
-      const uploader = new EnvironmentContentFileUploader(strategy, bucket, abortSignal);
+      const uploader = new EnvironmentContentFileUploader(strategy, bucket);
 
       const handler = vi.fn();
 
       uploader.on('start', handler);
 
-      await uploader.upload(Readable.from(''), validFileSize);
+      await uploader.upload({
```
dajimenezriv, line 23: This just converts the props to object and I send the path so it can be logged inside the function and move the abort signal from the constructor to the call.

src/apps/main/virtual-root-folder/service.ts
```diff
@@ -51,14 +49,6 @@ export function getRootVirtualDrive(): string {
   return current;
 }
 
-export async function clearRootVirtualDrive(): Promise<void> {
-  try {
-    await fsPromises.rm(PATHS.QUEUE_MANAGER, { recursive: true, force: true });
```
dajimenezriv, line 56: The queue manager theoricaly stores the tasks of adding, deleting, hydrating and so in a json file (this is not happening very well). Now, each queue manager is going to have it's own workspace queue manager so we do not need to remove it since if we log with another account we are going to use another queue manager file.

src/apps/main/auth/handlers.ts
```diff
@@ -74,7 +72,6 @@ export function setupAuthIpcHandlers() {
     if (!canHisConfigBeRestored(data.user.uuid)) {
       setupRootFolder(data.user);
     }
-    await clearRootVirtualDrive();
```
AlexisMora, line 77: Why this is not necessary anymore?

src/context/virtual-drive/contents/domain/contentHandlers/ContentFileUploader.ts
```diff
@@ -1,6 +1,3 @@
-import { Readable } from 'stream';
-import { ContentsId } from '../ContentsId';
-
 type ConenentsId = string;
```
AlexisMora: since you deleted the interface that gave name to this file, we could either rename this file to better reflect whats inside or, move the content to another file where it has more sense

src/apps/main/auth/handlers.ts
dajimenezriv: For the reason I've put in the below comment (the one that defined the function clearRootVirtualDrive).

src/context/virtual-drive/contents/domain/contentHandlers/ContentFileUploader.ts
dajimenezriv: What name would you choose? I think that EnvironmentContentFileUploader is not bad, what it does is upload the file content, maybe just remove the Environment word

PR #842, opened 2025-06-21
Create node win module

dajimenezriv-internxt:
## What

Creates the module NodeWin that will help to interact with the C++. Currently we already have the VirtualDrive but it's not tested and not extracted into the different files. The point is to migrate from the god class VirtualDrive into the NodeWin infra module.

Currently there is an error in the pipeline because NodeWin is deadcode. However, this will be used in the next PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "package": "ts-node .erb/scripts/clean.js && npm run build && electron-builder build --publish never",
     "init:dev": "pnpm install && node node_modules/electron/install.js && npm run build:dll && npm run reload-native-deps",
     "========== Code style ==========": "",
-    "lint": "eslint . --ext .ts,.tsx --max-warnings 554",
+    "lint": "eslint . --ext .ts,.tsx --max-warnings 584",
     "lint:fix": "npm run lint -- --fix",
     "format": "prettier src examples --check",
     "format:fix": "prettier src examples --write",


src/node-win/addon/addon-zod.ts
```diff
@@ -9,7 +9,7 @@ export const addonZod = {
   createEntry: z.object({ success: z.boolean(), errorMessage: z.string().optional() }),
   createPlaceholderFile: z.object({ success: z.boolean(), errorMessage: z.string().optional() }),
   dehydrateFile: z.boolean(),
-  getFileIdentity: z.string(),
+  getFileIdentity: z.union([z.literal(''), z.string().startsWith('FILE:'), z.string().startsWith('FOLDER:')]),
```
dajimenezriv, line 12: This function from node-win should only return '', FILE:* or FOLDER:*. We apply zod to log an error in case the return value doesn't match but we don't stop code execution.

examples/handlers/handle-add.ts
```diff
@@ -7,7 +7,7 @@ import { QueueItem } from '@/node-win/queue/queueManager';
 export const handleAdd = (task: QueueItem) => {
```
dajimenezriv, line 7: Practically everything that is inside the examples folder is going to be deleted because nobody uses it, currently is broken and deleting it will help into no importing from a wrong place or not having to use callbacks when initializing VirtualDrive. This code does not run on the application, so you don't need to check it.

AlexisMora (approved):

PR #843, opened 2025-06-21
Fix add folder to drive

dajimenezriv-internxt:
## What

Improve and simplify the way we persist a folder in the server and without relying on the in memory repository. Basically, instead of looking for the parent folder for obtaining the uuid we search for it in the file explorer using the node-win.

### DIFF
--- a/src/context/virtual-drive/folders/infrastructure/HttpRemoteFolderSystem.ts
+++ b/src/context/virtual-drive/folders/infrastructure/HttpRemoteFolderSystem.ts
@@ -36,7 +36,7 @@ export class HttpRemoteFolderSystem {
         createdAt: data.createdAt,
         status: FolderStatuses.EXISTS,
       };
-    } catch (error: unknown) {
+    } catch (exc) {
       const existing = await this.existFolder(offline);
 
       if (existing.status !== FolderStatuses.EXISTS) {
@@ -45,7 +45,7 @@ export class HttpRemoteFolderSystem {
           msg: 'Error creating folder',
           basename: offline.basename,
           parentUuid: offline.parentUuid,
-          exc: error,
+          exc,
         });
       }
 


src/features/sync/add-item/create-folder.ts
```diff
@@ -0,0 +1,44 @@
+import { logger } from '@/apps/shared/logger/logger';
```
dajimenezriv, line 1: When we want to create a folder placeholder it can happen that folderCreator.run throws an exception because the parent folder does not exist. In that case we just run the function to create the parent folder and then run the create folder again.

src/context/virtual-drive/folders/infrastructure/HttpRemoteFolderSystem.ts
```diff
@@ -36,14 +36,20 @@ export class HttpRemoteFolderSystem {
         createdAt: data.createdAt,
```
dajimenezriv, line 36: When we try to persist a folder maybe that folder already exists (this can happen because when we try to create an inner file, if the parent folder does not exist it tries to create it again, so we can have multiple calls to create the same parent). We will investigate a way to docouple the same request. In this case, if it throws an exception because the folder exists we check if the folder exists using the function this.existFolder and return it if the status is EXISTS.

dajimenezriv: If the folder doesn't exist it means that we had another exception, so we just throw it.

src/context/virtual-drive/folders/application/FolderCreator.ts
```diff
@@ -1,9 +1,17 @@
 import { Service } from 'diod';
```
dajimenezriv, line 1: We stop using the InMemory. Now, to create a folder we try to get the placeholderId of the parent folder using the node-win. If it exists we extract the uuid from the parentId and we have everything we need, the uuid of the parent folder.

src/context/virtual-drive/files/application/FileSyncronizer.ts
```diff
@@ -87,44 +91,6 @@ export class FileSyncronizer {
     }
   };
 
-  private runFolderCreator(posixRelativePath: string): Promise<Folder> {
```
dajimenezriv, line 90: Everything deleted here is replaced by the create-folder file.

src/apps/sync-engine/callbacks/handleAdd.service.ts
```diff
@@ -14,44 +14,28 @@ export class HandleAddService {
...
-        itemPath: task.path,
-        // @ts-expect-error TODO: solve in next PR
-        id: itemId,
-      });
-      drive.updateSyncStatus({
-        itemPath: task.path,
-        isDirectory: task.isFolder,
-        sync: true,
-      });
+      await self.controllers.addFile.execute({ absolutePath: task.path, drive });
```
dajimenezriv, line 33: We don't need to return the placeholderId. Inside the function if it's a file or a folder it creates the placeholder. We removed this because previously the folder created the placeholder and since we returning the placeholderId, it was trying to create a file placeholder again (so for each folder it was trying to create a folder placeholder and a file placeholder).

src/apps/sync-engine/callbacks-controllers/controllers/AddController.ts
```diff
@@ -64,97 +65,7 @@ export class AddController extends CallbackController {
     }
   };
 
-  private createFolder = async (offlineFolder: OfflineFolder, attempts = 3): Promise<string> => {
```
dajimenezriv, line 67: This is done in the file create-folder.ts

PR #846, opened 2025-06-23
chore: delete examples scripts

cg27shokworks:


dajimenezriv:
Two comments.
First one, you still need to delete /examples/tmp from the .gitignore.
Second, you still have two errors in the pipeline deadcode:
image
The second one (node-win.module.ts) is intentional and it will be solved in my next PR. However, you can see that you are exporting TBody in logger.ts but it's not used outside that file so you should remove the export. The same for TFetchDataCallback.

cg27shokworks:
thanks @dajimenezriv-internxt , the changes you requested are resolved. In both cases these changes came from main.

PR #847, opened 2025-06-24
fix: fs provider busy file error handler

cg27shokworks:
## What
The user's file are busy, so it is not possible delete those files
https://inxt.atlassian.net/browse/BR-1231?atlOrigin=eyJpIjoiOGNkNzgyNTlhY2U1NDY4NGEyMTRkNmQxOGQ1MzRlZjkiLCJwIjoiaiJ9

This is getting this code when trying to delete a folder 0x800701AA basically this happens because “The required resource is in use”.

Inside the logs there are a bunch of these errors
[error] Error providing file: Error: UNKNOWN: unknown error, read

this occurs in FSLocalFileProvider, this occurs because the file is busy and it seems that the error code type is not being properly addressed.

Basically I created a pr focused on detecting what may be keeping the users files busy, it now recognizes EBUSY, EPERM, EACCES, ENOENT and messages containing “busy” or “access denied”,

This is not the definitive solution but I think that in the logs there are not enough elements to detect it, so this will help us to detect similar errors in other users

## Update

Previously the process for uploading a sync file (this was not the same in backups) was using a retry of 2 times of uploading a file. Also, inside this uploading we had two parts, the first one was checking if the file was locked by something else, which was the thing that Carlos changed and then it also was using the `watch` method from `fs` to watch the whole directory and check for changes in the file or the modification time, in case of that the contents upload was aborted. Since we had problems about large files not being uploaded and since this code wasn't tested and it was pretty complex, I removed it and I'm going to add a `setTimeout` that every 5 seconds checks if the size is still the same (this was recommended by chatgpt because `watch` the folder is to complex and the method `watchFile` from `fs` is not reliable). I've already created the task in the slack canvas and I will do it in the next PR. For now, the retry is removed and these checks so the upload is now the same as backups, that didn't have any retry, check or anything. The timeout for changed size will be aplied to backups and sync.

### DIFF
--- a/src/backend/features/local-sync/watcher/events/unlink/get-parent-uuid.test.ts
+++ b/src/backend/features/local-sync/watcher/events/unlink/get-parent-uuid.test.ts
@@ -13,7 +13,6 @@ describe('get-parent-uuid', () => {
   const props = mockProps<typeof getParentUuid>({ path });
 
   beforeEach(() => {
-    vi.clearAllMocks();
     getFolderUuidMock.mockReturnValue({ data: 'parentUuid' as FolderUuid });
     statMock.mockResolvedValue({ data: { size: 1024 } });
   });
--- a/src/context/local/localFile/application/update/file-batch-updater.test.ts
+++ b/src/context/local/localFile/application/update/file-batch-updater.test.ts
@@ -16,8 +16,6 @@ describe('file-batch-updater', () => {
   let props: Parameters<typeof service.process>[0];
 
   beforeEach(() => {
-    vi.clearAllMocks();
-
     props = mockProps<typeof service.process>({
       self: { backed: 0 },
       tracker: { currentProcessed: vi.fn() },
--- a/src/context/local/localFile/application/upload/file-batch-uploader.test.ts
+++ b/src/context/local/localFile/application/upload/file-batch-uploader.test.ts
@@ -22,8 +22,6 @@ describe('file-batch-uploader', () => {
   const parentPath = pathUtils.dirname(path);
 
   beforeEach(() => {
-    vi.clearAllMocks();
-
     props = mockProps<typeof service.run>({
       self: { backed: 0 },
       tracker: { currentProcessed: vi.fn() },
--- a/src/node-win/watcher/detect-context-menu-action.test.ts
+++ b/src/node-win/watcher/detect-context-menu-action.test.ts
@@ -18,8 +18,6 @@ describe('detect-context-menu-action', () => {
   let props: Parameters<typeof service.execute>[0];
 
   beforeEach(() => {
-    vi.clearAllMocks();
-
     props = mockProps<typeof service.execute>({
       absolutePath: 'absolutePath' as AbsolutePath,
       path: createRelativePath('file.txt'),


src/context/virtual-drive/contents/infrastructure/FSLocalFileProvider.ts
```diff
@@ -76,6 +89,15 @@ export class FSLocalFileProvider {
     const readStream = createReadStream(filePath);
     const controller = new AbortController();
 
+    readStream.on('error', (err: NodeJS.ErrnoException) => {
```
dajimenezriv: Are we sure is NodeJs.ErrnoException? Because if you enter in ReadStream it says is just error. Also, if we want to prevent the UNKNOWN, then we shouldn't use UNKNOWN. You can use the logger from logger.ts and add the whole error object.

src/context/virtual-drive/contents/infrastructure/FSLocalFileProvider.ts
```diff
@@ -34,22 +34,35 @@ export class FSLocalFileProvider {
...
+                path: err.path,
+                retriesLeft,
+              });
               reject(err);
             }
           });
         });
       } catch (error) {
-        Logger.error(`Error reading file: ${error}`);
+        Logger.error(`Error reading file: ${error}`, { filePath });
```
dajimenezriv: You can do here throw logger.error and add the error object as param.

dajimenezriv: And the same above, use our logger.

src/apps/backups/Backups.ts
```diff
@@ -108,79 +98,16 @@ export class Backup {
...
     await Promise.all([
-      this.uploadAndCreateFile(context, tracker, local.root.absolutePath, added, remote),
-      this.uploadAndUpdate(context, tracker, modified, remote),
+      this.fileBatchUploader.run({ self: this, tracker, context, remoteTree: remote, added }),
+      this.fileBatchUpdater.run({ self: this, tracker, context, modified }),
       deleteRemoteFiles({ context, deleted }),
     ]);
   }
 
-  private async uploadAndCreateFile(
```
dajimenezriv, line 127: These two functions were responsible of calling grouping the files to be uploaded or updated by batches, then call for each batch the FileBatchUpdater or FileBatchUploader, update the progress and do a try catch. However, know they are no longer necessary. Why? We are no longer grouping by batches and we send the whole list of files to be uploaded or updated because the bottleneck already handles the concurrency. Since we call that directly, the abort controller here doesn't do anything, then the progress is done inside each function and we perform a try catch inside the FileBatchUploader and FileBatchUpdater so we can't have an exception at this level. I recommend to look at FileBatchUploader and FileBatchUpdater at the same time to understand this code.

src/apps/sync-engine/callbacks-controllers/controllers/update-contents-id.test.ts
```diff
@@ -48,24 +47,14 @@ describe('update-contents-id', () => {
...
-    statSpy.mockResolvedValue({ data: { size: BucketEntry.MAX_SIZE + 1 } });
+    props.stats.size = BucketEntry.MAX_SIZE + 1;
     // When
     await updateContentsId(props);
     // Then
     expect(fileContentsUploader.run).toBeCalledTimes(0);
     expect(loggerMock.error).toBeCalledTimes(0);
   });
 
-  it('should not update contents id if stat returns error', async () => {
```
dajimenezriv, line 54: We send stats as prop so we don't have to calculate it.

src/context/local/localFile/application/update/FileBatchUpdater.ts
```diff
@@ -1,29 +1,72 @@
...
+    try {
+      const contentsId = await uploadFile({ context, localFile, uploader: this.uploader });
 
-      /**
-       * v2.5.3 Daniel Jiménez
-       * TODO: Check file can be null or contentsId maybe continue???
-       */
+      if (!contentsId) return;
 
-      await simpleFileOverride(file, contentsId, localFile.size.value);
```
dajimenezriv, line 26: This simpleFileOverride was just doing the driveServerWip call

src/context/virtual-drive/contents/application/ContentsUploader.ts
```diff
@@ -1,45 +1,47 @@
...
 
-  async run(path: string) {
-    try {
-      const win32RelativePath = PlatformPathConverter.posixToWin(path);
-
-      const absolutePath = this.relativePathToAbsoluteConverter.run(win32RelativePath) as AbsolutePath;
+  async run({ path, stats }: Props) {
+    const win32RelativePath = PlatformPathConverter.posixToWin(path);
 
-      const { readable, abortSignal, size } = await this.contentProvider.provide(absolutePath);
```
dajimenezriv, line 22: This was the function that was checking if the file was modified or changed size during uploading. In case of that it aborted the upload. However, we are going to replace this with a setTimeout checking the size, for now, we don't abort an upload.

src/infra/inxt-js/file-uploader/environment-file-uploader.ts
```diff
@@ -71,4 +74,8 @@ export class EnvironmentFileUploader {
       });
     });
   }
+
+  async run(props: TProps) {
```
dajimenezriv, line 78: Previously the limit was in RetryContentsUploader, now it's moved to the ContentsUploader class.

dajimenezriv-internxt (dismissed):

PR #859, opened 2025-06-30
Create file or folder in background sync

dajimenezriv-internxt:
## What

Now, since we have move the checkpoint outside the sqlite database we can create files and folders in the background sync.


src/apps/main/background-processes/sync-engine/services/spawn-workspace.ts
```diff
@@ -32,6 +32,7 @@ export async function spawnWorkspace({ workspace }: TProps) {
   });
 
   const config: Config = {
+    userUuid: user.uuid,
```
dajimenezriv, line 35: This is just for not doing getUser or getUserOrThrow everytime

AlexisMora (approved):

PR #865, opened 2025-07-01
Remove add from queue manager

dajimenezriv-internxt:


src/apps/main/background-processes/sync-engine/services/spawn-sync-engine-worker.ts
```diff
@@ -47,10 +48,10 @@ export async function spawnSyncEngineWorker({ config }: TProps) {
    * Since we can have a different status in our local database that in remote,
    * we want to run also this sync in background to update the statuses.
    */
-  void RemoteSyncModule.syncItemsByFolder({
-    rootFolderUuid: config.rootUuid,
-    context: config,
-  });
+  // void RemoteSyncModule.syncItemsByFolder({
```
dajimenezriv, line 51: This is executed before the first updateCheckpoint so we can lose it, comment for now and investigate this in more detail

src/apps/sync-engine/callbacks-controllers/controllers/AddController.ts
```diff
@@ -1,106 +1,55 @@
...
 
   constructor(
-    private readonly absolutePathToRelativeConverter: AbsolutePathToRelativeConverter,
     private readonly fileCreationOrchestrator: FileCreationOrchestrator,
     private readonly folderCreator: FolderCreator,
-  ) {
-    super();
-  }
-
-  private createFile = async (posixRelativePath: string, attempts = 3): Promise<FilePlaceholderId> => {
```
dajimenezriv, line 28: Remove attemps. As you can see it was increasing the attemps when FolderNotFoundError, that doesn't make sense because is not an error. We already have retries in drive-server-wip and content uploader, so we don't need retries here.

src/context/local/localFile/infrastructure/AbsolutePath.ts
```diff
@@ -8,14 +8,21 @@ export function createRelativePath(...parts: string[]): RelativePath {
   let path = posix.join(posix.sep, ...parts);
   path = path.replaceAll(win32.sep, posix.sep);
   path = posix.normalize(path);
+  path = path !== '/' ? path.replace(/\/+$/, '') : path;
```
dajimenezriv: If it ends with "/" but it's not "/", remove it: /a/b/ -> /a/b

src/context/virtual-drive/boundaryBridge/application/FileCreationOrchestrator.ts
```diff
@@ -1,29 +1,36 @@
...
     private readonly fileCreator: FileCreator,
   ) {}
 
-  async run(posixRelativePath: string): Promise<File['uuid']> {
-    const path = new FilePath(posixRelativePath);
+  async run({ path }: TProps) {
+    const filePath = new FilePath(path);
 
-    const fileContents = await this.contentsUploader.run(posixRelativePath);
+    const fileContents = await limiter.schedule(() => this.contentsUploader.run(path));
```
dajimenezriv, line 23: Add a bottleneck of upload max files at the same time, which was the previous concurrency using the queue manager.

src/node-win/addon-wrapper.ts
```diff
@@ -189,27 +189,7 @@ export class Addon {
 
   convertToPlaceholder({ path, id }: { path: string; id: FilePlaceholderId | FolderPlaceholderId }) {
     const result = addon.convertToPlaceholder(path, id);
-
-    this.parseAddonZod('convertToPlaceholder', result);
-
-    if (result.success) {
```
dajimenezriv, line 195: Move this log to the virtual drive instead of the addron wrapper.

AlexisMora (approved):

AlexisMora (approved):

PR #871, opened 2025-07-05
Remove queue manager

dajimenezriv-internxt:
## What

1. Remove queue manager and call directly the hydrate and dehydrate methods. We add a bottleneck of maxConcurrent 1 to hydrate and we allow dehydrate to be executed in parallel.
2. Also, create the FileDTO por events FILE_CREATED and FILE_UPDATED.


src/apps/sync-engine/BindingManager.ts
```diff
@@ -26,17 +23,13 @@ import { Watcher } from '@/node-win/watcher/watcher';
 export type CallbackDownload = (data: boolean, path: string, errorHandler?: () => void) => Promise<{ finished: boolean; progress: number }>;
 
 export class BindingsManager {
-  progressBuffer = 0;
   controllers: IControllers;
 
-  lastHydrated = '';
```
dajimenezriv, line 32: This was a way that was used before to solve the bug the a file received two times the hydrated event when we were were hydrating using double click in a file. However, this is already solved in detect-context-menu-action.service.ts.

src/apps/sync-engine/callbacks-controllers/controllers/update-contents-id.ts
```diff
@@ -25,6 +27,8 @@ export async function updateContentsId({ path, uuid, fileContentsUploader, repos
       newSize: contents.size,
     });
 
+    virtualDrive.updateSyncStatus({ itemPath: path, isDirectory: false, sync: true });
```
dajimenezriv, line 30: We need to force the update of the placeholder icon when we change the contents of a file.

src/apps/sync-engine/callbacks/fetchData.service.ts
```diff
@@ -15,76 +14,91 @@ type TProps = {
...
     try {
-      Logger.debug('[Fetch Data Callback] Donwloading begins');
-
       const startTime = Date.now();
       const path = await self.controllers.downloadFile.execute(filePlaceholderId, callback);
+      const nameWithExtension = basename(path);
 
-      const trimmedPlaceholderId = trimPlaceholderId({ placeholderId: filePlaceholderId });
-      const parsedPlaceholderId = trimmedPlaceholderId.split(':')[1];
-      const file = self.controllers.downloadFile.fileFinderByUuid({ uuid: parsedPlaceholderId });
```
dajimenezriv, line 25: We were using the file to obtain the file name, however we can use the path from before and obtain it using basename.

src/apps/sync-engine/callbacks/handleHydrate.service.ts
```diff
@@ -1,51 +1,30 @@
-import Logger from 'electron-log';
 import VirtualDrive from '@/node-win/virtual-drive';
-import { BindingsManager } from '../BindingManager';
-import configStore from '../../../apps/main/config';
-import { QueueItem } from '@/node-win/queue/queueManager';
+import { logger } from '@/apps/shared/logger/logger';
+import { RelativePath } from '@/context/local/localFile/infrastructure/AbsolutePath';
+import Bottleneck from 'bottleneck';
+
+const limiter = new Bottleneck({ maxConcurrent: 1 });
```
dajimenezriv, line 6: In the queue manager, the tasks were run one by one, so we add here the bottleneck.

src/node-win/watcher/detect-context-menu-action.service.ts
```diff
@@ -19,10 +20,7 @@ export class DetectContextMenuActionService {
     const { prev, curr } = details;
 
     const status = self.virtualDrive.getPlaceholderState({ path });
-    const { data: uuid } = NodeWin.getFileUuid({ drive: self.virtualDrive, path });
```
dajimenezriv, line 22: Move this below, since we don't need the uuid for hydrating or dehydrating.

src/node-win/watcher/detect-context-menu-action.service.ts
```diff
@@ -19,10 +20,7 @@ export class DetectContextMenuActionService {
...
-    const isInDevice = self.fileInDevice.has(absolutePath);
-
-    if (!uuid) return;
+    const isInDevice = self.fileInDevice.has(path);
```
dajimenezriv, line 23: fileInDevice stores all paths that are hydrated, previously we were saving paths that are AbsolutePath, however, it's easy to track RelativePaths.

PR #874, opened 2025-07-07
Extract handle dehydrate

dajimenezriv-internxt:


src/apps/main/database/entities/DriveFile.ts
```diff
@@ -1,5 +1,8 @@
+import { Brand } from '@/context/shared/domain/Brand';
 import { Column, Entity, PrimaryColumn } from 'typeorm';
 
+export type ContentsId = Brand<string, 'ContentsId'>;
```
dajimenezriv, line 4: Create branded type for usual variables so we can differenciate strings, like FileUuid, FolderUuid, FilePlaceholderId, FolderPlaceholderId. This is the same as FileId in `drive-server-wip`.

src/apps/sync-engine/callbacks-controllers/controllers/update-contents-id.ts
```diff
@@ -25,6 +24,10 @@ export async function updateContentsId({ path, uuid, fileContentsUploader, repos
       newSize: contents.size,
     });
 
+    virtualDrive.updateSyncStatus({ itemPath: path, isDirectory: false, sync: true });
```
dajimenezriv, line 43: We need to force the sync of the icon in the placeholder, otherwise it will be broke.

tests/vitest/utils.helper.test.ts
```diff
@@ -14,3 +14,20 @@ export function mockProps<T extends (...args: any[]) => unknown>(props: DeepPart
...
+
+/**
+ * v2.5.6 Daniel Jiménez
+ * Code extracted from vitest
+ * https://github.com/vitest-dev/vitest/blob/c1f78d2adc78ef08ef8b61b0dd6a925fb08f20b6/packages/spy/src/index.ts#L464
+ */
+type Procedure = (...args: any[]) => any;
+type Methods<T> = keyof { [K in keyof T as T[K] extends Procedure ? K : never]: T[K] };
+type Classes<T> = { [K in keyof T]: T[K] extends new (...args: any[]) => any ? K : never }[keyof T] & (string | symbol);
+export function partialSpyOn<T, M extends Classes<Required<T>> | Methods<Required<T>>>(obj: T, methodName: M, mock = true) {
```
dajimenezriv, line 26: This will help applying the spyOn with partial implementation and also by defualt we mock the implementation so we don't have strange behaviour.

src/apps/sync-engine/callbacks/handleDehydrate.service.ts
```diff
@@ -1,20 +1,27 @@
-import Logger from 'electron-log';
 import VirtualDrive from '@/node-win/virtual-drive';
```
AlexisMora: since you added the test file for this with kebab-case you could also rename it

src/node-win/watcher/detect-context-menu-action.service.ts
```diff
@@ -57,8 +58,8 @@ export class DetectContextMenuActionService {
       // }
 
       self.fileInDevice.delete(absolutePath);
-      self.queueManager.enqueue({ path: absolutePath, type: typeQueue.dehydrate, uuid });
-      return 'Liberar espacio';
+      handleDehydrate({ drive: self.virtualDrive, path });
+      return;
```
AlexisMora, line 62: Why this change in return type?

dajimenezriv: The return type is used in the file that calls this function to log the action, however, we are already logging it inside handleDehydrate.

PR #876, opened 2025-07-07
Add file dto to notifications

dajimenezriv-internxt:
## What

Add all properties need in NOTIFICATION_SCHEMA for file, so it matches the type of FileDto. Now we can add it directly to sqlite.


src/apps/main/usage/service.test.ts
```diff
@@ -13,7 +13,7 @@ describe('UserUsageService', () => {
   });
 
   it('should calculate usage correctly', async () => {
-    getUsageMock.mockResolvedValueOnce({ data: { drive: 5000 } });
```
dajimenezriv, line 16: The type of getUsage has changed in drive-server-wip, add the missing properties for test.

AlexisMora (approved):

PR #882, opened 2025-07-09
Improve hydrate queue manager

dajimenezriv-internxt:
## What

Improve the queue manager for the hydrate file. Now we parse the queue with zod, we have simplied it by just using it for hydrate (we don't need to use typeQueue.hydrate everywhere).

### DIFF
--- a/src/apps/sync-engine/BindingManager.ts
+++ b/src/apps/sync-engine/BindingManager.ts
@@ -178,7 +178,7 @@ export class BindingsManager {
     watcher.watchAndWait();
 
     await this.polling({ watcher });
-    await queueManager.processQueue();
+    void queueManager.processQueue();
   }
 
   stop() {


src/apps/sync-engine/BindingManager.ts
```diff
@@ -28,13 +26,11 @@ export class BindingsManager {
   progressBuffer = 0;
   controllers: IControllers;
 
-  lastHydrated = '';
```
dajimenezriv, line 31: This was used because when we were calling hydrate it was calling fetchData internally and it was emiting the event of hydrate two times, so we were setting the value of the path hydrated and then comparing to see that it wasn't doubled call. However, now technically this is stopped in the watcher and I've been doing some tests and it works without this.

src/apps/sync-engine/BindingManager.ts
```diff
@@ -163,12 +159,7 @@ export class BindingsManager {
   }
 
   async watch() {
-    const callbacks = {
```
dajimenezriv, line 166: Now the queueManager calls handleHydrate directly without callbacks

src/apps/sync-engine/callbacks/fetchData.service.ts
```diff
@@ -19,15 +19,12 @@ export class FetchDataService {
...
-      const trimmedPlaceholderId = trimPlaceholderId({ placeholderId: filePlaceholderId });
-      const parsedPlaceholderId = trimmedPlaceholderId.split(':')[1];
-      const file = self.controllers.downloadFile.fileFinderByUuid({ uuid: parsedPlaceholderId });
+      const uuid = NodeWin.getFileUuidFromPlaceholder({ placeholderId: filePlaceholderId });
+      const file = self.controllers.downloadFile.fileFinderByUuid({ uuid });
 
       Logger.debug('[Fetch Data Callback] Preparing begins', path);
       Logger.debug('[Fetch Data Callback] Preparing begins', file.path);
 
-      self.lastHydrated = file.path;
```
dajimenezriv, line 29: This was the part where we were settings the value of lastHydrated to later check in handleHydrate

src/apps/sync-engine/callbacks/handle-hydrate.ts
```diff
@@ -0,0 +1,39 @@
+import VirtualDrive from '@/node-win/virtual-drive';
```
dajimenezriv, line 1: The old file was called handleHydrate and some other things to check the variable lastHydrated, however, it's no longer necessary so we just have to call virtualDrive.hydrateFile, that calls C++ code, then goes to fetch data in javascript and C++

src/context/virtual-drive/contents/application/RetryContentsUploader.ts
```diff
@@ -23,8 +23,6 @@ export class RetryContentsUploader {
           retry,
         });
 
-        retry += 1;
```
dajimenezriv, line 26: This is just a fix because we were updating the retry variable when success, not on the catch, so it didn't make so much sense

src/node-win/watcher/detect-context-menu-action.service.ts
```diff
@@ -40,8 +36,8 @@ export class DetectContextMenuActionService {
         return 'Doble click en el archivo';
       }
 
-      self.queueManager.enqueue({ path: absolutePath, type: typeQueue.hydrate, uuid });
-      return 'Mantener siempre en el dispositivo';
```
dajimenezriv, line 44: This return value was for logging in the function that is calling this function, but we are already loggin that the task has been enqueued

src/node-win/queue/queue-manager.ts
```diff
@@ -1,128 +1,97 @@
...
+      logger.debug({
+        tag: 'SYNC-ENGINE',
+        msg: 'Task already exists in queue. Skipping.',
+        task,
+      });
+      return;
     }
-  }
 
-  async processAll(): Promise<void> {
```
dajimenezriv, line 123: Since now we just have the hydrate key, we no longer need this method, we just call processQueue

src/node-win/queue/queue-manager.ts
```diff
@@ -1,128 +1,97 @@
...
-    this.logger?.debug({ msg: 'enqueue', task });
-    const existingTask = this.queues[task.type].find((item) => item.path === task.path && item.type === task.type);
+    const data = readFileSync(this.queuePath, 'utf-8');
 
-    if (existingTask) {
-      this.logger?.info({ msg: 'Task already exists in queue. Skipping.' });
-      return;
+    try {
+      const parsed = JSON.parse(data);
+      this.queue = QueueSchema.parse(parsed);
```
dajimenezriv, line 43: Now we parse the file against zod to check that the schema is right

src/node-win/queue/queue-manager.ts
```diff
@@ -1,128 +1,97 @@
...
+    private readonly virtualDrive: VirtualDrive,
+    private readonly queuePath: string,
+  ) {
+    if (existsSync(this.queuePath)) {
+      this.loadQueue();
     }
   }
 
-  private saveQueueStateToFile(): void {
-    if (!this.persistPath) return;
```
dajimenezriv, line 41: persistPath cannot be undefined

src/node-win/queue/queue-manager.ts
```diff
@@ -1,128 +1,97 @@
-import fs from 'fs';
+import { existsSync, readFileSync, writeFileSync } from 'fs';
```
dajimenezriv, line 1: Explanation of queue manager. We have the method enqueue that expects an hydrate task, when we call it we call processQueue, this sets processing to true and starts processing all tasks or ignores it if processing was already set to true. processQueue calls processSequentially because hydrate tasks cannot be parallelized yet. We also have the functions loadQueue and saveQueue to file.

PR #883, opened 2025-07-09
Use name with extension on decrypt

dajimenezriv-internxt:


src/context/virtual-drive/files/domain/file-decrypt-name.ts
```diff
@@ -0,0 +1,15 @@
+import * as crypt from '@/context/shared/infrastructure/crypt';
+
+type TProps = {
+  plainName?: string | null;
+  encryptedName: string;
+  parentId: number | null;
+  extension: string | null;
```
AlexisMora, line 7: why not make them all optional?

dajimenezriv: Because you can forgot to send them, also, the values are null not undefined, which is not the same. And in this case you always need to send the encryptedName, parentId and extension to build the name in case the plainName is undefined.

PR #888, opened 2025-07-10
Move rename outside c++

dajimenezriv-internxt:
## What

Move all functionality of moving/renaming a file or a folder from C++ to javascript. Now we listen for add/addDir events using the watcher and base on some conditions we go to move-file or move-folder, where we check if we have to rename or move, it performs the network call, then updates the local sqlite and finally shows the FILE_RENAMED or FILE_RENAME_ERROR in the app. Tested it and works perfectly.


src/apps/sync-engine/BindingManager.ts
```diff
@@ -64,39 +62,6 @@ export class BindingsManager {
       notifyDeleteCompletionCallback: () => {
         Logger.info('Deletion completed');
       },
-      notifyRenameCallback: async (
```
dajimenezriv, line 67: We no longer send the callback to C++

src/apps/sync-engine/BindingManager.ts
```diff
@@ -257,7 +220,8 @@ export class BindingsManager {
       await this.update(tree);
 
       ipcRendererSyncEngine.send('CHANGE_SYNC_STATUS', workspaceId, 'SYNCED');
-    } catch {
+    } catch (exc) {
```
dajimenezriv, line 223: For some reason we were not logging the error, wtf

src/backend/features/local-sync/watcher/events/rename-or-move/get-parent-uuid.ts
```diff
@@ -0,0 +1,30 @@
+import { getConfig } from '@/apps/sync-engine/config';
```
dajimenezriv, line 1: The flow is: watcher checks an add or addDir event, then it goes to move-file or move-folder, which checks the current name and parentUuid in the sqlite (the one before moving/renaming), then we go to move-item and here we call getParentUuid. This function just obtains the newParentUuid of the moved file/folder using the placeholder.

src/backend/features/local-sync/watcher/events/rename-or-move/get-parent-uuid.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export function getParentUuid({ self, path, props, item }: TProps) {
+  const parentPath = pathUtils.dirname(path);
+  const { data: parentUuid, error } = NodeWin.getFolderUuid({
+    drive: self.virtualDrive,
+    rootUuid: getConfig().rootUuid,
+    path: parentPath,
+  });
+
+  if (!item || error) {
```
dajimenezriv, line 24: If the item was not in sqlite or we cannot retrieve the current parentUuid, ignore it. Technically this should never happen because in case that the item is not in sqlite it wouldn't have a placeholder so it would have go to on-add instead of move.

src/backend/features/local-sync/watcher/events/rename-or-move/move-item.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    if (type === 'file') {
+      await ipcRendererDriveServerWip.invoke('moveFileByUuid', { uuid, parentUuid });
+    } else {
+      await ipcRendererDriveServerWip.invoke('moveFolderByUuid', { uuid, parentUuid });
+    }
+  }
+
+  if ((isRenamed || isMoved) && type === 'file') {
+    self.virtualDrive.updateSyncStatus({ itemPath: path, isDirectory: false, sync: true });
```
dajimenezriv, line 64: This is to fix the icon when we move a file. The folder icon is currently borken if we move a folder, but I don't know the solution yet (of course this wasn't solve previously)

src/infra/drive-server-wip/out/ipc-main.ts
```diff
@@ -3,6 +3,7 @@ import { ipcMain } from 'electron';
 import { FromMain, FromProcess } from './ipc';
```
dajimenezriv, line 3: I have to add also the MOVE events but didn't exist for some reason, so I'm going to skip them for now since that functionality wasn't here before and this is just a refactor from C++ to javascript.

AlexisMora (approved):

PR #890, opened 2025-07-10
Feat/use windows defender instead of clamav when possible

egalvis27:


AlexisMora:
This pr is too big, you should split it into smaller ones

PR #892, opened 2025-07-10
feat: Implement Windows Defender antivirus scanning functionality and tests

egalvis27:


src/apps/main/antivirus/windows-defender/test/antivirus-windows-defender.test.ts
```diff
@@ -0,0 +1,88 @@
+import { describe, expect, it, vi } from 'vitest';
```
AlexisMora: The tests should always go alongside the code, not in a separate folder, for example this tests should go in `src/apps/main/antivirus/windows-defender/antivirus-windows.defender.test.ts`

src/apps/main/antivirus/windows-defender/antivirus-windows-defender.ts
```diff
@@ -0,0 +1,41 @@
...
+  isInitialized = false;
+  mpCmdRunPath = '';
+
+  static async createInstance(): Promise<AntivirusWindowsDefender> {
+    const instance = new AntivirusWindowsDefender();
+    await instance.initialize();
+    return instance;
+  }
+
+  initialize(): Promise<void> {
```
AlexisMora: Even though I see a benefit in adding the return type, we decided not to add them so I suggest to remove the return type of the created functions so we can keep a consistency of the code

src/apps/main/antivirus/windows-defender/antivirus-windows-defender.ts
```diff
@@ -0,0 +1,41 @@
...
+    try {
+      this.mpCmdRunPath = findMpCmdRun();
+      this.isInitialized = true;
+      return Promise.resolve();
+    } catch (error) {
+      logger.debug({
```
AlexisMora: logger.debug returns an error so you could return the result of logger.debug

src/apps/main/antivirus/windows-defender/antivirus-windows-defender.ts
```diff
@@ -0,0 +1,41 @@
...
+        tag: 'ANTIVIRUS',
+        msg: 'Error Initializing Windows Defender',
+        error,
+      });
+      return Promise.reject(error);
```
AlexisMora: Unless there is a really specific use case, I am against throwing errors, because makes the code more dificult to follow and prone to errors, since this method seems to be synchronous, I would suggest you to return either a data: boolean or a error:Error, this way when we use the initialize method. we know from the return type that this could return an error and we are required to deal with said possibility

src/apps/main/antivirus/windows-defender/antivirus-windows-defender.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+  }
+
+  async scanFile(filePath: string): Promise<ScanResult> {
+    if (!this.isInitialized) {
+      throw new Error('Windows Defender is not initialized');
```
AlexisMora: same

src/apps/main/antivirus/windows-defender/find-mcp-command.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Finds the full path to the Windows Defender MpCmdRun.exe executable.
+ * Searches the Defender Platform directory for the latest version, falling back to the default path if necessary.
+ * @returns {string} The full path to MpCmdRun.exe.
+ * @throws {Error} If MpCmdRun.exe cannot be found in any of the expected locations.
+ */
+export function findMpCmdRun() {
+  const DEFENDER_PLATFORM_PATH = 'C:\\ProgramData\\Microsoft\\Windows Defender\\Platform';
+  const DEFENDER_FALLBACK_PATH = 'C:\\Program Files\\Windows Defender\\MpCmdRun.exe';
```
AlexisMora, line 8: Just out of curiosity, will it work if a user have the path in other languages?

src/apps/main/antivirus/windows-defender/find-mcp-command.ts
```diff
@@ -0,0 +1,31 @@
...
+      .filter((name) => statSync(join(DEFENDER_PLATFORM_PATH, name)).isDirectory())
+      .sort((a, b) => b.localeCompare(a, undefined, { numeric: true })); // Order by version descending
+
+    if (versions.length > 0) {
+      fullPath = join(DEFENDER_PLATFORM_PATH, versions[0], 'MpCmdRun.exe');
+    }
+  }
+
+  if (!existsSync(fullPath)) {
+    throw new Error('MpCmdRun.exe not found.');
```
AlexisMora: It could be nice to have a specific type of errors for the antivirus, also remember to handle the throwed errors

src/apps/main/antivirus/windows-defender/find-mcp-command.ts
```diff
@@ -0,0 +1,31 @@
...
+export function findMpCmdRun() {
+  const DEFENDER_PLATFORM_PATH = 'C:\\ProgramData\\Microsoft\\Windows Defender\\Platform';
+  const DEFENDER_FALLBACK_PATH = 'C:\\Program Files\\Windows Defender\\MpCmdRun.exe';
+
+  let fullPath = DEFENDER_FALLBACK_PATH;
+
+  if (existsSync(DEFENDER_PLATFORM_PATH)) {
+    const versions = readdirSync(DEFENDER_PLATFORM_PATH)
+      .filter((name) => statSync(join(DEFENDER_PLATFORM_PATH, name)).isDirectory())
+      .sort((a, b) => b.localeCompare(a, undefined, { numeric: true })); // Order by version descending
```
AlexisMora: why is the sorting needed?

src/apps/main/antivirus/windows-defender/find-mcp-command.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Finds the full path to the Windows Defender MpCmdRun.exe executable.
+ * Searches the Defender Platform directory for the latest version, falling back to the default path if necessary.
+ * @returns {string} The full path to MpCmdRun.exe.
+ * @throws {Error} If MpCmdRun.exe cannot be found in any of the expected locations.
+ */
+export function findMpCmdRun() {
+  const DEFENDER_PLATFORM_PATH = 'C:\\ProgramData\\Microsoft\\Windows Defender\\Platform';
+  const DEFENDER_FALLBACK_PATH = 'C:\\Program Files\\Windows Defender\\MpCmdRun.exe';
+
+  let fullPath = DEFENDER_FALLBACK_PATH;
```
AlexisMora: Unless it is strictly necessary, try to write functions in an inmutable manner, because it make the code more easy to read and reduces mental overhead when reading it.

When variables dont change after being defined, the function becomes more predictable and easier to reason about, test and maintain.

src/apps/main/antivirus/windows-defender/parse-virus-names.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Parses the output from MpCmdRun.exe to extract actual virus names
+ * @param stdout Standard output from MpCmdRun.exe
+ * @param stderr Standard error from MpCmdRun.exe
+ * @param isInfected Whether the file is infected
+ * @returns Array of virus names found
+ */
+export function parseVirusNames(stdout: string, stderr: string, isInfected: boolean): string[] {
+  if (!isInfected) {
+    return [];
+  }
```
AlexisMora, line 24: Since this is a param that you are passing, should this function even be called if is not infected? it feels like this should not be the responsability of the function

src/apps/main/antivirus/windows-defender/parse-virus-names.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const output = stdout + stderr;
```
AlexisMora: Try to avoid concatenate strings this way

src/apps/main/antivirus/windows-defender/parse-virus-names.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Parses the output from MpCmdRun.exe to extract actual virus names
+ * @param stdout Standard output from MpCmdRun.exe
+ * @param stderr Standard error from MpCmdRun.exe
+ * @param isInfected Whether the file is infected
+ * @returns Array of virus names found
+ */
+export function parseVirusNames(stdout: string, stderr: string, isInfected: boolean): string[] {
```
AlexisMora: For consistency, we usually pass parameters as an object. It improves readability at the call site, avoids issues with parameter order, and makes the function easier to extend in the future.

You could do something like
```suggestion
interface Props {stdout: string, stderr: string, isInfected: boolean}

export function parseVirusNames({ stdout, stderr, isInfected }: Props) {
```

src/apps/main/antivirus/windows-defender/parse-virus-names.ts
```diff
@@ -0,0 +1,45 @@
...
+  const virusNames: string[] = [];
+
+  const threatPatterns = [
+    /Threat\s+detected:\s+(.+)/gi,
+    /Threat\s+(.+?)\s+was\s+detected/gi,
+    /Found\s+(.+?)\s+threat/gi,
+    /Malware\s+(.+?)\s+detected/gi,
+    /Virus\s+(.+?)\s+found/gi,
+    /Infected\s+with\s+(.+?)([\s\n]|$)/gi,
+  ];
```
AlexisMora: Where did you get this from?

src/apps/main/antivirus/windows-defender/parse-virus-names.ts
```diff
@@ -0,0 +1,45 @@
...
+        if (virusName && !virusNames.includes(virusName)) {
+          virusNames.push(virusName);
+        }
+      }
+    }
+  }
+
+  if (virusNames.length === 0) {
+    virusNames.push('Windows.Defender.Threat.Detected');
+  }
```
AlexisMora: if there are no virus then we push a Threat detected?

src/apps/main/antivirus/windows-defender/parse-virus-names.ts
```diff
@@ -0,0 +1,45 @@
...
+        const virusName = match[1]
+          .trim()
+          .replace(/^file:/i, '')
+          .replace(/["']/g, '');
+        if (virusName && !virusNames.includes(virusName)) {
+          virusNames.push(virusName);
+        }
+      }
+    }
+  }
```
AlexisMora: you can extract this into a function so its easier to test (and more readable)

src/apps/main/antivirus/windows-defender/scan-file.ts
```diff
@@ -0,0 +1,38 @@
+import { spawn } from 'child_process';
+import { ScanResult } from './types';
+import { parseVirusNames } from './parse-virus-names';
+
+export function scanFile(filePath: string, mpCmdRunPath: string): Promise<ScanResult> {
+  return new Promise((resolve, reject) => {
+    const process = spawn(mpCmdRunPath, ['-Scan', '-ScanType', '3', '-File', filePath]);
```
AlexisMora: Just out of curiosity, what is this doing?

src/apps/main/antivirus/windows-defender/scan-file.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    process.stderr.on('data', (data) => {
+      stderr += data.toString();
+    });
+
+    process.on('close', (code) => {
+      const exitCode = code || 0;
+      const isInfected = exitCode === 2;
+
+      // Parse the output to extract actual threat names
```
AlexisMora: I think this comment is not necessary since the statement is pretty self explanatory

src/apps/main/antivirus/windows-defender/scan-file.ts
```diff
@@ -0,0 +1,38 @@
...
+      stdout += data.toString();
+    });
+
+    process.stderr.on('data', (data) => {
+      stderr += data.toString();
+    });
+
+    process.on('close', (code) => {
+      const exitCode = code || 0;
+      const isInfected = exitCode === 2;
```
AlexisMora, line 28: is there any way that we can have a look at the possible codes?

src/apps/main/antivirus/windows-defender/scan-file.ts
```diff
@@ -0,0 +1,38 @@
...
+
+      resolve({
+        file: filePath,
+        isInfected,
+        viruses: detectedViruses,
+      });
+    });
+
+    process.on('error', (error) => {
+      reject(error);
```
AlexisMora: The same, try to not throw errors when possible

src/apps/main/antivirus/windows-defender/test/antivirus-windows-defender.test.ts
```diff
@@ -0,0 +1,88 @@
...
+import { AntivirusWindowsDefender } from '../antivirus-windows-defender';
+import { findMpCmdRun } from '../find-mcp-command';
+import { deepMocked } from 'tests/vitest/utils.helper.test';
+
+// Mock dependencies
+vi.mock('../find-mcp-command');
+vi.mock('../scan-file');
+
+describe('AntivirusWindowsDefender', () => {
+  // Usar deepMocked para tipar correctamente los mocks
```
AlexisMora: Try to avoid comments in spanish

src/apps/main/antivirus/windows-defender/test/antivirus-windows-defender.test.ts
```diff
@@ -0,0 +1,88 @@
+import { describe, expect, it, vi } from 'vitest';
+import { AntivirusWindowsDefender } from '../antivirus-windows-defender';
+import { findMpCmdRun } from '../find-mcp-command';
+import { deepMocked } from 'tests/vitest/utils.helper.test';
+
+// Mock dependencies
+vi.mock('../find-mcp-command');
+vi.mock('../scan-file');
```
AlexisMora: For consistency and better type safety, we mock using import + vi.mock() instead of string-based mocks. This ensures proper type inference and makes the mocks easier to work with in TypeScript.

src/apps/main/antivirus/windows-defender/test/antivirus-windows-defender.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  beforeEach(() => {
+    vi.clearAllMocks();
+    findMpCmdRunMock.mockReturnValue('C:\\mock\\path\\to\\MpCmdRun.exe');
+  });
+
+  describe('createInstance', () => {
+    it('creates and initializes an instance', async () => {
+      // Given
+      const mockMpCmdRunPath = 'C:\\mock\\path\\to\\MpCmdRun.exe';
+      findMpCmdRunMock.mockReturnValue(mockMpCmdRunPath);
```
AlexisMora: This mock is already set in the beforeEach right?, is it necessary to repeat it here?

src/apps/main/antivirus/windows-defender/antivirus-windows-defender.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export class AntivirusWindowsDefender {
+  isInitialized = false;
+  mpCmdRunPath = '';
+
+  static async createInstance(): Promise<AntivirusWindowsDefender> {
+    const instance = new AntivirusWindowsDefender();
+    await instance.initialize();
+    return instance;
+  }
```
AlexisMora: Since you are creating a class, you could try to do a singleton so we always return just one instance

src/apps/main/antivirus/windows-defender/find-mcp-command.ts
egalvis27: In practice, Windows system folders like `C:\ProgramData` and `C:\Program Files` are always stored under those English paths at the file system level, regardless of the system language. Even in localized systems (e.g. Spanish or French), those folder names remain in English internally.

What changes is the **display name shown in Explorer**, not the actual path. So the hardcoded paths in this case should still work reliably on any language.

src/apps/main/antivirus/windows-defender/find-mcp-command.ts
egalvis27: The `Platform` folder contains one or more subfolders named after version numbers like `4.18.24050.7-0`, which correspond to different Defender engine versions.

By sorting the directories in descending order (`b.localeCompare(a, ..., { numeric: true })`), we ensure we pick the latest version available.

This is important because `MpCmdRun.exe` may exist in multiple version folders, but only the most recent one is guaranteed to be fully functional and up to date.

src/apps/main/antivirus/windows-defender/parse-virus-names.ts
egalvis27: I performed an initial scan to verify the structure of the output and then relied on AI to analyze the patterns and extract the expressions that needed to be reviewed. Scanning with Windows Defender is challenging because infected files are immediately removed. The list of expressions was generated by AI based on a few examples found online, although there are no large-scale scan examples available to ensure complete accuracy.

src/apps/main/antivirus/windows-defender/scan-file.ts
egalvis27: image

src/apps/main/antivirus/windows-defender/antivirus-windows-defender.ts
egalvis27: This class is based on ClamAV. If we make changes here, we should also update the corresponding class for ClamAV to prevent inconsistencies when switching between antivirus engines.

Do you agree with that?

src/apps/main/antivirus/windows-defender/scan-file.ts
egalvis27: image

src/apps/main/antivirus/windows-defender/types.ts
```diff
@@ -0,0 +1,12 @@
+/**
```
dajimenezriv: Unnecesary comments. If you have a different opinion, please, write it here:
https://internxt.slack.com/archives/C094D54FY10/p1752653313495669

src/apps/main/antivirus/windows-defender/scan-file.test.ts
```diff
@@ -0,0 +1,135 @@
+import { describe, expect, it, vi } from 'vitest';
+import { scanFile } from './scan-file';
+import * as childProcess from 'child_process';
+import { EventEmitter } from 'events';
+import { deepMocked } from 'tests/vitest/utils.helper.test';
+import { parseVirusNames } from './parse-virus-names';
+
+// Definimos una interfaz para el proceso simulado
```
dajimenezriv: Never a comment in spanish and also the comment is unnecesary.

src/apps/main/antivirus/windows-defender/antivirus-windows-defender.test.ts
```diff
@@ -0,0 +1,100 @@
...
+import { AntivirusWindowsDefender } from './antivirus-windows-defender';
+import { findMpCmdRun } from './find-mcp-command';
+import { deepMocked } from 'tests/vitest/utils.helper.test';
+
+// Mock dependencies
+vi.mock(import('./find-mcp-command'));
+vi.mock(import('./scan-file'));
+
+describe('AntivirusWindowsDefender', () => {
+  const findMpCmdRunMock = deepMocked(findMpCmdRun);
```
dajimenezriv: We are migrating from vi.mock to partialSpyOn from 'utils.helper.test.ts'

src/apps/main/antivirus/windows-defender/parse-virus-names.ts
```diff
@@ -0,0 +1,39 @@
...
+    return [];
+  }
+
+  const extractedVirusNames = extractVirusNamesFromOutput({ output });
+
+  /**
+   * v2.5.6 Esteban Galvis
+   * MpcmdRun.exe in some cases may not return any virus names,
+   * in which case we return a default value indicating a threat was detected.
+   * based on the codes (0x80508023, 0x80508007) validation above.
```
dajimenezriv, line 32: Why if it doesn't return any virus we add a default virus?

src/apps/main/antivirus/windows-defender/scan-file.test.ts
```diff
@@ -0,0 +1,90 @@
+import { describe, expect, it, vi } from 'vitest';
+import { scanFile } from './scan-file';
+import * as childProcess from 'child_process';
+import { EventEmitter } from 'events';
+import { deepMocked, partialSpyOn } from 'tests/vitest/utils.helper.test';
+import * as parseVirusNames from './parse-virus-names';
+
+interface MockChildProcess extends EventEmitter {
```
dajimenezriv: Why don't use mockDeep from 'vitest-mock-extended'?

src/apps/main/antivirus/windows-defender/parse-virus-names.ts
egalvis27: Even if a virus name is not returned, the HR code still indicates the presence of a threat, so the result should still be marked as infected.

src/apps/main/antivirus/windows-defender/extract-virus-names.ts
```diff
@@ -0,0 +1,29 @@
+export function extractVirusNamesFromOutput({ output }: { output: string }) {
```
dajimenezriv: The file name should match the function name. Also for the test file name.

src/apps/main/antivirus/windows-defender/find-mcp-command.test.ts
```diff
@@ -0,0 +1,95 @@
+import { describe, expect, it, vi } from 'vitest';
+import { findMpCmdRun } from './find-mcp-command';
+import { deepMocked } from 'tests/vitest/utils.helper.test';
+import { stat } from '@/infra/file-system/services/stat';
+import { readdirSync, existsSync } from 'fs';
+import { join } from 'path';
+
+vi.mock(import('fs'));
+vi.mock(import('path'));
+vi.mock('@/infra/file-system/services/stat');
```
dajimenezriv: Here you should use partialSpyOn

src/apps/main/antivirus/windows-defender/find-mcp-command.ts
```diff
@@ -0,0 +1,48 @@
...
+import { stat } from '@/infra/file-system/services/stat';
+import { logger } from '@/apps/shared/logger/logger';
+
+export async function findMpCmdRun() {
+  const DEFENDER_PLATFORM_PATH = 'C:\\ProgramData\\Microsoft\\Windows Defender\\Platform';
+  const DEFENDER_FALLBACK_PATH = 'C:\\Program Files\\Windows Defender\\MpCmdRun.exe';
+
+  let platformSubdirs: string[] = [];
+  if (existsSync(DEFENDER_PLATFORM_PATH)) {
+    const entries = readdirSync(DEFENDER_PLATFORM_PATH);
```
dajimenezriv: Change all sync operations that can be async to async await

src/apps/main/antivirus/windows-defender/find-mcp-command.ts
```diff
@@ -0,0 +1,48 @@
...
+    const validDirs = await Promise.all(
+      entries.map(async (name) => {
+        const fullPath = join(DEFENDER_PLATFORM_PATH, name);
+        const { data, error } = await stat({ absolutePath: fullPath });
+        return !error && data?.isDirectory() ? name : null;
```
dajimenezriv: Here you don't need to do `!error` because if you have error then you don't have data and viceversa.

dajimenezriv: You can do:
```typescript
const { data } = await stat({ absolutePath: fullPath });
return data?.isDirectory() ? name : null;
```

src/apps/main/antivirus/windows-defender/parse-virus-names.test.ts
```diff
@@ -0,0 +1,140 @@
+import { describe, expect, it, vi, beforeEach } from 'vitest';
+import { parseVirusNames } from './parse-virus-names';
+
+describe('parseVirusNames', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    vi.restoreAllMocks();
```
dajimenezriv: You don't need to do both

dajimenezriv: Use just clearAllMocks

src/apps/main/antivirus/windows-defender/scan-file.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    spawnMock.mockReturnValue(mockProcess as unknown as childProcess.ChildProcess);
+    // When
+    const result = await scanFile({ filePath, mpCmdRunPath });
+    // Then
+    expect(result.isInfected).toBe(true);
+    expect(result.viruses.length).toBeGreaterThan(0);
+  });
+});
+
+function createMockChildProcessImmediate(exitCode: number, stdout: string, stderr: string): MockChildProcess {
```
dajimenezriv, line 78: Isn't there any way to test instead of creating this mock function? I haven't tried in this case, but I'm never needed to create a mock function for a test.

PR #896, opened 2025-07-14
Move delete to C++

dajimenezriv-internxt:


src/backend/features/local-sync/watcher/events/unlink/is-move-event.ts
```diff
@@ -0,0 +1,34 @@
+import { FileUuid } from '@/apps/main/database/entities/DriveFile';
```
dajimenezriv, line 1: We use trackAddEvent and trackAddDirEvent on add event and add-dir event, so when we have an unlink event, we check if there has been an add event to know if it's unlink or move.

src/backend/features/local-sync/watcher/events/unlink/unlink-file.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export async function unlinkFile({ virtualDrive, absolutePath }: TProps) {
+  const path = pathUtils.absoluteToRelative({
+    base: virtualDrive.syncRootPath,
+    path: absolutePath,
+  });
+
+  try {
+    const parentUuid = getParentUuid({ path, virtualDrive });
+    if (!parentUuid) return;
```
dajimenezriv, line 24: If the parentUuid doesn't exist it means that the root unlink event is in another place.

src/backend/features/local-sync/watcher/events/unlink/unlink-file.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    const nameWithExtension = basename(path);
+    const { data: file } = await ipcRendererSqlite.invoke('fileGetByName', { parentUuid, nameWithExtension });
```
dajimenezriv, line 27: We need to obtain the uuid of the file, so we need to find it in the database using the parentUuid and the name of the file/folder. This is the same as in the file `unlink-folder.ts`.

src/backend/features/local-sync/watcher/events/unlink/unlink-file.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    if (!file) {
```
dajimenezriv, line 29: This should never happen, because we shouldn't have an item that has a placeholder but it's not in the sqlite.

src/backend/features/local-sync/watcher/events/unlink/unlink-file.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (isMove) return;
+
+    logger.debug({ tag: 'SYNC-ENGINE', msg: 'File unlinked', path });
+
+    const { error } = await ipcRendererDriveServerWip.invoke('storageDeleteFileByUuid', {
+      uuid: file.uuid,
+      workspaceToken: getConfig().workspaceToken,
+    });
+
+    if (error) throw error;
```
AlexisMora, line 44: This means that, the file can be "unlinked" but we can recieve am error while deleting a file right? if we get an error on this function the "unlink" is already done right?

dajimenezriv: Yes, exactly, however since the file/folder has not been deleted in remote it will be recreated in the next sync. Here, we could do a rollback, but for folders is more difficult so I will let it wait until next sync. This is more a problem with move, there probably we should do a rollback, because otherwise a file/folder will be duplicated. However, we want to wait until Andres joins the endpoint of rename and move as talked with Larry, otherwise is a bit complicated because maybe rename works and move not or viceversa.

PR #897, opened 2025-07-15
Load in memory paths

dajimenezriv-internxt:


src/apps/sync-engine/BindingManager.ts
```diff
@@ -98,8 +100,7 @@ export class BindingsManager {
 
   async load(tree: Tree): Promise<void> {
     const addFilePromises = tree.files.map((file) => this.container.fileRepository.add(file));
-    const addFolderPromises = tree.folders.map((folder) => this.container.folderRepository.add(folder));
```
dajimenezriv, line 101: This was loading folders into the InMemory, no longer necessary.

src/apps/sync-engine/BindingManager.ts
```diff
@@ -113,10 +114,21 @@ export class BindingsManager {
...
+
+    deleteItemPlaceholders({
+      remotes: tree.trashedFiles,
+      virtualDrive: this.container.virtualDrive,
+      isFolder: false,
+    });
+
+    const { folders } = await loadInMemoryPaths({ drive: this.container.virtualDrive });
     await Promise.all([
-      this.container.filesPlaceholderDeleter.run(tree.trashedFiles),
```
dajimenezriv, line 117: We extract the delete placeholders outside since they are not promises.

src/apps/sync-engine/BindingManager.ts
```diff
@@ -113,10 +114,21 @@ export class BindingsManager {
...
+      isFolder: true,
+    });
+
+    deleteItemPlaceholders({
+      remotes: tree.trashedFiles,
+      virtualDrive: this.container.virtualDrive,
+      isFolder: false,
+    });
+
+    const { folders } = await loadInMemoryPaths({ drive: this.container.virtualDrive });
```
dajimenezriv, line 129: We retrieve from the file explorer a Map pointing each file/folder uuid to an absolute path

src/apps/sync-engine/index.ts
```diff
@@ -52,29 +50,9 @@ async function setUp() {
     }
   });
 
-  ipcRenderer.on('UPDATE_FILE_PLACEHOLDER', async (_, fileAttributes: FileAttributes) => {
```
dajimenezriv, line 55: This was the place were we were creating the placeholders as soon as we were retriving it. However, this has the problem that it can access to a legacy Map of uuids and absolute paths, and here we need to use the folder-store to obtain the path of the item. So, it's a shit. We are going to remove all of this, and the placeholders are going to be created only in the other sync, the background sync, the one that iterates per each folder because that one already has the absolute path. We have `sync by checkpoint` and `sync by folder` which is currently disabled. We are going to fix it, enable it and create the placeholders on fly with this second sync.

src/backend/features/local-sync/watcher/events/unlink/unlink-file.ts
```diff
@@ -26,6 +26,11 @@ export async function unlinkFile({ virtualDrive, absolutePath }: TProps) {
     const nameWithExtension = basename(path);
     const { data: file } = await ipcRendererSqlite.invoke('fileGetByName', { parentUuid, nameWithExtension });
 
+    /**
+     * v2.5.6 Daniel Jiménez
+     * TODO: check because this event it's also triggered when we change remove something
+     * in remote and it automatically markes it as TRASHED in sqlite, so this error is always logged.
+     */
     if (!file) {
```
dajimenezriv, line 34: This doesn't cause any error, however it logs everytime a warning to the log.

src/backend/features/remote-sync/file-explorer/has-to-be-moved.ts
```diff
@@ -0,0 +1,30 @@
+import { AbsolutePath } from '@/context/local/localFile/infrastructure/AbsolutePath';
```
dajimenezriv, line 1: This is a function used by folders that is going to be the same for files, that checks if an item has been renamed, moved or both. Basically, to know if a folder has been moved or renamed, the path has to be different. If the parent path is the same it means that only the last part has been changed, which is the name of the item. Otherwise we need to move if the parent uuids from local and remote path are different, otherwise the move root item is in another place

src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
```diff
@@ -0,0 +1,72 @@
...
+    const path = remote.path;
+
+    try {
+      if (path === '/') return;
+
+      const { isValid } = validateWindowsName({ path, name: remote.name });
+      if (!isValid) return;
+
+      const remotePath = this.relativePathToAbsoluteConverter.run(path) as AbsolutePath;
+      const localPath = folders[remote.uuid as FolderUuid];
```
dajimenezriv, line 26: This folders dict stores the Map between uuids and absolute paths in the file explorer.

src/context/virtual-drive/folders/infrastructure/HttpRemoteFolderSystem.ts
```diff
@@ -1,21 +1,22 @@
-import { FolderAttributes } from '../domain/Folder';
```
dajimenezriv, line 1: Same functionality but know we just need to retrive the folder dto.

src/apps/sync-engine/BindingManager.ts
```diff
@@ -98,8 +100,7 @@ export class BindingsManager {
 
   async load(tree: Tree): Promise<void> {
     const addFilePromises = tree.files.map((file) => this.container.fileRepository.add(file));
-    const addFolderPromises = tree.folders.map((folder) => this.container.folderRepository.add(folder));
-    await Promise.all([addFolderPromises, addFilePromises]);
+    await Promise.all([addFilePromises]);
```
AlexisMora, line 103: Isnt addFilePromises already an array? also we can make it void right

src/apps/sync-engine/index.ts
AlexisMora: So, if im not mistaken, both:
`  ipcRenderer.on('UPDATE_FILE_PLACEHOLDER', async () => {});` and `  ipcRenderer.on('UPDATE_FOLDER_PLACEHOLDER', async () => {});` can be removed from this index.ts right?

src/backend/features/local-sync/watcher/events/unlink/unlink-file.ts
```diff
@@ -26,6 +26,11 @@ export async function unlinkFile({ virtualDrive, absolutePath }: TProps) {
     const nameWithExtension = basename(path);
     const { data: file } = await ipcRendererSqlite.invoke('fileGetByName', { parentUuid, nameWithExtension });
 
+    /**
+     * v2.5.6 Daniel Jiménez
+     * TODO: check because this event it's also triggered when we change remove something
+     * in remote and it automatically markes it as TRASHED in sqlite, so this error is always logged.
```
AlexisMora:
```suggestion
     * TODO: check why this event it's also triggered when we change or remove something
     * in remote and it automatically markes it as TRASHED in sqlite, hence why this error is always logged.
```

src/backend/features/remote-sync/file-explorer/delete-item-placeholders.ts
```diff
@@ -0,0 +1,29 @@
...
+  | { remotes: Folder[]; virtualDrive: VirtualDrive; isFolder: true };
+
+export function deleteItemPlaceholders({ remotes, virtualDrive, isFolder }: Props) {
+  for (const remote of remotes) {
+    let localUuid;
+
+    if (isFolder) {
+      localUuid = NodeWin.getFolderUuid({ path: remote.path, drive: virtualDrive }).data;
+    } else {
+      localUuid = NodeWin.getFileUuid({ path: remote.path, drive: virtualDrive }).data;
```
AlexisMora: You could simplify this with a ternary:
```suggestion
    const localUUID = isFolder
      ? NodeWin.getFolderUuid({ path: remote.path, drive: virtualDrive }).data
      : NodeWin.getFileUuid({ path: remote.path, drive: virtualDrive }).data;
```
its shorter and more expressive

src/backend/features/remote-sync/file-explorer/has-to-be-moved.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (remoteParentPath === localParentPath) return true;
+
+  const { data: remoteParentUuid } = NodeWin.getFolderUuid({ drive, path: remoteParentPath });
+  const { data: localParentUuid } = NodeWin.getFolderUuid({ drive, path: localParentPath });
+
+  if (!remoteParentUuid || !localParentUuid) return false;
+
+  const isMoved = remoteParentUuid !== localParentUuid;
+
+  // Moved
```
AlexisMora: This comment is not really necessary

src/backend/features/remote-sync/file-explorer/has-to-be-moved.ts
```diff
@@ -0,0 +1,30 @@
...
+  localPath: AbsolutePath;
+};
+
+export function hasToBeMoved({ drive, remotePath, localPath }: TProps) {
+  if (remotePath === localPath) return false;
+
+  const remoteParentPath = dirname(remotePath);
+  const localParentPath = dirname(localPath);
+
+  // Renamed
```
AlexisMora: Same goes for this, I think this if condition is explanatory enough

src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
```diff
@@ -0,0 +1,72 @@
...
+  constructor(
+    private readonly virtualDrive: VirtualDrive,
+    private readonly relativePathToAbsoluteConverter: RelativePathToAbsoluteConverter,
+  ) {}
+
+  async update({ remote, folders }: { remote: Folder; folders: InMemoryFolders }) {
+    const path = remote.path;
+
+    try {
+      if (path === '/') return;
```
AlexisMora: Maybe we could avoid calling this method if the path is root, and avoid the check for this inside this function, i dont think it is its responsability

src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
```diff
@@ -0,0 +1,72 @@
...
+import { InMemoryFolders } from '../sync-items-by-checkpoint/load-in-memory-paths';
+
+export class FolderPlaceholderUpdater {
+  constructor(
+    private readonly virtualDrive: VirtualDrive,
+    private readonly relativePathToAbsoluteConverter: RelativePathToAbsoluteConverter,
+  ) {}
+
+  async update({ remote, folders }: { remote: Folder; folders: InMemoryFolders }) {
+    const path = remote.path;
```
AlexisMora:
```suggestion
    const { path, uuid } = remote
```

src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    try {
+      if (path === '/') return;
+
+      const { isValid } = validateWindowsName({ path, name: remote.name });
+      if (!isValid) return;
+
+      const remotePath = this.relativePathToAbsoluteConverter.run(path) as AbsolutePath;
```
AlexisMora, line 25: Why you need to thad the as? it is not infering the type?

src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
```diff
@@ -0,0 +1,72 @@
...
+
+      if (hasToBeMoved({ drive: this.virtualDrive, remotePath, localPath })) {
+        logger.debug({
+          tag: 'SYNC-ENGINE',
+          msg: 'Moving folder placeholder',
+          remotePath,
+          localPath,
+        });
+
+        await rename(localPath, remotePath);
```
AlexisMora, line 54: if you are not waiting for the result of this, this could be a void

src/backend/features/remote-sync/sync-items-by-checkpoint/load-in-memory-paths.ts
```diff
@@ -0,0 +1,60 @@
...
+          if (uuid) folders[uuid] = absolutePath;
+        }
+
+        if (stats.isFile()) {
+          const { data: uuid } = NodeWin.getFileUuid({ drive, path: absolutePath });
+          if (uuid) files[uuid] = absolutePath;
+        }
+      }
+    }
+  }
```
AlexisMora, line 57: This is just a suggestion, but I think this block of code is doing quite a lot in one place. While it works, its hard to follow at first glance because its very imperative.

This feels like a great candidate for splitting into smaller functions that express **what** is being done, rather than **how**.
What do you think?

src/apps/sync-engine/BindingManager.ts
dajimenezriv: We can't make void, because we need to wait until the InMemory is loaded. Yes, we could use remove the array. Anyway, the InMemory is removed not in the next PR but in the next one. This one deletes the InMemory for folders, the other for files, so the whole load method is removed.

src/apps/sync-engine/index.ts
dajimenezriv: Yes, they are going to be removed and the folder-store. However, the invoke call is already on main and I didn't want to delete all in this PR, just remove the functionality. I'm already working on that.

src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
dajimenezriv: I did it just for path because it was called multiple times.

src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
dajimenezriv: This file has to check it because tree.folders has to contain the root folder, etiher we do it in this method or move it to the run method which is below, I don't care

src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
dajimenezriv: No, Folder is the god object Folder and path is a string, so we cannot use the pathUtils to convert to AbsolutePath

src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
dajimenezriv: We are waiting, if it throws an error we need to catch it

src/backend/features/remote-sync/sync-items-by-checkpoint/load-in-memory-paths.ts
dajimenezriv: The thing is that since we have folderPaths, files and folders variables, if we extract, the complexity of passing the variables and assigning the return to the dict and so is bigger. We already had this separated in another place, but if we want to separate we need to return from the other function the absolutePath, if it's folder or file and if in case of being a folder has an uuid, to know to which dict and array add the absolutePath, so the conditions are going to be practically the same.

src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
dajimenezriv: What I can do is const { path } = remote if you want.

AlexisMora: Okay thats fine as well

src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
AlexisMora: I would call it outside

AlexisMora (approved):

PR #898, opened 2025-07-15
Remove non used code

dajimenezriv-internxt:
## What

This is just a PR preparing code that is no longer used to remove the InMemoryFile finally in the next PR.


src/apps/sync-engine/callbacks-controllers/controllers/DownloadFileController.ts
```diff
@@ -3,30 +3,26 @@ import { ContentsDownloader } from '../../../../context/virtual-drive/contents/a
...
   private async action(uuid: string, callback: CallbackDownload): Promise<string> {
-    const file = this.fileFinderByUuid({ uuid });
+    const file = await this.fileFinderByUuid({ uuid });
 
-    Logger.info('[Begin] Download: ', file.path);
     return await this.downloader.run(file, callback);
   }
 
-  fileFinderByUuid({ uuid }: { uuid: string }) {
-    const file = this.repository.searchByPartial({ uuid });
```
dajimenezriv, line 29: This is the only real change in the PR. Instead of looking for the file in the InMemory we look in the sqlite

PR #899, opened 2025-07-15
Fix/get usage unauthorized errors

AlexisMora:
## What has been changed/added
- Now by default we will create the auth window pointing to '/login' to better check if a user is authenticated and prevent to render the widget if a user is not authenticated
- Extracted the route guard that was in charge of redirecting to '/login' route whenever the session was missing into its own file(Now called AuthGuard)
- Created WidgetSkeleton component loader that mimics actual widget structure (This will only be seen when mounting the Widget component)
-  Refactored Login component
    - Split into focused subcomponents: CredentialsSection and TwoFASection
    - Extracted UsageIndicator into its own component
  - Set up renderer-specific test configuration with jsdom

## Why
- Initially, we added the event `eventBus.emit('USER_LOGGED_OUT');` on the client.ts when the client got a unauthorized request, that event, among other things, was creating the auth window, that was pointing out to the '/' route. In the frontend, that route was mapped to the widget, this causing an infinite loop, because the widget was rendering, starting to send requests  about the usage/limit, those requests got unauthorized....Causing the infinite loop.


src/apps/main/windows/auth.ts
```diff
@@ -25,7 +25,7 @@ export const createAuthWindow = async () => {
     skipTaskbar: true,
   });
 
-  const authLoaded = authWindow.loadURL(resolveHtmlPath(''));
+  const authLoaded = authWindow.loadURL(resolveHtmlPath('login'));
```
dajimenezriv, line 28: What is that for?

src/apps/renderer/App.tsx
```diff
@@ -24,42 +25,16 @@ function LocationWrapper({ children }: { children: JSX.Element }) {
...
-function Loader() {
-  return <></>;
-}
-
 export default function App() {
   useLanguageChangedListener();
 
   return (
     <Router>
-      <Suspense fallback={<Loader />}>
```
dajimenezriv, line 58: Why did you removed the Loader?

src/apps/renderer/pages/Widget/Header.tsx
```diff
@@ -85,7 +65,7 @@ const Header: React.FC<HeadersProps> = ({ setIsLogoutModalOpen }) => {
           <p className="truncate text-sm font-medium text-gray-100" title={user?.email}>
             {user?.email}
           </p>
-          <p className="text-xs text-gray-50">{displayUsage}</p>
+          {user && <UsageIndicator />}
```
dajimenezriv, line 68: I assume here is the solution because if we don't have user we don't call UsageIndicator right?

src/apps/renderer/components/UsageIndicator/UsageIndicator.test.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+// Mock the hooks
+const mockTranslate = vi.fn();
+const mockUseUsage = vi.fn();
+
+vi.mock('../../context/LocalContext', () => ({
+  useTranslationContext: () => ({
+    translate: mockTranslate,
+  }),
+}));
```
dajimenezriv: Cannot mock this and below in a type-safety way?

src/apps/main/windows/auth.ts
AlexisMora: This is one of the key changes for the solution, because this what is doing is opening a new window on '/login'.
This createAuthWindow method is only called when a user is not logged in, and really opening a new window  pointing out to the widget view, doesnt really make sense

src/apps/renderer/App.tsx
AlexisMora: Because Loader was just an empty react fragment (<></>), and now we either use the WidgetSkeleton when we are loading the widget or we just load the window that we are trying to render.

src/apps/renderer/pages/Widget/Header.tsx
AlexisMora: Yeah among other things, it does not make any sense to render this component if we dont have the data of the user, this will avoid unnecessary calls.

dajimenezriv-internxt (approved):

Issue #545, opened 2025-02-02
[Proposal] Move function params to objects


dajimenezriv-internxt:
## What

I have a proposal for new typescript code and it will be nice to discuss it here. @sg-gs
I will mention people that I'm sure they are currently in drive-desktop. @cg27shokworks @ArceDanielShok @xabg2 

One thing that is very easy to implement and does and make a big difference is moving from multiple params to object parameter.

```typescript
// old way
function connect(host: string, port: string) {}
// new way
function connect({ host, port }: { host: string, port: string }) {}
// default parameters
function connect({ host, port }: { host: string, port?: string }) {
port = port ?? 5432;
}
```

## Why

1. If we have two params like in the example that are both string, we can call the function with the parameters it the wrong order and it doesn't complain.
2. When we need to add a mandatory parameters and there is already an optional parameter, adding this parameter becomes tricky because we have to check carefully all places where we call this function. With object parameters it doesn't happen.
3. Explicitly set the name of the parameter inside the objects helps us to see that we are sending the correct parameter.

## Resources

https://www.reddit.com/r/typescript/comments/16vb496/multiple_params_vs_object_function_parameter/
https://medium.com/hackernoon/follow-single-argument-principle-in-typescript-with-parameter-objects-c8a259dd7191

PR #901, opened 2025-07-16
fix: use logic to show loader until fetching is actually done

cg-shokworks:
This PR is to fix an issue when fetching the Folders available for backup. Even when the endpoint that returns all the folders available hasn't finished yet, the user would see a message "No folders available for download", making them think that there were actually no folders, and after a couple seconds the folders would appear.

Now the loader will be shown until the folders fetching is finalized and the user will see them inmediately


src/apps/renderer/hooks/folders/useGetItems.tsx
```diff
@@ -1,26 +1,15 @@
-import { useEffect, useState } from 'react';
```
AlexisMora, line 1: We could try and rename this hook and give it a more descriptive name, what do you think?

src/apps/renderer/hooks/folders/useGetItems.tsx
```diff
@@ -1,26 +1,15 @@
...
-      setItems([]);
-      window.electron.getItemByFolderUuid(folderUuid).then((fetchedItems) => {
-        cache.set(folderUuid, { data: fetchedItems, timestamp: Date.now() });
-        setItems(fetchedItems);
-      });
-    }
-  }, [folderUuid]);
-
-  return items;
+  return { items: items ?? [], status };
```
AlexisMora: There is one issue I faced this week with the `status`,  If you return it like this, when you use the hook in a component, typescript will infer that status is just a string, and not a specific union of 'loading' | 'error' | 'ready'

To solve this issue you should type status so typescript knows it can only be one of those values,

src/apps/renderer/pages/Settings/Backups/Selector/LoadingFolders.tsx
```diff
@@ -25,7 +26,7 @@ export function LoadingFolders({ state, messageText }: LoadingFoldersProps) {
 
   return (
     <div className="flex h-full items-center justify-center">
-      {isLoading ? (
+      {isLoading || loadingItems ? (
```
AlexisMora: maybe the isLoading sate can be deleted since you are already passing a smilar value from props

src/apps/renderer/api/use-get-items.ts
```diff
@@ -0,0 +1,12 @@
+import { useQuery } from '@tanstack/react-query';
+import { queryKeys } from '../core/tanstack-query/query-keys';
+
+const CACHE_DURATION = 2 * 60 * 1000;
+
+export const useItems = (folderUuid: string) => {
```
dajimenezriv: Remember to use object props https://github.com/internxt/drive-desktop/issues/545

src/apps/renderer/core/tanstack-query/query-keys.ts
```diff
@@ -1,3 +1,4 @@
 export const queryKeys = {
   usage: () => ['usage'],
+  items: (folderUuid: string) => ['items', folderUuid],
```
dajimenezriv: Here also object props: https://github.com/internxt/drive-desktop/issues/545

src/apps/renderer/hooks/folders/useGetItems.tsx
dajimenezriv: I just checked and useQuery already returns an status variable so we don't have to do the useMemo

dajimenezriv: So I think we can delete this whole file, and we can do the same for getUsage @AlexisMora

src/apps/renderer/hooks/folders/useGetItems.tsx
cg-shokworks: Sure will do

src/apps/renderer/pages/Settings/Backups/Selector/LoadingFolders.tsx
cg-shokworks: I will check that, my concern is that this is being used in another component so I will check how it affects

src/apps/renderer/hooks/folders/useGetItems.tsx
cg-shokworks: Ok let me see if that works, in that case we could use directly the file that configures from tanstack and that's it

src/apps/renderer/api/use-get-items.ts
cg-shokworks: done

src/apps/renderer/core/tanstack-query/query-keys.ts
cg-shokworks: done

src/apps/renderer/pages/Settings/Backups/Selector/LoadingFolders.tsx
cg-shokworks: I think we're fine to remove it, done

src/apps/renderer/hooks/folders/useGetItems.tsx
cg-shokworks: Yes actually it works, just updated without the hook

src/apps/renderer/hooks/folders/useGetItems.tsx
cg-shokworks: Done!

dajimenezriv-internxt (approved):

src/apps/renderer/pages/Settings/Backups/Selector/DownloadSelector.tsx
```diff
@@ -118,10 +118,14 @@ export default function DownloadFolderSelector({ onClose }: DownloadFolderSelect
         </div>
       </div>
       <div className="border-l-neutral-30 h-72 overflow-y-auto rounded-lg border border-gray-20 bg-white dark:bg-black">
-        {selected && items.length > 0 ? (
+        {selected && items && items.length > 0 && itemsStatus === 'success' ? (
```
dajimenezriv: I think this `itemStatus === 'success'` is not necessary, because you are already checking if items is defined.

cg-shokworks: done

AlexisMora (approved):

PR #910, opened 2025-07-21
Feature/remove axios

cg-shokworks:
The idea of this PR is to remove axios and use node native fetch.

Axios was only used for a legacy file download for the backups.

Update: I updated the PR 
 - Separated files from request.ts into different files so testing is easier
 - Created tests for the functions
 - Some interfaces were updated to types
 - Using object props


src/apps/main/network/get-auth-from-credentials.ts
```diff
@@ -0,0 +1,19 @@
+import { NetworkCredentials } from './types';
+import { createHash } from 'crypto';
+
+type Props = {
+  creds: NetworkCredentials;
+};
+
+export function getAuthFromCredentials({ creds }: Props): { Authorization: string } {
```
dajimenezriv: We just put return types for basic types.

src/apps/main/network/get-auth-from-credentials.ts
```diff
@@ -0,0 +1,19 @@
...
+  const username = creds.user;
+  const password = sha256(Buffer.from(creds.pass)).toString('hex');
+  const base64 = Buffer.from(`${username}:${password}`).toString('base64');
```
dajimenezriv, line 7: Why did you do `Buffer.from(`${username}:${password}`).toString('base64');`, I don't see it in the original implementation.

src/apps/main/network/get-file-info-with-auth.test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const creds = { user: 'test', pass: 'secret' };
+
+describe('get-file-info-with-auth', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('should call getFileInfo with correct headers and return file info', async () => {
+    const getFileInfoSpy = vi.spyOn(getFileInfoModule, 'getFileInfo').mockResolvedValue(mockFileInfo);
```
dajimenezriv: Use partialSpyOn, in different lines and then mock the value.

src/apps/main/network/get-file-info-with-auth.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    const getAuthSpy = vi.spyOn(getAuthModule, 'getAuthFromCredentials').mockReturnValue({ Authorization: 'Basic test' });
+
+    const result = await getFileInfoWithAuth({ bucketId: 'b', fileId: 'id', creds });
+
+    expect(getAuthSpy).toHaveBeenCalledWith({ creds });
```
dajimenezriv: You can do toBeCalledWith

src/apps/main/network/get-file-mirrors.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+describe('get-file-mirrors', () => {
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  const props = mockProps<typeof getFileMirrors>({});
+
+  it('should fetch and return mirrors', async () => {
+    // Given
```
dajimenezriv: Use Given, Where and Then always or never, but not sometimes. And remove the new lines before When and Then, they are not necessary

src/apps/main/network/get-auth-from-credentials.ts
cg-shokworks: Ok will change

src/apps/main/network/get-auth-from-credentials.ts
cg-shokworks: Because in the original implementation this is done behind the scenes by the Axios library. But since we're doing this with fetch, we need to do this manually

src/apps/main/network/get-file-mirrors.test.ts
cg-shokworks: Will change

src/apps/main/network/get-file-info-with-auth.test.ts
cg-shokworks: I think these are just the same, did a little research and one is an alias of the other, will change anyways

src/apps/main/network/get-auth-from-credentials.ts
cg-shokworks: Done

src/apps/main/network/get-file-info-with-auth.test.ts
cg-shokworks: Done

src/apps/main/network/get-file-info-with-auth.test.ts
cg-shokworks: Done

src/apps/main/network/get-file-mirrors.test.ts
cg-shokworks: Done

PR #911, opened 2025-07-21
Migrate env bridge url

dajimenezriv-internxt:
## What

Migrate env variable BRIDGE_URL and remove old DRIVE_URL. Also, remove aliases in webpack and tsconfig for local libraries since we can use `npm link` or similar options.


package.json
```diff
@@ -168,6 +168,7 @@
     "async": "^3.2.4",
     "axios": "^1.1.3",
     "better-sqlite3": "^11.10.0",
+    "bip39": "^3.1.0",
```
AlexisMora, line 171: Why is this dependency needed?

AlexisMora (approved):

PR #915, opened 2025-07-22
Feat/Create antivirus manager

egalvis27:
- Improved Architecture: Implemented a system to dynamically select between Windows Defender and ClamAV based on availability.
- Smooth Engine Switching: Added a mechanism to stop one engine and activate the other when needed.
- Resource Cleanup: Ensured proper resource release when switching from ClamAV to Windows Defender.

### DIFF
--- a/src/apps/main/antivirus/antivirus-manager/select-antivirus-engine.ts
+++ b/src/apps/main/antivirus/antivirus-manager/select-antivirus-engine.ts
@@ -17,13 +17,16 @@ export async function selectAntivirusEngine() {
     return 'windows-defender';
   }
 
-  if (!(await checkClamdAvailability())) await initializeClamAV();
-  if (await checkClamdAvailability()) {
-    logger.info({
-      tag: 'ANTIVIRUS',
-      msg: 'ClamAV selected as fallback antivirus',
-    });
-    return 'clamav';
+  const clamavAvailable = await checkClamdAvailability();
+  if (!clamavAvailable) {
+    const { antivirusEnabled } = await initializeClamAV();
+    if (antivirusEnabled) {
+      logger.info({
+        tag: 'ANTIVIRUS',
+        msg: 'ClamAV selected as fallback antivirus',
+      });
+      return 'clamav';
+    }
   }
 
   logger.warn({
--- a/src/apps/main/antivirus/antivirus-manager/select-antivirus-engine.test.ts
+++ b/src/apps/main/antivirus/antivirus-manager/select-antivirus-engine.test.ts
@@ -4,7 +4,6 @@ import { loggerMock } from '@/tests/vitest/mocks.helper.test';
 import * as ipcMainAntivirus from '@/apps/main/ipcs/ipcMainAntivirus';
 import * as initializeAntivirusModule from '../utils/initializeAntivirus';
 import * as clamAVDaemon from '../ClamAVDaemon';
-import * as sleepModule from '@/apps/main/util';
 
 describe('selectAntivirusEngine', () => {
   const isWindowsDefenderActiveMock = partialSpyOn(ipcMainAntivirus, 'isWindowsDefenderRealTimeProtectionActive');
@@ -30,9 +29,8 @@ describe('selectAntivirusEngine', () => {
   it('selects ClamAV when Windows Defender is not available but ClamAV is', async () => {
     // Given
     isWindowsDefenderActiveMock.mockResolvedValue(false);
-    checkClamdAvailabilityMock.mockResolvedValueOnce(false);
+    checkClamdAvailabilityMock.mockResolvedValue(false);
     initializeClamAVMock.mockResolvedValue({ antivirusEnabled: true });
-    checkClamdAvailabilityMock.mockResolvedValueOnce(true);
     // When
     const result = await selectAntivirusEngine();
     // Then
@@ -49,9 +47,8 @@ describe('selectAntivirusEngine', () => {
   it('returns null when neither Windows Defender nor ClamAV are available', async () => {
     // Given
     isWindowsDefenderActiveMock.mockResolvedValue(false);
-    checkClamdAvailabilityMock.mockResolvedValueOnce(false);
+    checkClamdAvailabilityMock.mockResolvedValue(false);
     initializeClamAVMock.mockResolvedValue({ antivirusEnabled: false });
-    checkClamdAvailabilityMock.mockResolvedValueOnce(false);
     // When
     const result = await selectAntivirusEngine();
     // Then


src/apps/main/antivirus/antivirus-manager/antivirus-manager.test.ts
```diff
@@ -0,0 +1,28 @@
+import { AntivirusManager, getAntivirusManager } from './antivirus-manager';
+
+describe('AntivirusManager', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    (AntivirusManager as unknown as { instance: AntivirusManager | null }).instance = null;
```
dajimenezriv: `clearAllMocks` is no longer necessary.
Also, you can do `AntivirusManager.instance = null;`

src/apps/main/antivirus/antivirus-manager/antivirus-manager.ts
```diff
@@ -0,0 +1,23 @@
...
+    }
+    return AntivirusManager.instance;
+  }
+
+  async getActiveEngine() {
+    return await getActiveEngine({ self: this });
+  }
+}
+
+export const getAntivirusManager = () => AntivirusManager.getInstance();
```
dajimenezriv: Why do you create a function that is only used in a test?

src/apps/main/antivirus/antivirus-manager/antivirus-manager.ts
```diff
@@ -0,0 +1,23 @@
+import { AntivirusType, AntivirusEngine } from './types';
```
dajimenezriv, line 1: I'm checking and I don't see the usage of this antivirus manager outside of the test. Remember that we are not doing OOP.

src/apps/main/antivirus/antivirus-manager/create-antivirus-engine.test.ts
```diff
@@ -0,0 +1,65 @@
...
+import { AntivirusClamAV } from '../antivirus-clam-av';
+import { logger } from '@/apps/shared/logger/logger';
+import { AntivirusType } from './types';
+
+describe('createEngine', () => {
+  const mockDefenderInstance = mockDeep<AntivirusWindowsDefender>();
+  const mockClamAVInstance = mockDeep<AntivirusClamAV>();
+  const windowsDefenderCreateInstanceMock = partialSpyOn(AntivirusWindowsDefender, 'createInstance');
+  const clamAVCreateInstanceMock = partialSpyOn(AntivirusClamAV, 'createInstance');
+  const loggerErrorMock = partialSpyOn(logger, 'error');
```
dajimenezriv: `loggerMock` is already mocked globally. You can import `loggerMock` and do `loggerMock.error`

src/apps/main/antivirus/antivirus-manager/create-antivirus-engine.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    expect(clamAVCreateInstanceMock).toBeCalled();
+    expect(loggerErrorMock).toBeCalledWith(
+      expect.objectContaining({
+        msg: expect.stringContaining('Error initializing antivirus engine, using fallback'),
+        exc: mockError,
+      }),
+    );
+  });
+
+  it('throws error for unsupported antivirus type', async () => {
```
dajimenezriv: You shouldn't need to test something that cannot happen because of typescript because it will lead to test so many things that in 99.9999% of cases are not going to happen, so is not worth test them.

src/apps/main/antivirus/antivirus-manager/create-antivirus-engine.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+describe('createEngine', () => {
+  const mockDefenderInstance = mockDeep<AntivirusWindowsDefender>();
+  const mockClamAVInstance = mockDeep<AntivirusClamAV>();
+  const windowsDefenderCreateInstanceMock = partialSpyOn(AntivirusWindowsDefender, 'createInstance');
+  const clamAVCreateInstanceMock = partialSpyOn(AntivirusClamAV, 'createInstance');
+  const loggerErrorMock = partialSpyOn(logger, 'error');
+
+  beforeEach(() => {
+    vi.clearAllMocks();
```
dajimenezriv: You can delete this

src/apps/main/antivirus/antivirus-manager/create-antivirus-engine.ts
```diff
@@ -0,0 +1,25 @@
...
+      });
+      return await AntivirusClamAV.createInstance();
+    }
+  }
+
+  if (type === 'clamav') {
+    return await AntivirusClamAV.createInstance();
+  }
+
+  throw new Error(`Unsupported antivirus type: ${type}`);
```
dajimenezriv: This is not needed.

src/apps/main/antivirus/antivirus-manager/create-antivirus-engine.ts
```diff
@@ -0,0 +1,25 @@
...
+        tag: 'ANTIVIRUS',
+        msg: 'Error initializing antivirus engine, using fallback',
+        exc: error,
+      });
+      return await AntivirusClamAV.createInstance();
+    }
+  }
+
+  if (type === 'clamav') {
+    return await AntivirusClamAV.createInstance();
```
dajimenezriv: And here you can just return the clamAV instance, instead of writing two times it for the catch and the if

src/apps/main/antivirus/antivirus-manager/select-antivirus-engine.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  if (await isWindowsDefenderRealTimeProtectionActive()) {
+    logger.info({
+      tag: 'ANTIVIRUS',
+      msg: 'Default antivirus selected as engine',
+    });
+    return 'windows-defender';
+  }
+
+  if (!(await checkClamdAvailability())) await initializeClamAV();
```
dajimenezriv: Why not store this in a variable?

src/apps/main/antivirus/antivirus-manager/create-antivirus-engine.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (type === 'windows-defender') {
+    try {
+      return await AntivirusWindowsDefender.createInstance();
+    } catch (error) {
+      logger.error({
+        tag: 'ANTIVIRUS',
+        msg: 'Error initializing antivirus engine, using fallback',
+        exc: error,
+      });
+      return await AntivirusClamAV.createInstance();
```
dajimenezriv: You can ignore this return, you are going to end up in the last return

src/apps/main/antivirus/antivirus-manager/select-antivirus-engine.ts
dajimenezriv: initializeClamAV already does checkClamdAvailability, so the sleep and consecuent checkClamdAvailability are not necessary. You can just read the return of initializeClamAV and return "clamav" or null

src/apps/main/antivirus/antivirus-manager/get-active-antivirus-engine.ts
```diff
@@ -0,0 +1,47 @@
...
+      return null;
+    }
+
+    if (self.currentEngine && self.currentType === selectedType) {
+      return self.currentEngine;
+    }
+
+    if (self.currentEngine) {
+      await self.currentEngine.stop();
+      if (self.currentType === 'clamav') clearAntivirus();
```
dajimenezriv, line 26: This is not a required change and it's outside the PR. Just to point that probably we should merge in any moment the functions clearAntivirus and stop in clamAV, I don't know why they are two different functions

src/apps/main/antivirus/antivirus-manager/select-antivirus-engine.ts
egalvis27: In that function, there is no validation of the availability of the clamAV daemon, but there is a validation of payments that will be removed from the function in the next PR.

With this in mind, should you put the validation of clamAV availability in the function that initialises the daemon?

PR #917, opened 2025-07-22
Feature/remove axios 1

cg-shokworks:


src/apps/main/network/get-auth-from-credentials.ts
```diff
@@ -0,0 +1,15 @@
+import { NetworkCredentials } from './requests';
+import { createHash } from 'crypto';
+
+export function getAuthFromCredentials({ creds }: { creds: NetworkCredentials }): { Authorization: string } {
+  const username = creds.user;
+  const password = sha256(Buffer.from(creds.pass)).toString('hex');
+  const base64 = Buffer.from(`${username}:${password}`).toString('base64');
```
dajimenezriv, line 7: Why this line? Where it comes from?

src/apps/main/network/get-file-info.ts
```diff
@@ -0,0 +1,22 @@
...
+  logger.info({ tag: 'BACKUPS', msg: `Fetching file info for bucketId downloadV1: ${bucketId}, fileId: ${fileId}` });
+  const url = `${process.env.BRIDGE_URL}/buckets/${bucketId}/files/${fileId}/info`;
+  const res = await fetch(url, {
+    method: 'GET',
+    headers: {
+      ...(opts?.headers || {}),
+    },
+  });
+
+  if (!res.ok) throw new Error(`Failed to fetch file info: ${res.status} ${res.statusText}`);
```
dajimenezriv: It's better to throw a logged error. You can do throw logger.error({ msg: "whatever", status, statusText })

src/apps/main/network/get-file-info.ts
```diff
@@ -0,0 +1,22 @@
+import { logger } from '@/apps/shared/logger/logger';
+import { FileInfo } from './requests';
+
+type Props = {
+  bucketId: string;
+  fileId: string;
+  opts?: { headers?: Record<string, string> };
```
dajimenezriv: opts is always defined so you can mark it as required and then you can remove below check of opts?.headers

src/apps/main/network/get-file-mirrors.ts
```diff
@@ -0,0 +1,25 @@
+import { Mirror } from './requests';
+
+type Props = {
+  bucketId: string;
+  fileId: string;
+  limit: number;
+  skip: number;
+  excludeNodes?: string[];
+  opts?: { headers?: Record<string, string> };
```
dajimenezriv: Here the sae with opts

src/apps/main/network/get-auth-from-credentials.ts
cg-shokworks: Because in the original implementation this is done behind the scenes by the Axios library. But since we're doing this with fetch, we need to do this manually

src/apps/main/network/get-mirrors.ts
```diff
@@ -0,0 +1,69 @@
+import { getAuthFromCredentials } from './get-auth-from-credentials';
+import { getFileMirrors } from './get-file-mirrors';
+import { replaceMirror } from './replace-mirror';
+import { Mirror, NetworkCredentials } from './requests';
+
+type Props = {
+  bucketId: string;
+  fileId: string;
+  creds: NetworkCredentials | null;
+  token?: string;
```
dajimenezriv: And here token can be removed? I think it's never used

src/apps/main/network/replace-mirror.ts
```diff
@@ -0,0 +1,30 @@
+import { getFileMirrors } from './get-file-mirrors';
+import { Mirror } from './requests';
+
+type Props = {
+  bucketId: string;
+  fileId: string;
+  pointerIndex: number;
+  excludeNodes?: string[];
+  opts?: { headers?: Record<string, string> };
```
dajimenezriv: Here the same with opts

src/apps/main/network/get-file-info.ts
cg-shokworks: done

src/apps/main/network/get-file-info.ts
cg-shokworks: done

src/apps/main/network/get-file-mirrors.ts
cg-shokworks: done

src/apps/main/network/get-mirrors.ts
cg-shokworks: Yep, just removed

src/apps/main/network/replace-mirror.ts
cg-shokworks: done

src/apps/main/network/get-file-info.ts
```diff
@@ -0,0 +1,22 @@
...
+import { FileInfo } from './requests';
+
+type Props = {
+  bucketId: string;
+  fileId: string;
+  opts: { headers?: Record<string, string> };
+};
+
+export async function getFileInfo({ bucketId, fileId, opts }: Props): Promise<FileInfo> {
+  logger.info({ tag: 'BACKUPS', msg: `Fetching file info for bucketId downloadV1: ${bucketId}, fileId: ${fileId}` });
```
dajimenezriv, line 11: Btw, it's not necessary for this time, but it's better to put the variables outside the string because the logger adds colors and they are easier to be identified with a look

PR #918, opened 2025-07-22
Feature/remove axios 2

cg-shokworks:


src/apps/main/network/get-file-info-with-auth.test.ts
```diff
@@ -0,0 +1,40 @@
+import { describe, it, expect, vi, afterEach } from 'vitest';
```
dajimenezriv: I think that this function is not even needed and that you can call `getFileInfo` directly from download

src/apps/main/network/get-file-info.test.ts
```diff
@@ -0,0 +1,34 @@
...
+      index: '0',
+    };
+    global.fetch = vi.fn().mockResolvedValue({ ok: true, json: () => mockFileInfo });
+    const result = await getFileInfo(props);
+    expect(global.fetch).toHaveBeenCalled();
+    expect(result).toEqual(mockFileInfo);
+  });
+
+  it('should throw if fetch fails', async () => {
+    global.fetch = vi.fn().mockResolvedValue({ ok: false, status: 500, statusText: 'fail', json: () => ({}) });
```
dajimenezriv: Here you don't really need to provide json: () => ({})

src/apps/main/network/get-file-info.test.ts
```diff
@@ -0,0 +1,34 @@
...
+
+describe('get-file-info', () => {
+  const props = mockProps<typeof getFileInfo>({
+    bucketId: 'b',
+    fileId: 'id',
+    opts: { headers: {} },
+  });
+
+  it('should fetch and return file info', async () => {
+    const mockFileInfo = {
```
dajimenezriv, line 13: You don't need to mock the wholeFile

src/apps/main/network/get-file-info.test.ts
```diff
@@ -0,0 +1,34 @@
...
+      frame: 'fr',
+      size: 1,
+      id: 'id',
+      created: new Date(),
+      hmac: { value: '', type: '' },
+      index: '0',
+    };
+    global.fetch = vi.fn().mockResolvedValue({ ok: true, json: () => mockFileInfo });
+    const result = await getFileInfo(props);
+    expect(global.fetch).toHaveBeenCalled();
```
dajimenezriv: Since you are already mocking all props you could check with which info is fetch called

src/apps/main/network/get-file-mirrors.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+describe('get-file-mirrors', () => {
+  const props = mockProps<typeof getFileMirrors>({
+    bucketId: 'b',
+    fileId: 'id',
+    opts: { headers: {} },
+  });
+
+  it('should fetch and return mirrors', async () => {
+    const mockMirrors = [
```
dajimenezriv, line 11: Same here, you don't need to provide the whole info, just the needed

src/apps/main/network/get-file-info-with-auth.test.ts
cg-shokworks: We're using getFileInfoWithAuth from the download.ts file
imagen

src/apps/main/network/get-file-mirrors.test.ts
cg-shokworks: done.

I just left minimal properties

src/apps/main/network/get-file-info.test.ts
cg-shokworks: done

src/apps/main/network/get-file-info.test.ts
cg-shokworks: done

src/apps/main/network/get-file-info-with-auth.test.ts
dajimenezriv: That's what I mean, since this function has so little logic I wouldn't create a `getFileInfoWithAuth` to call `getFileInfo`, I would just call `getFileInfo` directly in the screenshot that you sent

cg-shokworks: Oh got it, I will update then. I was just like replicating the previous logic without trying to change or build what was working, but I think in this case we can do that

cg-shokworks: Just updated @dajimenezriv-internxt

src/apps/main/network/get-file-info.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    const expectedUrl = `${process.env.BRIDGE_URL}/buckets/${props.bucketId}/files/${props.fileId}/info`;
+    global.fetch = vi.fn().mockResolvedValue({ ok: true, json: () => mockFileInfo });
+    const result = await getFileInfo(props);
+    expect(global.fetch).toHaveBeenCalledWith(expectedUrl, {
+      method: 'GET',
+      headers: {
+        ...props.opts.headers,
+      },
+    });
+    expect(result).toEqual(mockFileInfo);
```
dajimenezriv: Use toStrictEqual

dajimenezriv: And the same for the other tests

src/apps/main/network/get-mirrors.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  it('should throw if replaced mirror has no url', async () => {
+    const unhealthyMirror = { ...mockMirror, farmer: {} };
+    const noUrlMirror = { ...mockMirror, url: '', farmer: baseFarmer };
+    const getFileMirrorsMock = partialSpyOn(getFileMirrorsModule, 'getFileMirrors');
+    const replaceMirrorMock = partialSpyOn(replaceMirrorModule, 'replaceMirror');
+    getFileMirrorsMock.mockResolvedValueOnce([unhealthyMirror]).mockResolvedValueOnce([noUrlMirror]).mockResolvedValue([]);
+    replaceMirrorMock.mockResolvedValue(noUrlMirror);
+
+    await expect(getMirrors(props)).rejects.toThrow(/Missing download url for shard/);
```
dajimenezriv, line 72: Just curiosity, why do you do a regex here?

dajimenezriv-internxt (approved):

PR #922, opened 2025-07-23
Remove on sync stopped

dajimenezriv-internxt:
## What

Delete the componnent SyncBannerError because it was based on the function `useSyncStopped` or if the reason because the sync stopped was `STOPPED_BY_USER`. However, `useSyncStopped` was returning always `stopReason` as `null` and `STOPPED_BY_USER` doesn't exist. Basically the user cannot stop the sync manually, so this component is no longer used.


src/apps/main/preload.js
```diff
@@ -73,18 +73,6 @@ contextBridge.exposeInMainWorld('electron', {
   getSyncStatus() {
     return ipcRenderer.invoke('get-sync-status');
   },
-  onSyncStatusChanged(func) {
-    const eventName = 'sync-status-changed';
```
dajimenezriv: We can delete all of this because 'sync-status-changed' and 'sync-stopped' didn't exist on the backend so they were not used.

AlexisMora (approved):

PR #924, opened 2025-07-23
Move from ans to log in logger

dajimenezriv-internxt:
## What

Basically we are going to remove `ans` because when the size of the file is big it takes too much time to load colors. Also, using the extension .log already uses colors in vscode.

### DIFF
--- a/src/apps/backups/Backups.ts
+++ b/src/apps/backups/Backups.ts
@@ -52,7 +52,7 @@ export class Backup {
     });
 
     if (filesDiff.dangled.size > 0) {
-      logger.info({
+      logger.debug({
         msg: 'Dangling files found, handling them',
         tag: 'BACKUPS',
       });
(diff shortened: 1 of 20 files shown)


src/apps/backups/Backups.ts
```diff
@@ -52,7 +52,7 @@ export class Backup {
     });
 
     if (filesDiff.dangled.size > 0) {
-      logger.info({
```
dajimenezriv, line 55: Now we just have logger.debug so it's simpler

src/apps/main/logger.ts
```diff
@@ -21,11 +21,15 @@ export function setupElectronLog() {
    */
```
dajimenezriv, line 21: Basically now we are going to use the levels:
`Electron.silly`, `Electron.debug` and `Electron.info`.
We are going to use `Electron.silly` for colored logs, so in `writeFn` we just console.log these logs.
For files we apply a filter to `debug` so `silly` logs aren't write to a file.
Now, `Electron.debug` will add the log to `drive-desktop.log` and `Electron.info` will add to `drive-desktop-important.log` (this is already done).

cg-shokworks (dismissed):

cg-shokworks (approved):

PR #925, opened 2025-07-23
Migrate logger

dajimenezriv-internxt:


src/apps/main/preload.js
```diff
@@ -25,9 +23,9 @@ contextBridge.exposeInMainWorld('electron', {
   },
 
   logger: {
-    info: (rawBody) => Logger.info(inspect(rawBody, { colors: true, depth: Infinity, breakLength: Infinity })),
```
dajimenezriv, line 28: This is only part of the code that still uses electron-log, we need to change the preload to typescript, which is done in the next PR to be able to import our custom logger

AlexisMora (approved):

PR #926, opened 2025-07-24
Never add return types

dajimenezriv-internxt:


CONTRIBUTING
```diff
@@ -54,33 +54,27 @@ function connect({ host, port }: { host: string; port: number }) {}
...
-function getAccessToken(): Promise<string | undefined> {
-  return new Promise((resolve, reject) => {
-    resolve('');
-  });
+function getNumber() {
+  return 8;
 }
 ```
 
-### Stop using event bus
```
dajimenezriv, line 73: We no longer need this. We are already removing event bus and it's needed in the CONTRIBUTING.

src/node-win/watcher/events/on-raw.service.ts
```diff
@@ -36,11 +36,7 @@ export class OnRawService {
           return;
         }
 
-        const action = await this.detectContextMenuAction.execute({ self, details, absolutePath, path });
```
dajimenezriv, line 39: The 'Doble click en el archivo' was returned here and logged. Now, we just log it in the above function.

PR #927, opened 2025-07-24
Feat/allow switching between windows defender and clamav

egalvis27:
- Antivirus Scan: Scans now use the Antivirus Manager to select the most suitable antivirus engine.
- Removed Checks: Eliminated validations related to Windows Defender real-time protection.
- Refactored Initialization: Updated initializeAntivirus to include payment conditions before initializing the antivirus engine.


src/apps/main/antivirus/windows-defender/is-windows-defender-available.ts
```diff
@@ -0,0 +1,14 @@
+import { exec } from 'child_process';
```
dajimenezriv, line 2: Can you check the usage of `import { promisify } from 'util'`. I think is going to be easier to execute exec without using callbacks and `new Promise`.

`const exec = promisify(execCallback);`

src/apps/main/antivirus/antivirus-manager/select-antivirus-engine.ts
```diff
@@ -18,15 +18,13 @@ export async function selectAntivirusEngine() {
...
-    const { antivirusEnabled } = await initializeClamAV();
-    if (antivirusEnabled) {
-      logger.debug({
-        tag: 'ANTIVIRUS',
-        msg: 'ClamAV selected as fallback antivirus',
-      });
-      return 'clamav';
-    }
+  if (!clamavAvailable) await initializeClamAV();
+  if (await checkClamdAvailability()) {
```
dajimenezriv: You don't need to do this check again, just read the result value from initializeClamAV

src/apps/main/antivirus/antivirus-clam-av.ts
```diff
@@ -45,12 +46,16 @@ export class AntivirusClamAV {
 
       this.isInitialized = true;
     } catch (error) {
-      console.error('Error Initializing ClamAV:', error);
+      logger.error({
```
dajimenezriv: You can do throw logger.error so we can have the stack trace using the msg of the logger error

src/apps/main/antivirus/antivirus-manager/select-antivirus-engine.ts
egalvis27: Following your recommendation this morning, I kept only one validation. After running some tests, I noticed that ClamAV was never selected as the antivirus engine. This is because when the full antivirus interface is initialized, ClamAV gets loaded. But with just the negated condition remaining, the code never enters the block that returns ClamAV as the engine to be used.

PR #928, opened 2025-07-25
Fix multiple auth windows

dajimenezriv-internxt:


src/apps/main/auth/handlers.ts
```diff
@@ -79,7 +81,16 @@ export function setupAuthIpcHandlers() {
 }
 
 async function emitUserLoggedIn() {
+  const context: AuthContext = {
```
dajimenezriv, line 84: We create a context with an abort controller that we are going to send using prop drilling so we can stop background tasks.

src/apps/main/auth/handlers.ts
```diff
@@ -79,7 +81,16 @@ export function setupAuthIpcHandlers() {
...
+    abortController: new AbortController(),
+  };
+
+  eventBus.once('USER_LOGGED_OUT', async () => {
```
dajimenezriv, line 88: We create this event for just one time so we don't create multiple auth windows on logout.

src/apps/main/logger.ts
```diff
@@ -26,6 +26,7 @@ export function setupElectronLog() {
    * Levels: silly < debug < verbose < info < log < warn < error
    */
   ElectronLog.transports.file.level = 'debug';
+  ElectronLog.transports.console.format = (message) => [...message.data];
```
dajimenezriv, line 29: We don't need to print the datetime for console logs

src/backend/features/remote-sync/sync-items-by-folder/sync-items-by-folder.ts
```diff
@@ -2,17 +2,22 @@ import { sleep } from '@/apps/main/util';
...
-  context: Config;
+  context: SyncContext;
   rootFolderUuid: string;
 };
 
 export async function syncItemsByFolder({ context, rootFolderUuid }: TProps) {
   const folderUuids = [rootFolderUuid];
 
   while (folderUuids.length > 0) {
+    if (context.abortController.signal.aborted) {
```
dajimenezriv, line 16: Here we use the abort controller to stop the sync by folder background task, otherwise it won't stop after logging out.

AlexisMora (approved):

PR #931, opened 2025-07-29
Fix workspace decrypt mnemonic

dajimenezriv-internxt:


src/apps/main/background-processes/sync-engine/services/spawn-workspace.ts
```diff
@@ -28,27 +28,32 @@ export async function spawnWorkspace({ context, workspace }: TProps) {
...
-    bridgeUser: credentials.credentials.networkUser,
-    bridgePass: credentials.credentials.networkPass,
-  };
-
-  await spawnSyncEngineWorker({ context: syncContext });
+  try {
+    const mnemonic = await decryptMessageWithPrivateKey({
+      encryptedMessage: Buffer.from(workspace.mnemonic, 'base64').toString(),
+      privateKeyInBase64: user.keys.ecc.privateKey,
+      userMnemonic: user.mnemonic,
```
dajimenezriv, line 35: If we cannot decrypt use the user.mnemonic.

src/apps/main/background-processes/sync-engine/services/spawn-workspace.ts
```diff
@@ -28,27 +28,32 @@ export async function spawnWorkspace({ context, workspace }: TProps) {
...
-    bucket: credentials.bucket,
-    bridgeUser: credentials.credentials.networkUser,
-    bridgePass: credentials.credentials.networkPass,
-  };
-
-  await spawnSyncEngineWorker({ context: syncContext });
+  try {
+    const mnemonic = await decryptMessageWithPrivateKey({
+      encryptedMessage: Buffer.from(workspace.mnemonic, 'base64').toString(),
+      privateKeyInBase64: user.keys.ecc.privateKey,
```
dajimenezriv: We have replace this because they have this replaced in web.
https://github.com/internxt/drive-web/blob/0617927804bca066c289e67ec79dfd05f4a98591/src/app/share/services/share.service.ts#L602

src/apps/main/background-processes/sync-engine/services/spawn-workspace.ts
```diff
@@ -28,27 +28,32 @@ export async function spawnWorkspace({ context, workspace }: TProps) {
 
   const user = getUserOrThrow();
```
AlexisMora, line 29: You could add this inside the trycatch right?

dajimenezriv: Technically there are many getUserOrThrow among the whole app because we shouldn't reach that point without being logged. Anyway, in the new AuthContext that we use to pass the abortController I plan to add also the user so we don't to do this `getUserOrThrow`. So, if there is no user the app is going to break in so many parts that this is not going to be a problem

AlexisMora: Fair enough

AlexisMora (approved):

PR #933, opened 2025-07-29
Use issues from core

dajimenezriv-internxt:
## What

Use issues from `drive-desktop-core` package.


src/apps/main/background-processes/issues.ts
```diff
@@ -1,94 +1,9 @@
 import { ipcMain } from 'electron';
```
dajimenezriv, line 1: File moved to `drive-desktop-core`.

src/apps/main/background-processes/process-issues.ts
```diff
@@ -1,38 +1,8 @@
-import { Notification } from 'electron';
```
dajimenezriv, line 1: File moved to `drive-desktop-core`.

PR #935, opened 2025-07-31
feat: improve get backups from device function

cg-shokworks:
This PR contains:

* Adding try/catch to the getBackupsFromDevice function
* Add more logs so we have better traceability on what happens in that function
* Add unit tests for the function so we improve our test %

### DIFF
--- a/src/apps/main/device/backup-folder-uuid.ts
+++ b/src/apps/main/device/backup-folder-uuid.ts
@@ -1,5 +1,6 @@
 import { driveServerWipModule } from '@/infra/drive-server-wip/drive-server-wip.module';
 import configStore from '../config';
+import { logger } from '@/apps/shared/logger/logger';
 
 type TBackup = { enabled: boolean; folderId: number; folderUuid?: string };
 type TBackupList = Record<string, TBackup>;
@@ -19,7 +20,10 @@ export class BackupFolderUuid {
 
   async ensureBackupUuidExists({ backupsList }: { backupsList: TBackupList }): Promise<void> {
     const entries = Object.entries(backupsList);
-
+    logger.debug({
+      tag: 'BACKUPS',
+      msg: `Ensuring backup UUIDs exist for ${entries.length} backups`,
+    });
     const promises = entries.map(async ([pathname, backup]) => {
       if (!backup.folderUuid && backup.enabled) {
         backup.folderUuid = await this.getBackupFolderUuid({ backup });
@@ -28,7 +32,7 @@ export class BackupFolderUuid {
     });
 
     await Promise.all(promises);
-
+    logger.debug({ tag: 'BACKUPS', msg: 'Finished ensuring UUIDs' });
     this.store.set('backupList', backupsList);
   }
 }
--- a/src/apps/main/device/get-backups-from-device.test.ts
+++ b/src/apps/main/device/get-backups-from-device.test.ts
@@ -1,0 +1,62 @@
+import { partialSpyOn, mockProps } from '@/tests/vitest/utils.helper.test';
+import { getBackupsFromDevice } from './get-backups-from-device';
+import * as serviceModule from './service';
+import * as backupFolderUuidModule from './backup-folder-uuid';
+import configStore from '../config';
+import { app } from 'electron';
+
+describe('getBackupsFromDevice', () => {
+  const props = mockProps<typeof getBackupsFromDevice>({ bucket: 'bucket-1' });
+  const backupChild = { id: 1, uuid: 'folder-uuid' };
+  const folder = { children: [backupChild] };
+  const fetchFolderMock = partialSpyOn(serviceModule, 'fetchFolder');
+  const findBackupPathnameFromIdMock = partialSpyOn(serviceModule, 'findBackupPathnameFromId');
+  const backupFolderUuidMock = partialSpyOn(backupFolderUuidModule, 'BackupFolderUuid');
+  const configStoreMock = partialSpyOn(configStore, 'get');
+  const getPathMock = partialSpyOn(app, 'getPath');
+
+  it('should return filtered and mapped backups when isCurrent is true', async () => {
+    fetchFolderMock.mockResolvedValueOnce(folder);
+    findBackupPathnameFromIdMock.mockReturnValueOnce('/path/to/backup1');
+    backupFolderUuidMock.mockImplementationOnce(() => ({
+      ensureBackupUuidExists: vi.fn(),
+    }));
+
+    configStoreMock.mockReturnValueOnce({ '/path/to/backup1': { enabled: true } });
+    getPathMock.mockReturnValueOnce('/tmp');
+
+    const result = await getBackupsFromDevice(props, true);
+    expect(result).toStrictEqual([
+      {
+        ...backupChild,
+        pathname: '/path/to/backup1',
+        folderId: 1,
+        folderUuid: 'folder-uuid',
+        tmpPath: '/tmp',
+        backupsBucket: 'bucket-1',
+      },
+    ]);
+  });
+
+  it('should return mapped backups when isCurrent is false', async () => {
+    fetchFolderMock.mockResolvedValueOnce(folder);
+    const result = await getBackupsFromDevice(props, false);
+    expect(result).toStrictEqual([
+      {
+        ...backupChild,
+        folderId: 1,
+        folderUuid: 'folder-uuid',
+        backupsBucket: 'bucket-1',
+        tmpPath: '',
+        pathname: '',
+      },
+    ]);
+  });
+
+  it('should throw and log error if fetchFolder fails', async () => {
+    const error = new Error('fetch error');
+    fetchFolderMock.mockRejectedValue(error);
+
+    await expect(getBackupsFromDevice(props, true)).rejects.toThrow('fetch error');
+  });
+});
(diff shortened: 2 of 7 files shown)


src/apps/main/device/get-backups-from-device.ts
```diff
@@ -0,0 +1,55 @@
...
+        ...backup,
+        folderId: backup.id,
+        folderUuid: backup.uuid,
+        backupsBucket: device.bucket,
+        tmpPath: '',
+        pathname: '',
+      }));
+    }
+  } catch (error) {
+    logger.error({ tag: 'BACKUPS', msg: 'Error getting backups', error });
```
dajimenezriv: You should `throw logger.error`

dajimenezriv: However, are we handling the error in the frontend or do we want to return an empty array.

package.json
```diff
@@ -16,7 +16,7 @@
     "init:ci": "npm ci --ignore-scripts && node node_modules/electron/install.js",
     "init:dev": "npm install && node node_modules/electron/install.js && npm run build:dll && npm run reload-native-deps",
     "========== Code style ==========": "",
-    "lint": "eslint . --ext .ts,.tsx --max-warnings 505",
```
dajimenezriv: You don't have to increase warnings, you just left some things imported in service.ts that are not used.

src/apps/main/device/backup-folder-uuid.ts
```diff
@@ -19,7 +20,10 @@ export class BackupFolderUuid {
 
   async ensureBackupUuidExists({ backupsList }: { backupsList: TBackupList }): Promise<void> {
     const entries = Object.entries(backupsList);
-
+    logger.debug({
+      tag: 'BACKUPS',
+      msg: `Ensuring backup UUIDs exist for ${entries.length} backups`,
```
dajimenezriv: Put variables in different fields because:
- it will print well if they are objects or not (in this case it's obvious a number), but to make it consistent.
- it will colorize the numbers.

src/apps/main/device/get-backups-from-device.ts
```diff
@@ -0,0 +1,55 @@
...
+import { getUser, setUser } from '../auth/service';
+import { BackupFolderUuid } from './backup-folder-uuid';
+import { BackupInfo } from '@/apps/backups/BackupInfo';
+import { logger } from '@/apps/shared/logger/logger';
+import { app } from 'electron';
+
+export async function getBackupsFromDevice(device: Device, isCurrent?: boolean): Promise<Array<BackupInfo>> {
+  try {
+    logger.debug({ tag: 'BACKUPS', msg: 'Fetching folder for device', deviceUuid: device.uuid, isCurrent });
+    const folder = await fetchFolder({ folderUuid: device.uuid });
```
dajimenezriv: Can you migrate the fetchFolder inside the infra `drive-server-wip`? With that we will have the `clientWrapper` functionality, it will retry, add logs and so, so you won't need these two logs.

src/apps/main/device/get-backups-from-device.ts
```diff
@@ -0,0 +1,55 @@
...
+        .map((backup) => ({
+          ...backup,
+          pathname: backup.pathname as string,
+          folderId: backup.id,
+          folderUuid: backup.uuid,
+          tmpPath: app.getPath('temp'),
+          backupsBucket: device.bucket,
+        }));
+
+      logger.debug({ tag: 'BACKUPS', msg: `Found ${backups.length} enabled backups` });
```
dajimenezriv: Same here about backups.length

src/apps/main/device/get-backups-from-device.ts
cg-shokworks: We're handling it in the FE

src/infra/drive-server-wip/services/backup/fetch-folder.test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    clientMock.GET.mockResolvedValue({ data: mockData });
+
+    const { data, error } = await fetchFolder({ folderUuid: 'folder-uuid' });
+
+    expect(error).toBeUndefined();
+    expect(data).toEqual(mockData);
+  });
+
+  it('should return error when folder is not found (404)', async () => {
```
dajimenezriv, line 20: I think I understand more or less why you are doing this test, however, from what I see you are testing a 404 status code, but does this endpoint can return a 404? If so and you want to add a test, we have to handle the error in the endpoint because we are using typed error from the infra layer. You can see an example in `create-device.ts`. That taking into account that if the folder is not retrieved it returns a 404. I assume it's true, but I don't know.

src/apps/main/device/get-backups-from-device.test.ts
```diff
@@ -0,0 +1,73 @@
+import { partialSpyOn, mockProps, deepMocked } from '@/tests/vitest/utils.helper.test';
+import { getBackupsFromDevice } from './get-backups-from-device';
+import * as serviceModule from './service';
+import * as backupFolderUuidModule from './backup-folder-uuid';
+import { driveServerWip } from '@/infra/drive-server-wip/drive-server-wip.module';
+import configStore from '../config';
+import { app } from 'electron';
+
+vi.mock(import('@/infra/drive-server-wip/drive-server-wip.module'));
```
dajimenezriv: Why don't use partialSpyOn here?

src/infra/drive-server-wip/services/backup/fetch-folder.ts
```diff
@@ -0,0 +1,29 @@
...
+  const method = 'GET';
+  const endpoint = '/folders/content/{uuid}';
+  const key = getRequestKey({ method, endpoint, context });
+
+  const promiseFn = () =>
+    client.GET(endpoint, {
+      params: { path: { uuid: context.folderUuid } },
+    });
+
+  const { data, error } = await clientWrapper({
```
dajimenezriv: Here also you are losing type safety. It's better to do `return await`. Because now you are telling that you always have data and error, when it's not true.

dajimenezriv: You can test it by hover in the function and see that the return type is not the same doing `return await` that this.

src/apps/main/device/service.ts
```diff
@@ -316,21 +280,15 @@ export async function addBackup(): Promise<void> {
...
-
-  return res.data;
-}
-
 async function fetchFolders({ folderUuids }: { folderUuids: string[] }) {
-  const folders = await Promise.all(folderUuids.map((folderUuid) => fetchFolder({ folderUuid })));
-  return folders;
+  const results = await Promise.all(
+    folderUuids.map(async (folderUuid) => {
+      const { data, error } = await driveServerWipModule.backup.fetchFolder({ folderUuid });
```
dajimenezriv: Same here with return await

src/apps/main/device/service.ts
```diff
@@ -298,10 +261,11 @@ export async function addBackup(): Promise<void> {
...
     let folderStillExists;
-    try {
-      const existFolder = await fetchFolder({ folderUuid: existingBackup.folderUuid });
-      folderStillExists = !existFolder.removed;
-    } catch {
+    const { data, error } = await driveServerWipModule.backup.fetchFolder({ folderUuid: existingBackup.folderUuid });
+
+    if (data) {
+      folderStillExists = !data.removed;
+    } else if (error) {
```
dajimenezriv: And here if you change to `return await` in fetchFolder, you don't need else if, because it's data or error, so can you can do just else

src/apps/main/device/get-backups-from-device.ts
```diff
@@ -0,0 +1,60 @@
...
+import { BackupInfo } from '@/apps/backups/BackupInfo';
+import { logger } from '@/apps/shared/logger/logger';
+import { app } from 'electron';
+import { driveServerWipModule } from '@/infra/drive-server-wip/drive-server-wip.module';
+
+export async function getBackupsFromDevice(device: Device, isCurrent?: boolean): Promise<Array<BackupInfo>> {
+  try {
+    const { data: folder, error } = await driveServerWipModule.backup.fetchFolder({ folderUuid: device.uuid });
+
+    if (error || !folder) {
```
dajimenezriv: Here two things. The client wrapper used in the drive-server-wip module already logs when a request is wrong, so here you can just do:
`ìf (error) throw error`
Because it's already logged and also you don't need to check if folder doesn't exist.

src/apps/main/device/get-backups-from-device.test.ts
cg-shokworks: Yes, I tried to do that but it was not working so I guided myself by some tests in service.ts where it's being mocked. 

However, I just tried to use partialSpyOn instead of deepMock and it's working now

src/infra/drive-server-wip/services/backup/fetch-folder.test.ts
cg-shokworks: Yes I tested putting a random UUID, and it returns 404

image

cg-shokworks: Done!

src/infra/drive-server-wip/services/backup/fetch-folder.ts
cg-shokworks: Done!

src/apps/main/device/service.ts
cg-shokworks: Done

src/apps/main/device/service.ts
cg-shokworks: Done

src/apps/main/device/get-backups-from-device.ts
cg-shokworks: Done

PR #936, opened 2025-07-31
Remove file modified error from sync

dajimenezriv-internxt:
## What

Remove FILE_MODIFIED error from issues in sync since we try to upload it again automatically so we don't need to show the error. Also, for big files, when we modify it we receive many events for each of it's chunks. Now we add like a debounce so we wait until we have received the last chunk before start uploading it.


src/node-win/watcher/detect-context-menu-action.service.ts
```diff
@@ -14,51 +14,49 @@ type TProps = {
   path: RelativePath;
```
dajimenezriv, line 14: Same code, just moved outside a class.

src/node-win/watcher/events/on-raw.service.ts
```diff
@@ -1,4 +1,4 @@
-import { DetectContextMenuActionService } from '../detect-context-menu-action.service';
```
dajimenezriv, line 1: Same code, just moved outside a class.

AlexisMora (approved):

PR #937, opened 2025-08-01
Automate windows build

dajimenezriv-internxt:
## What

1. Automate windows build in a github action pipeline.
2. Don't use the user mnemonic if we can't decrypt the workspace mnemonic.
3. Create a script to load clamAV.
4. Remove ANALYZE and ROOT_FOLDER_NAME from environment variables.


.erb/configs/webpack.config.main.prod.ts
```diff
@@ -38,7 +38,7 @@ const configuration: webpack.Configuration = {
   },
 
   plugins: [
-    new BundleAnalyzerPlugin({ analyzerMode: process.env.ANALYZE === 'true' ? 'server' : 'disabled' }),
```
dajimenezriv: This is just to measure the size of the bundle, but we have always in false. Whenever we need to measure just uncomment this line and the one from the file below.

README.md
```diff
@@ -1,11 +1,4 @@
...
...ic=reliability_rating)](https://sonarcloud.io/summary/new_code?id=internxt_drive-desktop) [![Vulnerabilities](https://sonarcloud.io/api/project_badges/measure?project=internxt_drive-desktop&metric=vulnerabilities)](https://sonarcloud.io/summary/new_code?id=internxt_drive-desktop) [![Code Smells](https://sonarcloud.io/api/project_badges/measure?project=internxt_drive-desktop&metric=code_smells)](https://sonarcloud.io/summary/new_code?id=internxt_drive-desktop) [![Duplicated Lines (%)](https://sonarcloud.io/api/project_badges/measure?project=internxt_drive-desktop&metric=duplicated_lines_density)](https://sonarcloud.io/summary/new_code?id=internxt_drive-desktop) [![Technical Debt](https://sonarcloud.io/api/project_badges/measure?project=internxt_drive-desktop&metric=sqale_index)](https://sonarcloud.io/summary/new_code?id=internxt_drive-desktop) [![Coverage](https://sonarcloud.io/api/project_badges/measure?project=internxt_drive-desktop&metric=coverage)](https://sonarcloud.io/summary/new_code?id=internxt_drive-desktop) [![Bugs](https://sonarcloud.io/api/project_badges/measure?project=internxt_drive-desktop&metric=bugs)](https://sonarcloud.io/summary/new_code?id=internxt_drive-desktop)
```
dajimenezriv, line 1: I've added some more metrics and also the logo of node 20.

README.md
```diff
@@ -37,143 +31,13 @@ Install using win store
...
+![alt text](public/image.png)
 
 ## Build Steps
 
 ```bash
 npm run init:dev
+npm run clamav # optional
 npm run start
 ```
-
```
dajimenezriv, line 50: Now we don't need the info for the release. In case we need it it's still in confluence or we can go back to a previous version of the README.

clamAV/clamd.conf
```diff
@@ -0,0 +1,103 @@
+CommandReadTimeout 120
```
dajimenezriv, line 1: This one and the one below are the only two files configurable for clamAV, now they are in the repo.

src/apps/main/antivirus/ClamAVDaemon.ts
```diff
@@ -15,7 +15,7 @@ let timer: NodeJS.Timeout | null = null;
 
 export const startClamdServer = (): Promise<void> => {
   return new Promise((resolve, reject) => {
-    const clamdPath = path.join(RESOURCES_PATH, 'clamd-inxt.exe');
+    const clamdPath = path.join(RESOURCES_PATH, 'clamd.exe');
```
dajimenezriv, line 18: Since now we download in the pipeline the original file was called clamd.exe

.erb/configs/webpack.config.main.prod.ts
sg-gs: Isn't it better to set it to true rather than commenting and uncommenting it?

.erb/configs/webpack.config.renderer.prod.ts
```diff
@@ -102,15 +102,8 @@ const configuration: webpack.Configuration = {
     }),
 
     new Dotenv({ ignoreStub: true }),
-
-    new MiniCssExtractPlugin({
-      filename: 'style.css',
-    }),
-
-    new BundleAnalyzerPlugin({
-      analyzerMode: process.env.ANALYZE === 'true' ? 'server' : 'disabled',
```
sg-gs, line 111: Same here

src/apps/shared/crypto/service.ts
```diff
@@ -25,8 +23,7 @@ export const decryptMessageWithPrivateKey = async ({
   });
 
   if (!comparePrivateKeyCiphertextIDs({ privateKey, message })) {
-    logger.warn({ tag: 'SYNC-ENGINE', msg: 'The key does not correspond to the ciphertext' });
-    return userMnemonic;
+    throw logger.error({ tag: 'SYNC-ENGINE', msg: 'The key does not correspond to the ciphertext' });
```
sg-gs, line 26: The return of the logger error is an error?

README.md
sg-gs: The GitHub Action is the documentation now, although if we can specify in some place (preferably here) where each value can be obtained from the DigiCert platform, better

src/apps/shared/crypto/service.ts
dajimenezriv: Yes https://github.com/internxt/drive-desktop/blob/49590c0d9416ca74065a75c37124d52d34e092a5/src/apps/shared/logger/logger.ts#L101

PR #940, opened 2025-08-03
Add key to sync events

dajimenezriv-internxt:


src/apps/sync-engine/callbacks/fetchData.service.ts
```diff
@@ -1,10 +1,10 @@
-import Logger from 'electron-log';
 import { BindingsManager, CallbackDownload } from '../BindingManager';
 import { FilePlaceholderId } from '../../../context/virtual-drive/files/domain/PlaceholderId';
-import * as fs from 'fs';
 import { dirname } from 'path';
 import { ipcRendererSyncEngine } from '../ipcRendererSyncEngine';
 import { NodeWin } from '@/infra/node-win/node-win.module';
+import { logger } from '@/apps/shared/logger/logger';
+import { unlink } from 'fs/promises';
```
AlexisMora, line 7: Why this unlink is needed instead of the `fs` one?

AlexisMora:
What is the context behind this change? why is this change is needed?

PR #941, opened 2025-08-04
Simplify backups dangled files

dajimenezriv-internxt:
## What

Previously, the dangled files were handled in this way by backups:
1. Check if a file was uploaded between the dangled dates.
2. Check if we can download the file.
3. Upload the file in case of dangled.
However, we need to refactor the download function because the progress is unstable and currently we are sending the progress two times to the frontend because the code is duplicated. Since dangled files are just temporal and is not worth to refactor it, we have remove the second step, now in case the dangled file was between the dates and the patch wasn't applied in the electron-store, we upload the file.


src/apps/backups/diff/calculate-files-diff.ts
```diff
@@ -45,7 +44,14 @@ export function calculateFilesDiff({ local, remote }: TProps) {
...
     if (!isApplied && createdAt >= startDate && createdAt <= endDate) {
-      dangled.set(local, remoteFile);
+      logger.debug({
+        tag: 'BACKUPS',
+        msg: 'Dangled file found',
+        localPath: local.absolutePath,
+        remoteId: remoteFile.contentsId,
+      });
+
+      modified.set(local, remoteFile);
```
dajimenezriv, line 54: Now we consider a dangled file just as a modify one, we just update the contents in cloud, that's all.

src/context/virtual-drive/contents/infrastructure/download/EnvironmentContentFileDownloader.ts
```diff
@@ -36,30 +35,29 @@ export class EnvironmentContentFileDownloader {
...
+        this.state = this.environment.download(
           this.bucket,
           contentsId,
           {
-            progressCallback: (progress: number) => {
+            progressCallback: (progress) => {
               this.eventEmitter.emit('progress', progress);
             },
-            finishedCallback: (err: Error, stream: Readable) => {
+            finishedCallback: (err, stream) => {
```
dajimenezriv, line 45: Now we have increase the version of inxt-js to 2.2.5 and the types are provided.

src/apps/sync-engine/dependency-injection/contents/builder.ts
```diff
@@ -3,7 +3,6 @@ import { SharedContainer } from '../shared/SharedContainer';
 import { ContentsContainer } from './ContentsContainer';
 import { ContentsDownloader } from '../../../../context/virtual-drive/contents/application/ContentsDownloader';
 import { ContentsUploader } from '../../../../context/virtual-drive/contents/application/ContentsUploader';
-import { temporalFolderProvider } from '../../../../context/virtual-drive/contents/application/temporalFolderProvider';
```
dajimenezriv, line 6: This is a function, not a class, so the dependency injection was unnecessary.

AlexisMora (approved):

PR #942, opened 2025-08-04
Solve antivirus call stack error

dajimenezriv-internxt:
## What

We had a problem that when reading some folders the antivirus was throwing an error because of reaching the call stack.

### DIFF
--- a/src/apps/main/antivirus/ManualSystemScan.ts
+++ b/src/apps/main/antivirus/ManualSystemScan.ts
@@ -191,14 +191,9 @@ class ManualSystemScan {
         pathNames = [userSystemPath.path];
       }
 
-      const allFilePaths: string[] = [];
-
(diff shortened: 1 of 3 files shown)


src/apps/main/antivirus/ManualSystemScan.ts
```diff
@@ -191,14 +191,9 @@ class ManualSystemScan {
         pathNames = [userSystemPath.path];
       }
 
-      const allFilePaths: string[] = [];
-
-      await Promise.all(
-        pathNames.map(async (p) => {
-          const filePaths = await getFilesFromDirectory({ rootFolder: p });
-          allFilePaths.push(...filePaths);
```
dajimenezriv, line 199: For some reason the allFilePaths.push was causing the bug, not sure why.

PR #945, opened 2025-08-04
Fix/update on modification date mismatch

egalvis27:


src/backend/features/remote-sync/file-explorer/update-file-placeholder.ts
```diff
@@ -43,15 +46,36 @@ export class FilePlaceholderUpdater {
...
           msg: 'Moving file placeholder',
           remotePath,
           localPath,
         });
 
-        await rename(localPath, remotePath);
+        await rename(localPath.path, remotePath);
+      }
+
+      const remoteTime = Math.floor(remote.modificationTime.getTime() / 1000);
```
dajimenezriv: Why do you divide and then Math.floor?

dajimenezriv: Can you extract also this into a separate function?

src/infra/drive-server-wip/services/files.service.ts
```diff
@@ -118,14 +118,23 @@ async function renameFile(context: { uuid: string; name: string; extension: stri
...
+  };
+
+  if (context.modificationTime) {
+    body.modificationTime = context.modificationTime;
+  }
+
   const promiseFn = () =>
     client.PUT(endpoint, {
-      body: { fileId: context.newContentId, size: context.newSize },
+      body,
```
dajimenezriv: Why not?

```ts
  const promiseFn = () =>
    client.PUT(endpoint, {
      body: {
        fileId: context.newContentId,
        size: context.newSize,
        modificationTime: context.modificationTime,
      },
      params: { path: { uuid: context.uuid } },
    });
```

src/infra/drive-server-wip/services/files.service.ts
```diff
@@ -118,14 +118,23 @@ async function renameFile(context: { uuid: string; name: string; extension: stri
   });
 }
 
-async function replaceFile(context: { uuid: string; newContentId: string; newSize: number }) {
+async function replaceFile(context: { uuid: string; newContentId: string; newSize: number; modificationTime?: Date }) {
```
dajimenezriv: Can you make it mandatory? You can change it also in backups, how we pass the modificationTime. Probably the best option is to not parse it to number at the beginning.
File `FsLocalItemsGenerator.ts`:

```ts
        res.files.push({
          path: absolutePath,
          modificationTime: data.mtime.getTime(),
          size: data.size,
        });
```

Remove the `getTime()` there.

src/backend/features/remote-sync/file-explorer/sync-modified-file.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export async function syncModifiedFile({
+  remoteFile,
+  localFile,
+  relativePathToAbsoluteConverter,
+  fileContentsUploader,
+  virtualDrive,
+}: Props) {
+  const remotePath = relativePathToAbsoluteConverter.run(remoteFile.path) as AbsolutePath;
```
dajimenezriv: You already have remotePath in the previous function.

src/infra/drive-server-wip/services/files.service.ts
egalvis27: The parameter is optional both in the request and in the function. I was concerned that if it wasn’t passed to the function for some reason, it might overwrite the existing date.

dajimenezriv: Why do we want to make it optional? I understand why is optional in the drive-server-wip, so platforms are not forced to update it, but for us? Don't we want to override the existing date?

src/context/virtual-drive/contents/infrastructure/download/EnvironmentContentFileDownloader.ts
```diff
@@ -39,10 +39,10 @@ export class EnvironmentContentFileDownloader {
           this.bucket,
           contentsId,
           {
-            progressCallback: (progress) => {
+            progressCallback: (progress: number) => {
```
dajimenezriv: You don't need to include this type and the one below manually. The library `inxt-js` was updated. Just do `npm install`, or better, bring all new commits from main and do `npm install`.

src/backend/features/remote-sync/sync-items-by-checkpoint/load-in-memory-paths.test.ts
```diff
@@ -37,7 +37,9 @@ describe('load-in-memory-paths', () => {
     // When
     const { files, folders } = await loadInMemoryPaths(props);
     // Then
-    expect(files).toStrictEqual({ fileUuid2: 'C:\\Users\\user\\InternxtDrive\\folder\\file2' });
+    expect(Object.keys(files)).toContain('fileUuid2');
```
dajimenezriv: Maybe here is better to do strictEqual(["fileUuid2"]) so we are sure that only that id is inside

src/apps/sync-engine/BindingManager.ts
```diff
@@ -52,7 +55,7 @@ export class BindingsManager {
 
     this.container.virtualDrive.connectSyncRoot({ callbacks });
 
-    void addPendingItems({ controllers: this.controllers });
+    void addPendingItems({ controllers: this.controllers, fileContentsUploader: this.fileContentsUploader });
```
dajimenezriv: You don't need to create any other variable, you already have the controllers inside `addPendingItems`

src/backend/features/remote-sync/sync-items-by-checkpoint/load-in-memory-paths.test.ts
dajimenezriv: Why did you add a new line instead of the expect just being one?

src/apps/sync-engine/in/sync-modified-file.ts
```diff
@@ -0,0 +1,73 @@
...
+    });
+  }
+}
+
+type SyncModifiedFilesProps = {
+  fileContentsUploader: ContentsUploader;
+  virtualDrive: VirtualDrive;
+};
+
+export async function syncModifiedFiles({ fileContentsUploader, virtualDrive }: SyncModifiedFilesProps) {
```
dajimenezriv: Each function should be in a different file

src/apps/sync-engine/in/sync-modified-file.ts
```diff
@@ -0,0 +1,73 @@
...
+    const localFile = files[remoteDriveFile.uuid as FileUuid];
+    if (!localFile) return;
+
+    const { nameWithExtension } = fileDecryptName({
+      plainName: remoteDriveFile.plainName,
+      encryptedName: remoteDriveFile.name,
+      parentId: remoteDriveFile.folderId,
+      extension: remoteDriveFile.type,
+    });
+    const relativePath = createRelativePath('/', nameWithExtension);
```
dajimenezriv: What is this? Are all files in the root?

src/apps/sync-engine/in/sync-modified-file.test.ts
```diff
@@ -0,0 +1,122 @@
...
+import { VirtualDrive } from '@/node-win/virtual-drive';
+import { mockDeep } from 'vitest-mock-extended';
+import { Stats } from 'fs';
+
+describe('sync-modified-file', () => {
+  const virtualDrive = mockDeep<VirtualDrive>();
+  const fileContentsUploader = mockDeep<ContentsUploader>();
+  const updateContentsIdMock = partialSpyOn(updateContentsIdModule, 'updateContentsId');
+
+  const createRemoteFile = (modificationDate: Date) => ({
```
dajimenezriv: You are using uuid and modificationTime, why not just pass those props?

src/apps/sync-engine/in/sync-modified-file.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    userUuid: 'user-uuid-1',
+    name: 'test.txt',
+    isDangledStatus: false,
+  });
+
+  beforeEach(() => {
+    updateContentsIdMock.mockResolvedValue(undefined);
+  });
+
+  it('should not update remote file if local modification time is older', async () => {
```
dajimenezriv: Use the function `mockProps`, we don't want to write all props if we are not going to use them.

src/apps/sync-engine/in/sync-modified-file.ts
```diff
@@ -0,0 +1,73 @@
...
+type SyncModifiedFilesProps = {
+  fileContentsUploader: ContentsUploader;
+  virtualDrive: VirtualDrive;
+};
+
+export async function syncModifiedFiles({ fileContentsUploader, virtualDrive }: SyncModifiedFilesProps) {
+  const remoteDriveFiles = await getExistingFiles();
+  const { files } = await loadInMemoryPaths({ drive: virtualDrive });
+
+  remoteDriveFiles.forEach(async (remoteDriveFile) => {
```
dajimenezriv: The forEach doesn't return an array of promises than then is awaited. You have to use a map and then Promise.all

src/apps/sync-engine/in/sync-modified-file.ts
dajimenezriv: Not sure, but maybe you need this function `pathUtils.absoluteToRelative`

src/apps/sync-engine/BindingManager.ts
egalvis39: fileContentsUploader does not exist in the controllers object, but it is required to use updateContentsId.

src/apps/sync-engine/binding-manager.ts
```diff
@@ -47,8 +47,9 @@ export class BindingsManager {
```
dajimenezriv: Why did you rename this file and RemoteItemsGenerator? Was it necessary for this PR?

src/apps/sync-engine/binding-manager.ts
```diff
@@ -47,8 +47,9 @@ export class BindingsManager {
 
     this.container.virtualDrive.connectSyncRoot({ callbacks });
 
-    void addPendingItems({ controllers: this.controllers });
-
+    const tree = await this.container.traverser.run();
+    void addPendingItems({ controllers: this.controllers, fileContentsUploader: this.container.contentsUploader, tree });
+    await this.load(tree);
```
dajimenezriv: Why the load function?

src/apps/sync-engine/in/sync-modified-files.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ContentsUploader } from '@/context/virtual-drive/contents/application/ContentsUploader';
+import { FileUuid } from '@/apps/main/database/entities/DriveFile';
+import { syncModifiedFile } from './sync-modified-file';
+import { VirtualDrive } from '@/node-win/virtual-drive';
+import { Tree } from '@/context/virtual-drive/items/application/Traverser';
+
+type SyncModifiedFilesProps = {
+  fileContentsUploader: ContentsUploader;
+  virtualDrive: VirtualDrive;
+  tree: Tree;
```
dajimenezriv: Why you compute the whole tree if you only need files that EXISTS? Weren't you using previously something like the dangled files?

src/apps/sync-engine/binding-manager.ts
egalvis27: I made some changes here but later removed them because they weren’t necessary. However, the file name change remained. Should I rename the file back?

src/apps/sync-engine/binding-manager.ts
egalvis27: This code isn’t mine — I just added line 51 between lines 50 and 52 to use the existing tree.

src/apps/sync-engine/in/sync-modified-files.ts
egalvis27: While retrieving the existing files from the database, I couldn’t find a way to build the file paths to update them remotely. I tried different approaches and searched through the code, but didn’t find a solution. For that reason, I used the tree that is already initialized in the binding-manager.

src/apps/sync-engine/binding-manager.ts
dajimenezriv: Yes please, I have changes in other branches and they won't merge easily because the file would be renamed.

src/apps/sync-engine/binding-manager.ts
dajimenezriv: In the previous changes in these lines the tree and load was not here, they were just inside the `polling` function.

src/apps/sync-engine/in/sync-modified-files.ts
dajimenezriv: The `load-in-memory-paths.ts` doesn't contain the absolute path?

src/backend/features/remote-sync/sync-items-by-checkpoint/load-in-memory-paths.test.ts
```diff
@@ -33,9 +33,8 @@ describe('load-in-memory-paths', () => {
     getFileUuidMock.mockReturnValueOnce({});
     getFileUuidMock.mockReturnValueOnce({ data: 'fileUuid2' as FileUuid });
     // When
-    const { files, folders } = await loadInMemoryPaths();
+    const { files } = await loadInMemoryPaths();
     // Then
-    expect(files).toStrictEqual({ fileUuid2: 'C:\\Users\\user\\InternxtDrive\\folder\\file2' });
-    expect(folders).toStrictEqual({ folderUuid: 'C:\\Users\\user\\InternxtDrive\\folder' });
```
dajimenezriv, line 38: Why did you delete the folders check?

src/backend/features/remote-sync/sync-items-by-checkpoint/load-in-memory-paths.test.ts
```diff
@@ -33,9 +33,8 @@ describe('load-in-memory-paths', () => {
     getFileUuidMock.mockReturnValueOnce({});
     getFileUuidMock.mockReturnValueOnce({ data: 'fileUuid2' as FileUuid });
     // When
-    const { files, folders } = await loadInMemoryPaths();
+    const { files } = await loadInMemoryPaths();
     // Then
-    expect(files).toStrictEqual({ fileUuid2: 'C:\\Users\\user\\InternxtDrive\\folder\\file2' });
```
dajimenezriv, line 37: And now the check is less precise, you can do:

```
fileUuid2: expect.objectContaining({ path: "the path" })
```

src/apps/main/remote-sync/handlers.ts
```diff
@@ -74,6 +79,10 @@ export async function getUpdatedRemoteItems(workspaceId: string) {
   }
 }
 
+void ipcMainSyncEngine.handle('FIND_EXISTING_FILES', () => {
+  return getExistingFiles();
```
dajimenezriv: You can do the previous previous function inside here, also, now we have the function:
```
SqliteModule.FilesModule.getByWorkspaceId()
```
otherwise you are retrieving all files for each workspace instead of filtering by workspace

src/apps/sync-engine/in/sync-modified-file.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Stats } from 'fs';
+
+type Props = {
+  remoteFile: DriveFile;
+  localFile: { path: string; stats: Stats };
+  fileContentsUploader: ContentsUploader;
+  virtualDrive: VirtualDrive;
+};
+
+export async function syncModifiedFile({ remoteFile, localFile, fileContentsUploader, virtualDrive }: Props) {
```
dajimenezriv, line 17: Btw, I think we talked about this, not sure, but we need to use the size instead, I've tried it again and it starts to upload many things because since we were not setting the modificationTime before, now it's inconsistent.

src/apps/sync-engine/in/sync-modified-file.ts
```diff
@@ -0,0 +1,46 @@
...
+import { AbsolutePath } from '@/context/local/localFile/infrastructure/AbsolutePath';
+import { pathUtils } from '@/context/local/localFile/infrastructure/AbsolutePath';
+import { DriveFile } from '@/apps/main/database/entities/DriveFile';
+import { VirtualDrive } from '@/node-win/virtual-drive';
+import { logger } from '@/apps/shared/logger/logger';
+import { Stats } from 'fs';
+
+type Props = {
+  remoteFile: DriveFile;
+  localFile: { path: string; stats: Stats };
```
dajimenezriv: Here, probably you have to set path as `AbsolutePath`, basically we are using branded types to enforce types, but it performs like an string. So there are no downsides and it lets us differenciate between AbsolutePath and RelativePath.

src/apps/sync-engine/in/sync-modified-file.ts
```diff
@@ -0,0 +1,46 @@
...
+      uuid: remoteFile.uuid,
+      remoteDate: remoteFile.modificationTime,
+      localDate: localFile.stats.mtime.toISOString(),
+    });
+
+    await updateContentsId({
+      virtualDrive,
+      stats: localFile.stats,
+      path: fileRelativePath,
+      uuid: remoteFile.uuid as string,
```
dajimenezriv: Same here, don't do this cast.

src/apps/sync-engine/in/sync-modified-files.test.ts
```diff
@@ -0,0 +1,233 @@
...
+  const createLocalFile = (path = 'C:\\Users\\test\\Drive\\test.txt') => ({
+    path: path as AbsolutePath,
+    stats: {
+      mtime: new Date('2025-08-05T10:00:01.000Z'),
+      size: 1000,
+    } as Stats,
+  });
+
+  beforeEach(() => {
+    syncModifiedFileMock.mockResolvedValue(undefined);
```
dajimenezriv, line 38: This is already done by default, you don't need to explicit resolve undefined.

src/apps/sync-engine/in/sync-modified-files.ts
```diff
@@ -0,0 +1,24 @@
+import { loadInMemoryPaths } from '@/backend/features/remote-sync/sync-items-by-checkpoint/load-in-memory-paths';
+import { ContentsUploader } from '@/context/virtual-drive/contents/application/ContentsUploader';
+import { FileUuid } from '@/apps/main/database/entities/DriveFile';
+import { syncModifiedFile } from './sync-modified-file';
+import { VirtualDrive } from '@/node-win/virtual-drive';
+import { getExistingFiles } from '@/context/virtual-drive/items/application/remote-items-generator';
+
+type SyncModifiedFilesProps = {
```
dajimenezriv: Just `type Props`, they are not exported, we don't need to name this things.

src/apps/sync-engine/in/sync-modified-file.ts
egalvis27: Now we’re currently using both checks to validate whether the file has real changes.
image

Should I switch to using only the file size instead?

src/apps/main/remote-sync/handlers.ts
```diff
@@ -74,6 +74,11 @@ export async function getUpdatedRemoteItems(workspaceId: string) {
   }
 }
 
+void ipcMainSyncEngine.handle('FIND_EXISTING_FILES', async (_, workspaceId: string) => {
+  const existingFiles = await SqliteModule.FileModule.getByWorkspaceId({ workspaceId });
+  return existingFiles.data;
```
dajimenezriv: This can be undefined, maybe you have to add ?? [].

src/apps/sync-engine/in/sync-modified-file.ts
dajimenezriv: Yes, use only file, modificationTime is not reliable for now

src/context/virtual-drive/items/application/Traverser.test.ts
```diff
@@ -2,13 +2,13 @@ import { v4 } from 'uuid';
 import { Traverser } from './Traverser';
 import * as crypt from '@/context/shared/infrastructure/crypt';
 import { deepMocked } from 'tests/vitest/utils.helper.test';
-import { getAllItems } from './RemoteItemsGenerator';
+import { getAllItems } from './remote-items-generator';
```
dajimenezriv: Rename this one too to it's original name, since it's not necessary for this PR

src/backend/features/remote-sync/sync-items-by-checkpoint/load-in-memory-paths.test.ts
```diff
@@ -35,7 +35,9 @@ describe('load-in-memory-paths', () => {
     // When
     const { files, folders } = await loadInMemoryPaths();
     // Then
-    expect(files).toStrictEqual({ fileUuid2: 'C:\\Users\\user\\InternxtDrive\\folder\\file2' });
     expect(folders).toStrictEqual({ folderUuid: 'C:\\Users\\user\\InternxtDrive\\folder' });
+    expect(files).toStrictEqual(
+      expect.objectContaining({ fileUuid2: expect.objectContaining({ path: 'C:\\Users\\user\\InternxtDrive\\folder\\file2' }) }),
```
dajimenezriv: The first expect.objectContaining is not necessary

src/apps/shared/HttpClient/schema.d.ts
```diff
@@ -2896,9 +2896,9 @@ export interface components {
       /**
        * Format: date-time
        * @description The last modification time of the file (optional)
-       * @example 2023-05-30T12:34:56.789Z
+       * @example 2025-08-05T12:34:56.000Z
        */
-      modificationTime?: string;
+      modificationTime?: Date;
```
dajimenezriv: The schema is autogenerated, you cannot change it manually, probabily to have to do mtime.toISOString()

src/apps/sync-engine/in/sync-modified-file.ts
```diff
@@ -0,0 +1,37 @@
...
+type Props = {
+  remoteFile: DriveFile;
+  localFile: { path: AbsolutePath; stats: Stats };
+  fileContentsUploader: ContentsUploader;
+  virtualDrive: VirtualDrive;
+};
+
+export async function syncModifiedFile({ remoteFile, localFile, fileContentsUploader, virtualDrive }: Props) {
+  if (localFile.stats.size !== remoteFile.size) {
+    const fileRelativePath = pathUtils.absoluteToRelative({ base: virtualDrive.syncRootPath, path: localFile.path });
```
dajimenezriv: By default we use `path` for `RelativePath` and `absolutePath` for `AbsolutePath`, to keep it consistent among the whole codebase.

src/apps/sync-engine/in/sync-modified-file.ts
```diff
@@ -0,0 +1,37 @@
...
+export async function syncModifiedFile({ remoteFile, localFile, fileContentsUploader, virtualDrive }: Props) {
+  if (localFile.stats.size !== remoteFile.size) {
+    const filePath = pathUtils.absoluteToRelative({ base: virtualDrive.syncRootPath, path: localFile.path });
+    logger.debug({
+      tag: 'SYNC-ENGINE',
+      msg: 'File placeholder has been modified locally, updating remote',
+      path: filePath,
+      uuid: remoteFile.uuid,
+      remoteDate: remoteFile.modificationTime,
+      localDate: localFile.stats.mtime.toISOString(),
```
dajimenezriv: Here, we should log the sizes not the datetimes

src/apps/sync-engine/in/sync-modified-file.test.ts
```diff
@@ -0,0 +1,102 @@
...
+      name: 'test.txt',
+      path: createRelativePath('/test.txt'),
+      size,
+    });
+
+  beforeEach(() => {
+    pathUtilsMock.mockReturnValue(createRelativePath('/test.txt'));
+  });
+
+  it('should not update remote file if local and remote file sizes are equal', async () => {
```
dajimenezriv, line 29: What's the difference between this test and the last one?

src/apps/sync-engine/in/sync-modified-files.test.ts
```diff
@@ -0,0 +1,235 @@
...
+        [file1Uuid]: localFile1,
+        [file2Uuid]: localFile2,
+      },
+      folders: {},
+    });
+    // When
+    await syncModifiedFiles({ fileContentsUploader, virtualDrive });
+    // Then
+    expect(syncModifiedFileMock).toBeCalledTimes(2);
+    expect(syncModifiedFileMock).toHaveBeenNthCalledWith(1, {
```
dajimenezriv, line 67: I didn't know about this method, but looks pretty nice

src/apps/sync-engine/in/sync-modified-files.test.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    expect(syncModifiedFileMock).toHaveBeenNthCalledWith(2, {
+      remoteFile: remoteFile2,
+      localFile: localFile2,
+      fileContentsUploader,
+      virtualDrive,
+    });
+  });
+
+  it('should not sync files that do not exist locally', async () => {
```
dajimenezriv: Isn't this test already included in the previous one?

dajimenezriv: And in the 'should handle mixed scenario with some files existing locally and others not'?

src/apps/sync-engine/in/sync-modified-files.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    });
+
+    syncModifiedFileMock.mockRejectedValueOnce(new Error('Sync failed for file1')).mockResolvedValueOnce(undefined);
+    // When
+    await expect(syncModifiedFiles({ fileContentsUploader, virtualDrive })).rejects.toThrow();
+    // Then
+    expect(syncModifiedFileMock).toBeCalledTimes(2);
+  });
+
+  it('should wait for all sync operations to complete', async () => {
```
dajimenezriv: What are we testing here? Is there any case in which this can fail because we changed something in the code?

src/apps/sync-engine/in/sync-modified-files.test.ts
egalvis27: It’s really useful when you need to check, for example, that a loop is iterating through the elements of a list and that the function inside the loop is being called with different information each time.

src/apps/sync-engine/in/sync-modified-files.test.ts
egalvis27: This test verifies that syncModifiedFiles waits for all sync operations to complete before resolving. It uses mocks with different delays (10ms and 5ms) to simulate concurrent async operations and confirms both finish successfully, ensuring the function implements "Promise.all" behavior for multiple files.

dajimenezriv: Okey, but in which case this test is useful? Because I only see the possibility of forget the Promise.all, then you are going to have an eslint error because you are not using the variable `promises`. It's better more tests than none, but take into account that every code written needs to be mantained and I see a test file with 215 lines to test a function with 10 lines that doesn't have any if statement. So, what is what this test is safe guarding?

dajimenezriv: Maybe here is better allSettled? Or do a try catch? Because nothing is dependant from this code but it will throw an error in the function that is calling this function

dajimenezriv-internxt (approved):

PR #946, opened 2025-08-05
Add get parent uuid

dajimenezriv-internxt:
## What

1. Add get by uuid endpoint so we can obtain the current state of a item in remote.
2. Add get by parent uuid in sqlite so we can retrieve all items from a folder that EXISTS.
3. Create a parse dto that is going to parse files and folders from remote to improve its format.

### DIFF
--- a/src/infra/drive-server-wip/out/dto.ts
+++ b/src/infra/drive-server-wip/out/dto.ts
@@ -4,10 +4,10 @@ import { components } from '@/apps/shared/HttpClient/schema';
 
 export type FileDto = components['schemas']['FileDto'];
 export type FolderDto = components['schemas']['FolderDto'];
-type ParsedFileDto = FileDto & {
+export type ParsedFileDto = FileDto & {
   uuid: FileUuid;
 };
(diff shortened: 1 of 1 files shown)


src/backend/features/remote-sync/sync-items-by-folder/update-items/is-item-newer.ts
```diff
@@ -0,0 +1,14 @@
+import { SimpleDriveFile } from '@/apps/main/database/entities/DriveFile';
+import { SimpleDriveFolder } from '@/apps/main/database/entities/DriveFolder';
+import { ParsedFileDto, ParsedFolderDto } from '@/infra/drive-server-wip/out/dto';
+
+type Props = {
+  item: SimpleDriveFile | SimpleDriveFolder;
+  itemDto: ParsedFileDto | ParsedFolderDto;
+};
+
+export function isItemNewer({ item, itemDto }: Props) {
```
dajimenezriv, line 10: It just checks if the local item is newer than the remote or not

PR #947, opened 2025-08-05
Remove local items that missed checkpoint

dajimenezriv-internxt:
## What

The context is new background sync by folder. Basically we fetch all items that EXISTS inside a folder. We have three possibilities:
- Exists locally and remotely.
- Exists locally.
- Exists remotely.
Currently we were only handling the last one, if it existed remotely we were creating or updating the item in sqlite.
Now we have improve this.

The file responsible of checking which possibility are we handling is `update-in-items.ts`.

- For the first possibility now we check if the local updateAt is newer (>) than the remote updateAt. If it's newer, we don't update the local sqlite. This is handled in `update-in-both.ts`.
- For the second possibility, we retrieve the item from remote to see what is the current state of that item in remote. Why do we have it in local and not in remote? Was it moved? Did we miss a checkpoint? If the local sqlite updateAt is newer than the remote, we don't do anything, maybe we this some action in local and that's why the local updatedAt is newer, otherwise we update the local sqlite with the remote. This is handled in `update-only-in-items.ts`.
- The last possibility means that the item is not in local so maybe we can create the placeholder. Handled in `update-only-in-dtos.ts`.


src/backend/features/remote-sync/sync-items-by-folder/update-folder-statuses.ts
```diff
@@ -1,30 +1,35 @@
...
+import { FolderUuid } from '@/apps/main/database/entities/DriveFolder';
 
 type TProps = {
   context: SyncContext;
   folderUuid: string;
 };
 
 export async function updateFolderStatuses({ context, folderUuid }: TProps) {
-  let folderUuids: string[] = [];
+  let folderUuids: FolderUuid[] = [];
```
AlexisMora, line 14: what is this variable used for?

src/backend/features/remote-sync/sync-items-by-folder/update-items/update-in-both.ts
```diff
@@ -0,0 +1,22 @@
...
+import { SyncContext } from '@/apps/sync-engine/config';
+import { createOrUpdateFolder } from '../../update-in-sqlite/create-or-update-folder';
+import { isItemNewer } from './is-item-newer';
+
+type Props = { context: SyncContext } & (
+  | { type: 'file'; itemDto: ParsedFileDto; item: SimpleDriveFile }
+  | { type: 'folder'; itemDto: ParsedFolderDto; item: SimpleDriveFolder }
+);
+
+export async function updateInBoth({ context, type, itemDto, item }: Props) {
```
AlexisMora, line 14: I dont know if this function is descriptive enough, but my first thought reading this function is that this function would perform 2 operations at the same time, but seems like is doing either one or another (or neither of both)
You could name it for example `updateItemIfOld` or `updateItemIfOutdated` but at least for me the current name is misleading

src/backend/features/remote-sync/sync-items-by-folder/update-items/update-items.ts
```diff
@@ -0,0 +1,46 @@
+import { SimpleDriveFile } from '@/apps/main/database/entities/DriveFile';
```
AlexisMora, line 1: To me, this file is doing too much and at a first glance is hard to understand what is doing.
In both promise1 and promise2 you are using a find inside a map, for large list this could be it slow since the complexity of each promise list  would be of O(n²)

src/backend/features/remote-sync/sync-items-by-folder/update-items/update-only-in-dtos.ts
```diff
@@ -0,0 +1,16 @@
...
+type Props = { context: SyncContext } & ({ type: 'file'; itemDto: ParsedFileDto } | { type: 'folder'; itemDto: ParsedFolderDto });
+
+export async function updateOnlyInDtos({ context, type, itemDto }: Props) {
+  if (type === 'file') {
+    await createOrUpdateFile({ context, fileDto: itemDto });
+  } else {
+    await createOrUpdateFolder({ context, folderDto: itemDto });
+  }
+
+  // Create placeholder
```
AlexisMora, line 15: I feel like this comment will get forgotten somehow

src/backend/features/remote-sync/sync-items-by-folder/update-folder-statuses.ts
dajimenezriv: We return it and it's used to recursively iterate through the folders. We have the folderUuid and then we return the folderUuids inside that folder

AlexisMora: ah yes i see the return when expanding the whole file

src/backend/features/remote-sync/sync-items-by-folder/update-items/update-only-in-dtos.ts
dajimenezriv: Maybe, but it is one of the priorities for big drives hahaha, we can delete it

AlexisMora (approved):

PR #952, opened 2025-08-06
Add eslint plugin sonar

dajimenezriv-internxt:
## What

Add eslint plugin sonar and fix some sonar errors.


.eslintrc.js
```diff
@@ -1,12 +1,29 @@
...
+        project: './tsconfig.json',
+      },
+    },
+    {
+      files: ['*.test.*'],
+      rules: {
+        'sonarjs/assertions-in-tests': 'off',
+        'sonarjs/constructor-for-side-effects': 'off',
+        'sonarjs/function-return-type': 'off',
+        'sonarjs/no-empty-test-file': 'off',
```
AlexisMora, line 25: Is this rule necessary? Having an empty test file doesnt seem right so, we should remove that file instead right?

.eslintrc.js
```diff
@@ -30,6 +47,24 @@ module.exports = {
     'no-use-before-define': ['warn', { functions: false }],
     'object-shorthand': 'error',
     'require-await': 'warn',
+    'sonarjs/array-callback-without-return': 'off',
+    'sonarjs/cognitive-complexity': 'off',
```
AlexisMora: [This rule ](https://sonarsource.github.io/rspec/#/rspec/S3776/javascript) as well seems right to me, why not at least have it as a warn?

.eslintrc.js
```diff
@@ -30,6 +47,24 @@ module.exports = {
...
+    'sonarjs/different-types-comparison': 'warn',
+    'sonarjs/no-empty-collection': 'off',
+    'sonarjs/no-identical-functions': 'off',
+    'sonarjs/no-ignored-exceptions': 'off',
```
AlexisMora: [This one as well](https://sonarsource.github.io/rspec/#/rspec/S2486/javascript), at least have a warning

.eslintrc.js
dajimenezriv: helper.test.ts is considered a test file but doesn't include any test.

AlexisMora (approved):

PR #953, opened 2025-08-06
Improve add pending items

dajimenezriv-internxt:
## What

We have improve the add pending items (items that were never synced). Basically now we run this on the start of the BindingsManager.ts, and we call directly `createFile` and `createFolder` instead of `onAdd` and `onAddDir`. `onAdd` and `onAddDir` are the callbacks for the watcher events, but they perform other validations that are not necessary here.

For some reason I had to remove the caret in ^vitest because tests started to fail (we can go back to commits from previous days and tests were failing). In the version 3.0.8 they work.


src/apps/sync-engine/BindingManager.ts
```diff
@@ -140,21 +138,6 @@ export class BindingsManager {
     });
 
     try {
-      const { pendingFiles, pendingFolders } = await getPlaceholdersWithPendingState({
```
dajimenezriv, line 143: This has been extracted into the function `add-pending-items.ts`

src/apps/sync-engine/in/add-pending-files.ts
```diff
@@ -1,20 +1,22 @@
...
 type TProps = {
-  watcher: Watcher;
+  controllers: IControllers;
   pendingFiles: PendingPaths[];
 };
 
-export async function addPendingFiles({ watcher, pendingFiles }: TProps) {
+export async function addPendingFiles({ controllers, pendingFiles }: TProps) {
   await Promise.all(
-    pendingFiles.map(async (pendingPath) => {
```
dajimenezriv, line 12: Here, instead of calling `onAdd` inside the watcher that checks if the uuid exists, if the file is moved and so, now we call directly to `createFile`, because we already know the the file doesn't have a placeholders and is new.

src/apps/sync-engine/in/add-pending-folders.ts
```diff
@@ -1,20 +1,22 @@
...
-  watcher: Watcher;
+  controllers: IControllers;
   pendingFolders: PendingPaths[];
 };
 
-export async function addPendingFolders({ watcher, pendingFolders }: TProps) {
+export async function addPendingFolders({ controllers, pendingFolders }: TProps) {
   await Promise.all(
-    pendingFolders.map(async (pendingPath) => {
-      await onAddDir({
```
dajimenezriv, line 13: Here, instead of calling `onAddDir` inside the watcher that checks if the uuid exists, if the file is moved and so, now we call directly to `createFolder`, because we already know the the file doesn't have a placeholders and is new.

src/apps/sync-engine/in/add-pending-items.ts
```diff
@@ -0,0 +1,36 @@
+import { logger } from '@internxt/drive-desktop-core/build/backend';
```
dajimenezriv, line 1: This is the extracted function from `BindingManager.ts`

AlexisMora (approved):

PR #955, opened 2025-08-07
Extract refresh item placeholders

dajimenezriv-internxt:
## What

This PR is just to facilitate the next one. Basically here we extract the process of building the tree and updating and deleting all placeholders into a separate function.

### DIFF
--- a/src/apps/sync-engine/BindingManager.ts
+++ b/src/apps/sync-engine/BindingManager.ts
@@ -11,9 +11,8 @@ import { Callbacks } from '@/node-win/types/callbacks.type';
 import { INTERNXT_VERSION } from '@/core/utils/utils';
 import { updateContentsId } from './callbacks-controllers/controllers/update-contents-id';
 import { createWatcher } from './create-watcher';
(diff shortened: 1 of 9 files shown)


src/apps/sync-engine/BindingManager.ts
```diff
@@ -54,16 +53,14 @@ export class BindingsManager {
 
     void addPendingItems({ controllers: this.controllers });
 
-    const tree = await this.container.traverser.run();
-    await this.load(tree);
```
dajimenezriv, line 58: The load was used only in `polling` function, so I moved it there. It loads the InMemoryFileRepository, which is still used only for dangled files. The `polling` function is the one that handled dangled files so move the `load` function inside there.

src/apps/sync-engine/BindingManager.ts
```diff
@@ -98,35 +95,6 @@ export class BindingsManager {
     logger.debug({ msg: 'In memory repositories loaded', workspaceId: getConfig().workspaceId });
   }
 
-  async update(tree: Tree) {
```
dajimenezriv, line 101: This function has been extracted in `refresh-item-placeholders.ts` but now we create also the tree inside instead of passing it as a prop.

PR #956, opened 2025-08-07
Refresh placeholders every minute

dajimenezriv-internxt:
## What

Now refresh placeholders is called every minute and when the sync finishes.

### DIFF
--- a/src/backend/features/remote-sync/sync-items-by-folder/sync-items-by-folder.test.ts
+++ b/src/backend/features/remote-sync/sync-items-by-folder/sync-items-by-folder.test.ts
@@ -76,8 +76,8 @@ describe('sync-items-by-folder', () => {
     expect(updateFolderStatusesMock).toBeCalledTimes(2);
     expect(updateFileStatusesMock).toBeCalledTimes(2);
     expect(sleepMock).toBeCalledTimes(2);
-    expect(updateFolderStatusesMock).toHaveBeenCalledWith({ context: props.context, folderUuid: 'rootFolderUuid' });
-    expect(updateFolderStatusesMock).toHaveBeenCalledWith({ context: props.context, folderUuid: 'folderUuid' });
+    expect(updateFolderStatusesMock).toHaveBeenCalledWith({ context: props.context, folderUuid: 'rootFolderUuid', path: '/' });
+    expect(updateFolderStatusesMock).toHaveBeenCalledWith({ context: props.context, folderUuid: 'folderUuid', path: '/' });
     expect(updateFileStatusesMock).toHaveBeenCalledWith({ context: props.context, folderUuid: 'rootFolderUuid' });
     expect(updateFileStatusesMock).toHaveBeenCalledWith({ context: props.context, folderUuid: 'folderUuid' });
   });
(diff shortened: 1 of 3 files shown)


src/apps/sync-engine/BindingManager.ts
```diff
@@ -60,7 +60,11 @@ export class BindingsManager {
      * This one is for the first case, since maybe the sync engine failed in a previous fetching
      * and we have some placeholders pending from being created/updated/deleted
      */
-    await refreshItemPlaceholders({ container: this.container });
+    await trackRefreshItemPlaceholders({ container: this.container });
+    setInterval(async () => {
+      logger.debug({ tag: 'SYNC-ENGINE', msg: 'Scheduled refreshing item placeholders', workspaceId: getConfig().workspaceId });
+      await trackRefreshItemPlaceholders({ container: this.container });
+    }, 60 * 1000);
```
dajimenezriv, line 67: Every minute we refresh placeholders. This is just called in the beginning of the sync process.

src/apps/sync-engine/index.ts
```diff
@@ -33,11 +35,15 @@ async function setUp() {
 
   const bindings = new BindingsManager(container);
 
-  ipcRenderer.on('UPDATE_SYNC_ENGINE_PROCESS', async () => {
+  ipcRendererSyncEngine.on('REFRESH_ITEM_PLACEHOLDERS', async () => {
```
dajimenezriv, line 38: We prepare this to be called from notifications. However first we need to save to sqlite when a notification arrives.

src/apps/sync-engine/refresh-item-placeholders.ts
```diff
@@ -5,19 +5,17 @@ import { DependencyContainer } from './dependency-injection/DependencyContainer'
 
 type Props = {
   container: DependencyContainer;
+  workspaceId: string;
 };
 
-export async function refreshItemPlaceholders({ container }: Props) {
-  const startTime = performance.now();
```
dajimenezriv, line 11: We move this to another function called `track-refresh-item-placeholders.ts`.

src/apps/sync-engine/track-refresh-item-placeholders.ts
```diff
@@ -0,0 +1,45 @@
+import { logger } from '@internxt/drive-desktop-core/build/backend';
```
dajimenezriv, line 1: Basically, if this function is not running, then run, and if it's running then enqueue so when the function finish it gets called again. I've tested with around 50k files and it takes 30s more or less. We have three big operations, take from sqlite, decrypt names and perform operations against file explorer. The first one will be reduced when we move to lokijs, the second one is not going to be necessary because we are going to decrypt as we fetch the items and the last one is going to run in a separate process.

src/backend/features/remote-sync/sync-items-by-folder/sync-items-by-folder.ts
```diff
@@ -3,36 +3,41 @@ import { updateFolderStatuses } from './update-folder-statuses';
...
 
 type TProps = {
   context: SyncContext;
-  rootFolderUuid: string;
+  rootFolderUuid: FolderUuid;
 };
 
 export async function syncItemsByFolder({ context, rootFolderUuid }: TProps) {
-  const folderUuids = [rootFolderUuid];
+  const folders = [{ folderUuid: rootFolderUuid, path: createRelativePath('/') }];
```
dajimenezriv, line 15: We construct the path for logging purposes, because the uuid is not easy to understand.

PR #958, opened 2025-08-08
Simplify create thumbnail

dajimenezriv-internxt:
## What

Move create thumbnail to the main process and simplify some things:
- Now we just hardcode the extensions available for image thumbnail.
- We call directly with absolutePath.
- We stop using the OfflineFile class for the persistance of a file.

I've tested uploading a `.exe` and a `.png` and it still works. For the `.png` a thumbnail is uploaded.


src/infra/sqlite/ipc/ipc-main.ts
```diff
@@ -23,7 +24,17 @@ export function setupIpcSqlite() {
   });
 
   void ipcMainSqlite.handle('fileCreateOrUpdate', async (_, props) => {
```
dajimenezriv, line 26: This is called just one time in the renderer process, so we call here the createAndUploadThumnail directly from the main process.

src/context/virtual-drive/files/application/FileCreator.ts
```diff
@@ -37,53 +35,40 @@ export class FileCreator {
...
-      logger.error({
+      throw logger.error({
         tag: 'SYNC-ENGINE',
         msg: 'Error in file creator',
         filePath: filePath.value,
-        exc: error,
+        exc,
       });
-
-      ipcRendererSyncEngine.send('FILE_UPLOAD_ERROR', {
```
dajimenezriv: We move this just above the log, so we can throw the logger.error

AlexisMora (approved):

AlexisMora (approved):

PR #962, opened 2025-08-14
feat (WIP): verify folder existence to throw FolderNotFoundError

cg-shokworks:


src/context/virtual-drive/files/application/FileCreator.ts
```diff
@@ -10,6 +10,7 @@ import { NodeWin } from '@/infra/node-win/node-win.module';
 import VirtualDrive from '@/node-win/virtual-drive';
 import { ipcRendererSqlite } from '@/infra/sqlite/ipc/ipc-renderer';
 import { AbsolutePath } from '@/context/local/localFile/infrastructure/AbsolutePath';
+import * as path from 'path';
```
dajimenezriv: Why this kind of import and not import directly the function that you want to use like with other imports?

src/infra/drive-server-wip/services/files.service.ts
```diff
@@ -14,6 +15,15 @@ export const files = {
   createThumbnail,
 };
 
+export class CreateFileError extends DriveServerWipError {
```
dajimenezriv: Can you extract this into it's own file?

src/infra/drive-server-wip/services/files.service.ts
```diff
@@ -52,18 +62,26 @@ async function createFile(context: { path: string; body: TCreateFileBody }) {
...
       attributes: {
         method,
         endpoint,
       },
     },
   });
+
+  if (result.error?.code === 'UNKNOWN') {
+    switch (true) {
+      case result.error.response?.status === 404:
```
dajimenezriv: Can we use here the message text instead of the code, since it's not very clear and a 404 in a create file maybe is retrieve because other reasons (I don't know, maybe the workspace doesn't exist¿?)

src/context/virtual-drive/files/application/FileCreator.ts
```diff
@@ -73,4 +96,78 @@ export class FileCreator {
       throw error;
     }
   }
+
+  private getFolderUuid({ posixPath }: { posixPath: string }): string | undefined {
```
dajimenezriv: Is this function really necessary? You are just calling the function

src/context/virtual-drive/files/application/FileCreator.ts
```diff
@@ -73,4 +96,78 @@ export class FileCreator {
...
+    const { data: folderUuid } = NodeWin.getFolderUuid({
+      drive: this.virtualDrive,
+      path: posixPath,
+    });
+    return folderUuid;
+  }
+
+  private async persistAndIndex({ remote, folderUuid, filePath, contents, absolutePath }: PersistAndIndexParams) {
```
dajimenezriv: All these functions should go into a different file, each file is just for a function.

cg-shokworks: One question here @dajimenezriv-internxt : in this case I intended to do this to reuse that functionality, but it belongs to the same class , it's only used within the FileCreator class, do you think that we should also separate it in different files? Because is not like something reusable and depends on the FileCreator constructor (I could pass the used params if necessary)

src/context/virtual-drive/files/application/FileCreator.ts
cg-shokworks: well actually I think I can remove it

src/context/virtual-drive/files/infrastructure/HttpRemoteFileSystem.ts
```diff
@@ -120,4 +128,39 @@ export class HttpRemoteFileSystem {
 
     return persistedFile;
   }
+
+  async existParentFolder(offline: TProps) {
```
dajimenezriv: I think these 3 methods are not needed. You can do for the last 2:
```
const [{ error: error1 }, { error: error2 }] = await Promise.all(moveRequest, renameRequest)
```
Btw, in which moment are you setting the folder again to EXISTS?

src/context/virtual-drive/files/application/FileCreator.ts
dajimenezriv: Btw, this is too much duplicated logic, why?

src/features/sync/add-item/update-folder.ts
```diff
@@ -0,0 +1,38 @@
+import { logger } from '@/apps/shared/logger/logger';
```
dajimenezriv: Wtf? Why you duplicate a whole file?

cg-shokworks: Sorry @dajimenezriv-internxt  I commited that file that was something i thought do to in the past but didnt delete. Mistakenly added it.

Just removed

src/context/virtual-drive/files/application/FileCreator.ts
cg-shokworks: changed

src/infra/drive-server-wip/services/files.service.ts
cg-shokworks: done

src/infra/drive-server-wip/services/files.service.ts
cg-shokworks: done

src/context/virtual-drive/files/application/FileCreator.ts
cg-shokworks: done

src/context/virtual-drive/files/infrastructure/HttpRemoteFileSystem.ts
cg-shokworks: done

src/context/virtual-drive/files/application/FileCreator.ts
cg-shokworks: done

src/infra/drive-server-wip/services/create-file.service.ts
```diff
@@ -0,0 +1,48 @@
...
+      attributes: {
+        method,
+        endpoint,
+      },
+    },
+  });
+
+  if (result.error?.code === 'UNKNOWN') {
+    switch (true) {
+      case result.error.response?.status === 404:
```
dajimenezriv: Why don't check the "Folder not found" here? In the response.message or similar

src/context/virtual-drive/files/application/persist-and-index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+type PersistAndIndexParams = {
+  remote: HttpRemoteFileSystem;
+  folderUuid: string;
+  filePath: FilePath;
+  contents: RemoteFileContents;
+  absolutePath: AbsolutePath;
+};
+
+export async function persistAndIndex({ remote, folderUuid, filePath, contents, absolutePath }: PersistAndIndexParams) {
```
dajimenezriv: This code is nearly the same as FileCreator, but now it needs to be mantained in two places.

src/context/virtual-drive/files/application/ensure-parent-folder-exists.ts
```diff
@@ -0,0 +1,46 @@
...
+  remote: HttpRemoteFileSystem;
+  virtualDrive: VirtualDrive;
+  posixDir: string;
+};
+
+export async function ensureParentFolderExists({ remote, virtualDrive, posixDir }: EnsureParentFolderParams) {
+  const targetFolderName = path.posix.basename(posixDir);
+  const grandParentFolder = PlatformPathConverter.getFatherPathPosix(posixDir);
+
+  const { data: folderUuid } = NodeWin.getFolderUuid({
```
dajimenezriv: Don't we have already the parentUuid from the file? Otherwise how we did the request before?

cg-shokworks: In the try, not in the catch

src/context/virtual-drive/files/application/FileCreator.ts
```diff
@@ -35,29 +35,43 @@ export class FileCreator {
...
+        contents,
         absolutePath,
       });
 
-      if (error) throw error;
-
       return fileDto;
     } catch (error) {
+      const nestedMsg = getNestedCauseMessage(error);
+      if (nestedMsg === 'Folder not found') {
```
dajimenezriv: What I meant here about not instanceof was not just instanceof, but not to throw an error and then capture it, because it's not type safe and as the time goes no we don't if something can throw an exception or not. Also, the no instanceof is  because of this: https://dannyguo.medium.com/how-to-fix-instanceof-not-working-for-custom-errors-in-typescript-1df978100a27. throws are also very dangerous. Why not apply this solution directly in HttpRemoteFileSystem instead of throwing an error in case is "Not folder found"?

src/infra/drive-server-wip/services/create-file.service.ts
cg-shokworks: I did a research on this and the actual message is in response.error.cause.cause.message

So I'm veryfing in the file creator directly 
image

image


Error is of type unknown btw... so i need to verify that it exists

src/context/virtual-drive/files/application/persist-and-index.ts
cg-shokworks: I don't understand this part. It's not the same as in the filecreator, this is called from 2 parts: 

line 38 (try) and line 63 (catch). I'm just encapsulating that logic to avoid repeating it in file creator

src/context/virtual-drive/files/application/FileCreator.ts
cg-shokworks: I mean I could, but should we apply business logic in the HttpRemoteFileSystem? I think that the file creator should  be in charge of the business logic and the http remote file system only interacts with the endpoints, being agnostic from the business logic

dajimenezriv: `HttpRemoteFileSystem` is not infra, the only infra here is `driveServerWip.files.createFile` and the same for workspace, you can apply business logic here. I've merged a PR to main so that `persist` now is only used in `sync`, so you can apply the error logic there.

src/context/virtual-drive/files/infrastructure/HttpRemoteFileSystem.ts
```diff
@@ -64,9 +71,11 @@ export class HttpRemoteFileSystem {
 
       if (existingFile) return existingFile;
 
-      throw logger.error({
+      logger.error({
```
dajimenezriv: Now you can revert this

src/context/virtual-drive/files/infrastructure/restore-parent-folder.ts
```diff
@@ -0,0 +1,63 @@
...
+  const targetFolderName = path.posix.basename(posixDir);
+  const grandParentFolder = PlatformPathConverter.getFatherPathPosix(posixDir);
+
+  const { data: parentUuid } = NodeWin.getFolderUuid({
+    drive: offline.drive,
+    path: grandParentFolder,
+  });
+
+  if (!parentUuid) {
+    throw logger.error({ msg: 'Local parentUuid not found' });
```
dajimenezriv: Just putting this messages doesn't give any kind of info in the logs. The parentUuid of what? If you can, add always the path and be descriptive. For example, "Could not restore parent folder, parentUuid not found"

src/context/virtual-drive/files/infrastructure/restore-parent-folder.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  const { data: existsData } = await driveServerWip.folders.existsFolder({
```
dajimenezriv: Here, instead of duplicating this, you can make the function existsFolder in `HttpRemoteFolderSystem` static and try to access it.

src/infra/drive-server-wip/services/files/create-file.test.ts
```diff
@@ -0,0 +1,31 @@
+// tests/create-file.test.ts
```
dajimenezriv: Why this comment?

src/infra/drive-server-wip/services/files/create-file.test.ts
```diff
@@ -0,0 +1,31 @@
...
+import { clientMock } from 'tests/vitest/mocks.helper.test';
+import { createFile } from './create-file';
+import { mockProps } from '@/tests/vitest/utils.helper.test';
+
+describe('create-file', () => {
+  const props = mockProps({});
+  it('should return data when file is created', async () => {
```
dajimenezriv: What is this test testing? Is there any use case in which this isn't true? Can we change something in the function that breaks this?

src/context/virtual-drive/files/infrastructure/HttpRemoteFileSystem.ts
```diff
@@ -44,7 +46,12 @@ export class HttpRemoteFileSystem {
         workspaceId: this.workspaceId,
       });
 
-      if (error) throw error;
+      if (error) {
+        if (error.code === 'FOLDER_NOT_FOUND') {
+          return await restoreParentFolder({ offline, drive: this.virtualDrive, bucket: this.bucket, workspaceId: this.workspaceId });
```
dajimenezriv: Btw, restoreParentFolder is also calling inside this `create` function, when technically is outside of its scope.
Isn't it better to:
```ts
await restoreParentFolder();
return this.create(offline);
```

src/context/virtual-drive/files/infrastructure/restore-parent-folder.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  const fatherSpy = partialSpyOn(PlatformPathConverter, 'getFatherPathPosix');
+  const getFolderUuidSpy = partialSpyOn(NodeWin, 'getFolderUuid');
+  const existsSpy = partialSpyOn(driveServerWip.folders, 'existsFolder');
+  const getConfigSpy = partialSpyOn(getConfig, 'getConfig');
+  const moveSpy = partialSpyOn(driveServerWip.folders, 'moveFolder');
+  const renameSpy = partialSpyOn(driveServerWip.folders, 'renameFolder');
+  const createSpy = partialSpyOn(HttpRemoteFileSystem, 'create');
+  const loggerSpy = partialSpyOn(logger, 'error');
+
+  beforeEach(() => {
```
dajimenezriv: You don't need to do this, we already reseting them in `vitest.config.mts` with `clearMocks: true`

src/context/virtual-drive/files/infrastructure/restore-parent-folder.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (!parentUuid) {
+    throw logger.error({ msg: 'Could not restore parent folder, parentUuid not found', grandParentFolder });
+  }
+
+  const existentFolder = await HttpRemoteFolderSystem.existFolder({
+    parentUuid,
+    plainName: targetFolderName,
+    path: grandParentFolder,
+  });
+  const remoteParentFolder = existentFolder || null;
```
dajimenezriv: Why this? Isn't existingFolder always defined?

src/context/virtual-drive/files/infrastructure/restore-parent-folder.ts
```diff
@@ -0,0 +1,66 @@
...
+  const targetFolderName = path.posix.basename(posixDir);
+  const grandParentFolder = PlatformPathConverter.getFatherPathPosix(posixDir);
+
+  const { data: parentUuid } = NodeWin.getFolderUuid({
+    drive,
+    path: grandParentFolder,
+  });
+
+  if (!parentUuid) {
+    throw logger.error({ msg: 'Could not restore parent folder, parentUuid not found', grandParentFolder });
```
dajimenezriv: It's better if you add the file path, because we can always calculate the grand parent path looking at the path, but not the opposite.

src/context/virtual-drive/files/infrastructure/restore-parent-folder.ts
```diff
@@ -0,0 +1,66 @@
...
+      }),
+      driveServerWip.folders.renameFolder({
+        name: targetFolderName,
+        workspaceToken: config.workspaceToken,
+        uuid: offline.folderUuid,
+      }),
+    ]);
+
+    if (moveError || renameError) {
+      throw logger.error({ msg: 'Error restoring parent folder', moveError, renameError });
```
dajimenezriv: Add here the path too

src/context/virtual-drive/files/infrastructure/restore-parent-folder.ts
```diff
@@ -0,0 +1,54 @@
...
+  const targetFolderName = path.posix.basename(posixDir);
+  const grandParentFolder = pathUtils.dirname(posixDir);
+
+  const { data: parentUuid } = NodeWin.getFolderUuid({
+    drive,
+    path: grandParentFolder,
+  });
+
+  if (!parentUuid) {
+    throw logger.error({ msg: 'Could not restore parent folder, parentUuid not found', posixDir });
```
dajimenezriv: Better offline.path than posixDir

src/context/virtual-drive/files/infrastructure/restore-parent-folder.ts
```diff
@@ -0,0 +1,54 @@
...
+      }),
+      driveServerWip.folders.renameFolder({
+        name: targetFolderName,
+        workspaceToken: config.workspaceToken,
+        uuid: offline.folderUuid,
+      }),
+    ]);
+
+    if (moveError || renameError) {
+      throw logger.error({ msg: 'Error restoring parent folder', moveError, renameError, posixDir });
```
dajimenezriv: Same here

dajimenezriv: And the path always after the `msg` to keep the same structure in all logs.

src/context/virtual-drive/files/infrastructure/restore-parent-folder.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  const dirnameSpy = partialSpyOn(pathUtils, 'dirname');
+  const getFolderUuidSpy = partialSpyOn(NodeWin, 'getFolderUuid');
+  const existsFolderSpy = partialSpyOn(driveServerWip.folders, 'existsFolder');
+  const getConfigSpy = partialSpyOn(getConfig, 'getConfig');
+  const moveSpy = partialSpyOn(driveServerWip.folders, 'moveFolder');
+  const renameSpy = partialSpyOn(driveServerWip.folders, 'renameFolder');
+  const loggerSpy = partialSpyOn(logger, 'error');
+
+  it('does nothing when remote parent exists', async () => {
+    dirnameSpy.mockImplementationOnce(() => '/gp/child' as RelativePath).mockImplementationOnce(() => '/gp' as RelativePath);
```
dajimenezriv: Here it's better to use `mockResolvedValueOnce` two times.

src/context/virtual-drive/files/infrastructure/restore-parent-folder.test.ts
```diff
@@ -0,0 +1,97 @@
...
+import { FolderUuid } from '../../folders/domain/FolderPlaceholderId';
+
+describe('restoreParentFolder', () => {
+  const dirnameSpy = partialSpyOn(pathUtils, 'dirname');
+  const getFolderUuidSpy = partialSpyOn(NodeWin, 'getFolderUuid');
+  const existsFolderSpy = partialSpyOn(driveServerWip.folders, 'existsFolder');
+  const getConfigSpy = partialSpyOn(getConfig, 'getConfig');
+  const moveSpy = partialSpyOn(driveServerWip.folders, 'moveFolder');
+  const renameSpy = partialSpyOn(driveServerWip.folders, 'renameFolder');
+  const loggerSpy = partialSpyOn(logger, 'error');
```
dajimenezriv: You can import `loggerMock` since it's very used we have already mocked it.

src/context/virtual-drive/files/infrastructure/restore-parent-folder.test.ts
```diff
@@ -0,0 +1,97 @@
+import { describe, it, expect } from 'vitest';
```
dajimenezriv, line 1: If something is the same among tests don't duplicate it, just write it before all tests, like `props` or the `dirnameSpy.mock`

src/context/virtual-drive/files/infrastructure/restore-parent-folder.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    renameSpy.mockResolvedValue({ error: undefined });
+
+    const props = mockProps<typeof restoreParentFolder>({
+      offline: { path: '/gp/child/file.txt' as RelativePath, folderUuid: 'offline-folder-uuid' },
+      drive: {},
+    });
+
+    await restoreParentFolder(props);
+
+    expect(moveSpy).toHaveBeenCalledTimes(1);
```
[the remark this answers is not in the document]
dajimenezriv-internxt: Or if you want to use `getMockCalls` then ignore the first line and just do:
`expect(getMockCalls(moveSpy)).toMatchObject`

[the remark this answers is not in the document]
dajimenezriv-internxt: Or why do you do in this way?

src/context/virtual-drive/files/infrastructure/restore-parent-folder.ts
```diff
@@ -0,0 +1,54 @@
...
+  const { data: parentUuid } = NodeWin.getFolderUuid({
+    drive,
+    path: grandParentFolder,
+  });
+
+  if (!parentUuid) {
+    throw logger.error({ msg: 'Could not restore parent folder, parentUuid not found', posixDir });
+  }
+
+  const remoteParentFolder = await HttpRemoteFolderSystem.existFolder({
```
dajimenezriv: Btw, why this check? Why don't do just the call to rename and move and if the grand parent folder doesn't exist we just throw an error?

PR #969, opened 2025-08-19
Do not execute add pending items in first registration

dajimenezriv-internxt:


src/apps/main/background-processes/sync-engine/services/unregister-virtual-drives.test.ts
```diff
@@ -7,7 +7,7 @@ vi.mock(import('@/apps/shared/logger/logger'));
 
 describe('unregister-virtual-drives', () => {
   const getRegisteredSyncRootsMock = deepMocked(VirtualDrive.getRegisteredSyncRoots);
-  const unRegisterSyncRootByProviderIdMock = deepMocked(VirtualDrive.unRegisterSyncRootByProviderId);
```
dajimenezriv, line 10: Rename `unRegisterSyncRootByProviderId` to `unregisterSyncRoot`, since the old `unregisterSyncRoot` inside `VirtualDrive` was no longer used.

src/apps/main/remote-sync/handlers.ts
```diff
@@ -170,8 +171,11 @@ ipcMain.handle('GET_UNSYNC_FILE_IN_SYNC_ENGINE', (_, workspaceId = '') => {
 
 export async function initSyncEngine({ context }: { context: AuthContext }) {
   try {
-    const { providerId } = await spawnDefaultSyncEngineWorker({ context });
-    await spawnWorkspaceSyncEngineWorkers({ context, providerId });
+    const syncRoots = VirtualDrive.getRegisteredSyncRoots();
```
dajimenezriv, line 174: Before spawning workspaces and drive, get which drives are registered.

src/apps/sync-engine/index.ts
```diff
@@ -77,11 +77,16 @@ async function refreshToken() {
 ipcRenderer.once('SET_CONFIG', (event, config: Config) => {
   setConfig(config);
 
+  const ctx: SyncContext = {
```
dajimenezriv, line 80: We are going to start using the context everywhere so we can get rid of getConfig() that only works inside the process.

AlexisMora (approved):

PR #970, opened 2025-08-20
Fix move error

dajimenezriv-internxt:
## What

### DIFF
--- a/src/apps/main/main.ts
+++ b/src/apps/main/main.ts
@@ -118,7 +118,7 @@ app
     await migrate();
 
     registerUsageHandlers();
-    await setUpBackups();
+    void setUpBackups();
 
     await checkIfUserIsLoggedIn();
     const isLoggedIn = getIsLoggedIn();
--- a/src/apps/sync-engine/BindingManager.ts
+++ b/src/apps/sync-engine/BindingManager.ts
@@ -57,10 +57,6 @@ export class BindingsManager {
      * and we have some placeholders pending from being created/updated/deleted
      */
     await trackRefreshItemPlaceholders({ container: this.container });
-    setInterval(async () => {
-      logger.debug({ tag: 'SYNC-ENGINE', msg: 'Scheduled refreshing item placeholders', workspaceId: getConfig().workspaceId });
-      await trackRefreshItemPlaceholders({ container: this.container });
-    }, 60 * 1000);
   }
 
   watch() {
@@ -70,7 +66,6 @@ export class BindingsManager {
         addController: this.controllers.addFile,
         updateContentsId: async ({ stats, path, uuid }) =>
           await updateContentsId({
-            virtualDrive: this.container.virtualDrive,
             stats,
             path,
             uuid,
--- a/src/apps/sync-engine/callbacks-controllers/controllers/add-controller.ts
+++ b/src/apps/sync-engine/callbacks-controllers/controllers/add-controller.ts
@@ -51,13 +51,7 @@ export class AddController {
     }
   }
 
-  async createFolder({ path }: { path: RelativePath }) {
-    logger.debug({ msg: 'Create folder', path });
-
-    try {
-      await createFolder({ path, folderCreator: this.folderCreator });
-    } catch (error) {
-      logger.error({ tag: 'SYNC-ENGINE', msg: 'Error in folder creation', path, error });
-    }
+  async createFolder({ path, absolutePath }: { path: RelativePath; absolutePath: AbsolutePath }) {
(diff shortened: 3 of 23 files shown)


src/node-win/watcher/watcher.ts
```diff
@@ -42,8 +42,14 @@ export class Watcher {
     try {
       this.chokidar = watch(this.syncRootPath, this.options);
       this.chokidar
+        /**
+         * v2.5.7 Daniel Jiménez
+         * add events are triggered when:
+         * - we create an item locally.
```
dajimenezriv, line 48: For some reason I though that when we were creating a placeholder it was also triggering this, but it's not.

src/apps/main/main.ts
```diff
@@ -118,7 +118,7 @@ app
     await migrate();
 
     registerUsageHandlers();
-    await setUpBackups();
+    void setUpBackups();
```
dajimenezriv, line 121: Otherwise it blocks starting the sync engine until the backup is completed

src/apps/sync-engine/BindingManager.ts
```diff
@@ -57,10 +57,6 @@ export class BindingsManager {
      * and we have some placeholders pending from being created/updated/deleted
      */
     await trackRefreshItemPlaceholders({ container: this.container });
-    setInterval(async () => {
-      logger.debug({ tag: 'SYNC-ENGINE', msg: 'Scheduled refreshing item placeholders', workspaceId: getConfig().workspaceId });
-      await trackRefreshItemPlaceholders({ container: this.container });
-    }, 60 * 1000);
```
dajimenezriv, line 63: We cannot do this, because if we move an item locally and the refresh item placeholders is executed with an old status it's going to duplicate the item.

src/apps/sync-engine/callbacks-controllers/controllers/add-controller.ts
```diff
@@ -51,13 +51,7 @@ export class AddController {
...
-  async createFolder({ path }: { path: RelativePath }) {
-    logger.debug({ msg: 'Create folder', path });
-
-    try {
-      await createFolder({ path, folderCreator: this.folderCreator });
-    } catch (error) {
-      logger.error({ tag: 'SYNC-ENGINE', msg: 'Error in folder creation', path, error });
-    }
+  async createFolder({ path, absolutePath }: { path: RelativePath; absolutePath: AbsolutePath }) {
+    await createFolder({ path, absolutePath, folderCreator: this.folderCreator });
```
dajimenezriv, line 55: This function already has a catch inside

src/apps/sync-engine/callbacks-controllers/controllers/update-contents-id.ts
```diff
@@ -34,7 +33,7 @@ export async function updateContentsId({ virtualDrive, stats, path, uuid, fileCo
       newSize: contents.size,
     });
 
-    virtualDrive.updateSyncStatus({ itemPath: path, isDirectory: false, sync: true });
+    updateFileStatus({ path });
```
dajimenezriv, line 36: It's the same but extracted because it's used in 3 places

src/apps/sync-engine/index.ts
```diff
@@ -35,10 +34,6 @@ async function setUp({ ctx }: { ctx: SyncContext }) {
 
   const bindings = new BindingsManager(container);
 
-  ipcRendererSyncEngine.on('REFRESH_ITEM_PLACEHOLDERS', async () => {
-    await trackRefreshItemPlaceholders({ container });
-  });
```
dajimenezriv, line 40: This was for websockets but not yet used. However, as said, we cannot refresh item placeholders without fetching all items from checkpoint, otherwise errors can happen.

src/backend/features/local-sync/placeholders/update-folder-status.ts
```diff
@@ -0,0 +1,16 @@
+import { virtualDrive } from '@/apps/sync-engine/dependency-injection/common/virtualDrive';
+import { AbsolutePath, RelativePath } from '@/context/local/localFile/infrastructure/AbsolutePath';
+import { readdir } from 'fs/promises';
+
+type Props = {
+  path: RelativePath;
+  absolutePath: AbsolutePath;
+};
+
+export async function updateFolderStatus({ path, absolutePath }: Props) {
```
dajimenezriv, line 10: For now, we can just update sync status of empty folders because otherwise it starts dehydrating all inner files and so.

AlexisMora (approved):

PR #971, opened 2025-08-20
Some refactor

dajimenezriv-internxt:
## What

Some refactors I have accumulated.


src/apps/main/antivirus/utils/get-files-from-directory.ts
```diff
@@ -32,7 +32,7 @@ export async function getFilesFromDirectory({ rootFolder }: TProps) {
 
         if (name.includes('temp') || name.includes('tmp')) continue;
 
-        const fullPath = resolve(folder, item.name);
```
dajimenezriv: This is clearer using join and does the same job.

src/apps/main/antivirus/windows-defender/get-defender-version.ts
```diff
@@ -1,19 +1,18 @@
-import { fileSystem } from '@/infra/file-system/file-system.module';
+import { logger } from '@internxt/drive-desktop-core/build/backend';
 import { access, constants, readdir } from 'fs/promises';
-import { join } from 'path';
 
 export async function getDefenderVersions({ path }: { path: string }) {
   try {
     await access(path, constants.F_OK);
-    const entries = await readdir(path);
+    const entries = await readdir(path, { withFileTypes: true });
```
dajimenezriv, line 7: We get the stats here so we don't have to do `fileSystem.stat` everytime.

src/apps/sync-engine/dependency-injection/common/virtualDrive.ts
```diff
@@ -1,7 +1,14 @@
 import VirtualDrive from '@/node-win/virtual-drive';
+import { getConfig } from '../../config';
 
 export let virtualDrive: VirtualDrive;
 
-export function initializeVirtualDrive(drive = new VirtualDrive()) {
+export function initializeVirtualDrive(
```
dajimenezriv, line 6: Since there is going to be a moment in which VirtualDrive will be in the main process (I've tested and more or less we can already do that) it's better to just access getConfig() here, later we will pass the ctx: SyncContext instead

src/apps/sync-engine/in/get-pending-items.ts
```diff
@@ -26,7 +26,7 @@ async function processFolder({ virtualDrive, path }: TProps) {
    * We cannot use `withFileTypes` because it treats everything as a symbolic link,
    * so we have to use `stat` for each entry.
    */
-  const entries = await readdir(path);
+  const entries = await readdir(path, { recursive: true });
```
dajimenezriv, line 29: More performant

src/apps/sync-engine/tests/create-placeholder.infra.test.ts
```diff
@@ -109,10 +111,11 @@ describe('create-placeholder', () => {
 
     await sleep(100);
     await writeFile(file, 'content');
-    await sleep(10000);
+    await sleep(5000);
 
     // Then
     const status = container.virtualDrive.getPlaceholderState({ path: file });
     expect(status.pinState).toBe(PinState.AlwaysLocal);
+    expect(getMockCalls(onAllMock)).toStrictEqual([{ event: 'add', path: file }]);
```
dajimenezriv, line 119: The `add` event is also triggered when we add the file, not when the placeholder is created.

src/backend/features/remote-sync/sync-items-by-checkpoint/load-in-memory-paths.ts
```diff
@@ -19,35 +19,26 @@ export async function loadInMemoryPaths() {
...
-        if (stats.isFile()) {
-          const { data: uuid } = NodeWin.getFileUuid({ drive: virtualDrive, path: absolutePath });
-          if (uuid) files[uuid] = absolutePath;
-        }
+  /**
+   * v2.5.6 Daniel Jiménez
+   * We cannot use `withFileTypes` because it treats everything as a symbolic link,
+   * so we have to use `stat` for each entry.
+   */
+  const entries = await readdir(rootPath, { recursive: true });
```
dajimenezriv, line 27: Same here, we add the recursive so it's more performant and clearer.

src/node-win/virtual-drive.ts
```diff
@@ -172,81 +174,69 @@ export class VirtualDrive {
   }
 
   createFileByPath({
```
dajimenezriv, line 176: I've added some test to check that the code is the same because here we were doing an strange loop which in the end was just `dirname`

src/node-win/virtual-drive.ts
```diff
@@ -172,81 +174,69 @@ export class VirtualDrive {
...
+        basePath: parentPath,
       });
-    } catch (error) {
-      logger.error({ msg: 'Error creating placeholder', error });
+    } catch (exc) {
+      logger.error({ msg: 'Error creating file placeholder', path, exc });
     }
   }
 
   createFolderByPath({
```
dajimenezriv, line 210: Same here

src/node-win/watcher/events/on-all.service.ts
```diff
@@ -0,0 +1,7 @@
+type Props = {
+  event: string;
+  path: string;
+};
+
+// eslint-disable-next-line no-empty-pattern
+export function onAll({}: Props) {}
```
dajimenezriv, line 7: This is just for testing, so we can mock this function and check how many events are triggered, since this function captures al watcher/chokidar events.

PR #972, opened 2025-08-21
Revert previous provider ids

dajimenezriv-internxt:
## What

This was a solution to not execute the add pending items when the app was first registered. Basically if we quit the app all placeholders were removed but technically those items were not pending, so the solution was to not execute add pending items in the first registration. However, the solution is easier, just execute add pending items after refreshing the items 🙃.


src/apps/sync-engine/BindingManager.ts
```diff
@@ -57,6 +55,14 @@ export class BindingsManager {
...
      */
     await trackRefreshItemPlaceholders({ container: this.container });
+
+    /**
+     * v2.5.7 Daniel Jiménez
+     * If the cloud provider was not registered before it means that all items that
+     * were in the root folder have their placeholders gone, so we need to refresh first
+     * all item placeholders and the execute this function.
+     */
+    void addPendingItems({ ctx, controllers: this.controllers });
```
dajimenezriv: We just need to execute this after the previous function xD

egalvis27 (approved):

PR #973, opened 2025-08-21
Remove file

dajimenezriv-internxt:
## What

This is the first PR that gets rid of the File class, why? Because it adds complexity and also cannot be sent through the ipc. We are going to move everything to `SimpleDriveFile` and `ExtendedDriveFile` (maybe we can rename then in the future) and the same for folders.

Also, this PR extracts the find-deadcode check in the pipeline to another pipeline that doesn't block the merge. Why? Because sometimes we delete many things but not all fill in the 500 size, but they are no longer used in the code and the find-deadcode fails. With that we get the warning but it doesn't block the PR.

### DIFF
--- a/.github/workflows/find-deadcode.yml
+++ b/.github/workflows/find-deadcode.yml
@@ -5,7 +5,7 @@ on:
     types: [opened, synchronize, reopened]
 
 jobs:
-  pull_request:
+  find_deadcode:
     runs-on: windows-latest
     timeout-minutes: 5
 


src/apps/backups/diff/calculate-files-diff.ts
```diff
@@ -64,10 +62,7 @@ export function calculateFilesDiff({ local, remote }: TProps) {
   });
 
   Object.values(remote.files).forEach((remoteFile) => {
-    // Already deleted
-    if (remoteFile.status.value !== FileStatuses.EXISTS) return;
```
dajimenezriv, line 68: Here, all remote files can only be EXISTS, because fetch files by folders only retrieves the EXISTS values.

src/apps/backups/fetch-items/fetch-files-by-folder.ts
```diff
@@ -34,7 +35,7 @@ export async function fetchFilesByFolder({ folderUuid, allFiles, abortSignal, of
     hasMore = data.length === FETCH_LIMIT;
     offset += FETCH_LIMIT;
 
-    const filteredData = data.filter((file) => file.status === 'EXISTS');
```
dajimenezriv, line 37: As said, here we can only have EXISTS files because of the fetch, we don't need to filter.

src/apps/backups/remote-tree/traverser.ts
```diff
@@ -38,45 +37,15 @@ export class Traverser {
...
     if (context.abortController.signal.aborted) return;
 
-    const filesInThisFolder = items.files.filter((file) => file.folderUuid === currentFolder.uuid);
+    const filesInThisFolder = items.files.filter((file) => file.parentUuid === currentFolder.uuid);
     const foldersInThisFolder = items.folders.filter((folder) => folder.parentUuid === currentFolder.uuid);
 
-    filesInThisFolder.forEach((serverFile) => {
-      let relativePath: RelativePath | undefined;
+    filesInThisFolder.forEach((file) => {
+      const path = createRelativePath(currentFolder.path, file.nameWithExtension);
```
dajimenezriv, line 44: Name is already decrypted here so it cannot throw any error.

src/context/virtual-drive/files/infrastructure/HttpRemoteFileSystem.ts
```diff
@@ -67,7 +66,7 @@ export class HttpRemoteFileSystem {
     }
 
     const data = response.data;
-    if (data.status !== FileStatuses.EXISTS) return null;
```
dajimenezriv, line 85: FileStatuses was used only here, also, status is typed so it can only be "EXISTS", "TRASHED" or "DELETED", we don't need to use the class FileStatuses

src/apps/backups/fetch-items/fetch-files-by-folder.test.ts
```diff
@@ -31,17 +34,6 @@ describe('fetch-files-by-folder', () => {
     expect(getFilesByFolderMock).toHaveBeenCalledTimes(0);
   });
 
-  it('Add only files with status EXISTS', async () => {
```
dajimenezriv, line 34: As said, here we can only reach with EXISTS values so this is not necessary

src/infra/drive-server-wip/out/dto.ts
```diff
@@ -18,6 +19,30 @@ export function parseFileDto({ fileDto }: { fileDto: FileDto }): ParsedFileDto {
   };
 }
 
+export function newParseFileDto({ fileDto }: { fileDto: FileDto }): SimpleDriveFile {
```
dajimenezriv, line 22: This is more or less a temporal function, this one and the one below will be merged so we only work with `SimpleDriveFile` and `ExtendedDrive` in the whole. app.

PR #975, opened 2025-08-22
Make file persist static

dajimenezriv-internxt:
## What

Currently, the persist operation is used for create a file in sync, backups and dangled files. For the sync, we need to take into account that maybe we have a wrong status of the parent folder in the local database or that the file already exists and we need to take different actions. However, for backups we retrieve before starting the backup the current status in remote, so it's always going to be up to date. (We are going to ignore dangled files).

So, @cg-shokworks is implementing a solution to improve the error handling in the create file in sync, so to be able to use the tyfe safe error we need to split the create file function so that the backup doesn't use the same.

### DIFF
--- a/src/context/local/localFile/application/upload/file-batch-uploader.test.ts
+++ b/src/context/local/localFile/application/upload/file-batch-uploader.test.ts
@@ -24,6 +24,7 @@ describe('file-batch-uploader', () => {
   beforeEach(() => {
     props = mockProps<typeof service.run>({
       self: { backed: 0 },
+      context: { backupsBucket: 'bucket' },
       tracker: { currentProcessed: vi.fn() },
       remoteTree: { folders: { [parentPath]: { uuid: 'parentUuid' } } },
       added: [{ relativePath: path, size: { value: 1024 } }],


src/context/virtual-drive/files/infrastructure/HttpRemoteFileSystem.ts
```diff
@@ -11,28 +11,43 @@ import { getNameAndExtension } from '../domain/get-name-and-extension';
...
+    };
+
+    return offline.workspaceId
+      ? driveServerWip.workspaces.createFileInWorkspace({ body, workspaceId: offline.workspaceId, path: offline.path })
+      : driveServerWip.files.createFile({ body, path: offline.path });
+  }
+
   async persist(offline: { contentsId: string; folderUuid: string; path: string; size: number }) {
     try {
-      const { name, extension } = getNameAndExtension({ nameWithExtension: basename(offline.path) });
```
dajimenezriv, line 19: We have extracted this first part into an static function `create` so that it can be used from backups. Now the `persist` function is only used in `sync` and dangled files, so we can implement inside that function the custom error handling.

AlexisMora (approved):

PR #976, opened 2025-08-22
Add backups infra test

dajimenezriv-internxt:
## What

1. Stop using dependency injection in backups.
2. Add an infra test to backups (for now it doesn't take into account errors).


src/apps/backups/diff/calculate-folders-diff.ts
```diff
@@ -31,9 +30,6 @@ export function calculateFoldersDiff({ local, remote }: TProps) {
   });
 
   Object.values(remote.folders).forEach((folder) => {
-    // Already deleted
-    if (folder.status !== FolderStatuses.EXISTS) return;
```
dajimenezriv, line 35: Fetch by folder only brings EXISTS items so this is not needed.

src/apps/backups/fetch-items/fetch-folders-by-folder.ts
```diff
@@ -35,9 +35,8 @@ export async function fetchFoldersByFolder({ folderUuid, allFolders, abortSignal
     hasMore = data.length === FETCH_LIMIT;
     offset += FETCH_LIMIT;
 
-    const filteredData = data.filter((folder) => folder.status === 'EXISTS');
```
dajimenezriv, line 38: Fetch by folder only fetches EXISTS items so this is not need.

src/apps/main/background-processes/backups/build-file-uploader.ts
```diff
@@ -0,0 +1,27 @@
...
+import { getConfig } from '@/apps/sync-engine/config';
+import { INTERNXT_CLIENT, INTERNXT_VERSION } from '@/core/utils/utils';
+import { EnvironmentFileUploader } from '@/infra/inxt-js/file-uploader/environment-file-uploader';
+import { Environment } from '@internxt/inxt-js';
+
+type Props = {
+  backupInfo: BackupInfo;
+};
+
+export function buildFileUploader({ backupInfo }: Props) {
```
dajimenezriv, line 11: This has been extract from the `builder` of the dependency injection.

src/apps/backups/diff/calculate-files-diff.ts
```diff
@@ -34,9 +34,6 @@ export function calculateFilesDiff({ local, remote }: TProps) {
       return;
     }
 
-    const remoteModificationTime = Math.trunc(new Date(remoteFile.updatedAt).getTime() / 1000);
-    const localModificationTime = Math.trunc(local.modificationTime.getTime() / 1000);
```
dajimenezriv, line 38: This is really not good since we are comparing uodatedAt and modificationAt, so use for now size, which is more stable.

src/apps/backups/BackupInfo.ts
```diff
@@ -11,5 +12,6 @@ export type BackupInfo = {
 
 export type BackupsContext = BackupInfo & {
   abortController: AbortController;
+  fileUploader: EnvironmentFileUploader;
```
dajimenezriv, line 15: Instead of doing dependency injection for the upload we are going to send it in the backups context so it's available everywhere. We will do the same for sync.

src/apps/backups/backups.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    await mkdir(unmodifiedFolder);
+    await mkdir(addedFolder);
+    await writeFile(unmodifiedFile, 'content');
+    await writeFile(modifiedFile, 'content');
+    await writeFile(addedFile, 'content');
+  });
+
+  it('should perform a complete backup', async () => {
+    // Given
+    // @ts-expect-error not sure why this error
```
dajimenezriv, line 59: I'm not sure, but this happened to me before also, and when I do `mockImplementation` with different return values the types doesn't work.

src/apps/backups/diff/calculate-folders-diff.test.ts
```diff
@@ -20,8 +19,6 @@ describe('calculate-folders-diff', () => {
         folders: {
           ['/folder1' as RelativePath]: FolderMother.fromPartial({ path: '/folder1' }),
           ['/folder3' as RelativePath]: FolderMother.fromPartial({ path: '/folder3' }),
-          ['/folder4' as RelativePath]: FolderMother.fromPartial({ path: '/folder4', status: FolderStatuses.DELETED }),
```
dajimenezriv, line 23: We cannot reach `DELETED` here, fetch by folder only brings `EXISTS`.

PR #977, opened 2025-08-25
Sync modified file always local

dajimenezriv-internxt:
## What

1. Check that the file is hydrated before trying to update the remote file, otherwise is going to download the file and the upload it, and that is not what we want, at least for now.
2. Save the replaced file to the sqlite as soon as possible.


src/apps/sync-engine/in/sync-modified-file.ts
```diff
@@ -16,19 +17,31 @@ type Props = {
...
-      msg: 'File placeholder has been modified locally, updating remote',
-      path: filePath,
-      uuid: remoteFile.uuid,
+      msg: 'File has been modified locally, updating remote',
+      path,
       localSize: localFile.stats.size,
       remoteSize: remoteFile.size,
     });
 
+    if (pinState !== PinState.AlwaysLocal) {
```
dajimenezriv, line 31: We check if the file is hydrated (the content is available locally, not only in the cloud)

egalvis27 (dismissed):

egalvis27 (approved):

PR #979, opened 2025-08-25
Stop using Folder class

dajimenezriv-internxt:


src/apps/backups/fetch-items/fetch-folders-by-folder.ts
```diff
@@ -35,8 +36,9 @@ export async function fetchFoldersByFolder({ folderUuid, allFolders, abortSignal
     hasMore = data.length === FETCH_LIMIT;
     offset += FETCH_LIMIT;
 
-    newFolders.push(...data);
-    allFolders.push(...data);
+    const parsedData = data.map((folderDto) => newParseFolderDto({ folderDto }));
```
dajimenezriv, line 39: We parse from dto to `SimpleDriveFolder`

src/apps/backups/folders/create-folders.ts
```diff
@@ -45,19 +45,20 @@ export async function createFolders({ self, context, added, tree, tracker }: TPr
...
           parentFolderUuid: parent.uuid,
         },
       });
 
-      if (data) {
-        tree.folders[localFolder.relativePath] = Folder.from({
-          ...data,
+      if (folderDto) {
+        tree.folders[localFolder.relativePath] = {
+          ...newParseFolderDto({ folderDto }),
```
dajimenezriv, line 58: Same here, parse from dto to `SimpleDriveFile`

src/apps/backups/remote-tree/traverser.ts
```diff
@@ -48,36 +45,12 @@ export class Traverser {
       tree.files[path] = extendedFile;
     });
 
-    foldersInThisFolder.forEach((serverFolder) => {
```
dajimenezriv, line 51: Now we don't need the try catch or the decrypt name, we have already done that in the `newParseFolderDto`

src/apps/main/remote-sync/handlers.ts
```diff
@@ -205,21 +205,22 @@ ipcMain.handle('CHECK_SYNC_IN_PROGRESS', (_, workspaceId = '') => {
 ipcMain.handle('get-item-by-folder-uuid', async (_, folderUuid): Promise<ItemBackup[]> => {
   logger.debug({ msg: 'Getting items by folder uuid', folderUuid });
 
-  const abortController = new AbortController();
-
-  const { folders } = await fetchItems({
+  const { data: folders = [] } = await driveServerWip.folders.getFoldersByFolder({
```
dajimenezriv, line 208: This is nearly the same, basically previously we were using `fetchItems` which internally was using this endpoint. This response is used when we are looking into the folders backed from another device in the frontend. Previously we were also iterating the tree inside which is not correct. Now, we just fetch the folders inside the device and return them.

src/context/virtual-drive/folders/domain/Folder.ts
```diff
@@ -83,11 +82,6 @@ export class Folder {
     );
   }
 
-  static decryptName({ plainName, name, parentId }: { plainName?: string | null; name: string; parentId?: number | null }) {
```
dajimenezriv, line 86: We have extracted this into it's own function, like with `file-decrypt-name.ts`

src/infra/drive-server-wip/out/dto.ts
```diff
@@ -49,3 +50,20 @@ export function parseFolderDto({ folderDto }: { folderDto: FolderDto }): ParsedF
     uuid: folderDto.uuid as FolderUuid,
   };
 }
+
+export function newParseFolderDto({ folderDto }: { folderDto: FolderDto }): SimpleDriveFolder {
```
dajimenezriv, line 54: As with `newParseFileDto` this is more or less temporal, in the end we only want to work with `SimpleDriveFolder` and `ExtendedDriveFolder`.

AlexisMora (approved):

PR #983, opened 2025-08-26
Migrate to context

dajimenezriv-internxt:
## What

Start sending `ctx` using prop drilling to we can stop using getConfig and migrate the sync engine process to main.

### DIFF
--- a/src/apps/sync-engine/config.ts
+++ b/src/apps/sync-engine/config.ts
@@ -1,6 +1,7 @@
 import { AuthContext } from '@/backend/features/auth/utils/context';
 import { getUser } from '../main/auth/service';
 import { FolderUuid } from '../main/database/entities/DriveFolder';
+import { VirtualDrive } from '@/node-win/virtual-drive';
 
 export type Config = {
   userUuid: string;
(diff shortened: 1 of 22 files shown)


src/apps/sync-engine/config.ts
```diff
@@ -20,6 +21,10 @@ export type Config = {
 
 export type SyncContext = AuthContext & Config;
 
+export type ProcessSyncContext = SyncContext & {
+  virtualDrive: VirtualDrive;
+};
```
dajimenezriv, line 26: In the end we will just have one `SyncContext`, however, for now we need separate contexts because the one from `main` cannot access the node-win in C++ yet.

src/apps/sync-engine/index.ts
```diff
@@ -28,8 +28,6 @@ async function setUp({ ctx }: { ctx: SyncContext }) {
 
   await ensureTheFolderExist(rootPath);
 
-  initializeVirtualDrive();
```
dajimenezriv, line 31: We need to move this function to be executed asap, so line 73, just after `setConfig`

AlexisMora (approved):

PR #986, opened 2025-08-26
Make traverser static

dajimenezriv-internxt:


src/context/virtual-drive/contents/application/ContentsUploader.ts
```diff
@@ -1,27 +1,19 @@
...
 };
 
 export class ContentsUploader {
-  constructor(
-    private readonly remoteContentsManagersFactory: EnvironmentRemoteFileContentsManagersFactory,
-    private readonly relativePathToAbsoluteConverter: RelativePathToAbsoluteConverter,
-  ) {}
-
-  async run({ path, stats }: Props) {
-    const win32RelativePath = PlatformPathConverter.posixToWin(path);
```
dajimenezriv, line 21: This was used just to calculate the absolute path for dangled files, so we have moved it inside `FileContentsHardUpdater.ts`

PR #989, opened 2025-08-27
Improve logs

dajimenezriv-internxt:
## What

Send just the logs folder to `drive-desktop-core`.


package.json
```diff
@@ -10,7 +10,7 @@
     "build:main": "cross-env NODE_ENV=production TS_NODE_TRANSPILE_ONLY=true webpack --config .erb/configs/webpack.config.main.prod.ts",
     "build:renderer": "cross-env NODE_ENV=production TS_NODE_TRANSPILE_ONLY=true webpack --config .erb/configs/webpack.config.renderer.prod.ts",
     "build:sync-engine": "cross-env NODE_ENV=production TS_NODE_TRANSPILE_ONLY=true webpack --config .erb/configs/webpack.config.sync-engine.ts",
-    "build:preload": "npx esbuild src/apps/main/preload.ts --outfile=src/apps/main/preload.js --bundle --platform=node --format=cjs --external:electron --external:electron-log --external:path",
+    "build:preload": "npx esbuild src/apps/main/preload.ts --outfile=src/apps/main/preload.js --bundle --platform=node --format=cjs --external:electron --external:electron-log --external:path --external:@internxt",
```
dajimenezriv, line 13: Basically the `preload.js` is built on fly, and it tries to write all code in commonjs instead of importing it. We need to add to the external @internxt, so all @internxt packages are just imported and not added directly.

AlexisMora (approved):

PR #990, opened 2025-08-27
Remove changelog and contributing

dajimenezriv-internxt:
## What

1. Move CONTRIBUTING.md to `drive-desktop-core`.
2. Remove CHANGELOG.md since it's already pretty log and most webpack infra things have been already modified.
3. Move to MIT license. Why? I asked Sergio when we created `drive-desktop-core` and he told that we can use MIT license, so to use the same everywhere, move to MIT. Also, in the project `sdk` that was created by Sergio they also use MIT (https://github.com/internxt/sdk).


.github/workflows/check-pr-size.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
           # Calculate added and deleted lines between the merge base and the head commit
           TOTAL_CHANGED=$(git diff --numstat "$MERGE_BASE" "$HEAD_SHA" \
-            | grep -Ev "(package-lock.json|schema.d.ts|preload.js)" \
+            | grep -Ev "(package-lock.json|schema.d.ts|preload.js|CHANGELOG.md|LICENSE)" \
```
dajimenezriv, line 36: This is just temporal, it will be removed in next PR

PR #991, opened 2025-08-27
Fix restore parent folder

dajimenezriv-internxt:
## What

Fix two issues when restoring parent folder:

1. If we have moved the folder and is has been marked as TRASHED for some reason. When we move back and rename it maybe the name was already right and it throws an error. This has happen to me when doing a test for the release.
2. When an error because of the rename, move or whatever, it enters in the catch and tries to do the getFileByPath which also fails because the problem was the folder or other thing. Now we just execute the getFileByPath if the create-file endpoint gives a 409 (conflict, file already exists).

### DIFF
--- a/src/infra/drive-server-wip/services/files/rename-folder.ts
+++ b/src/infra/drive-server-wip/services/files/rename-folder.ts
@@ -1,0 +1,53 @@
+import { client, getWorkspaceHeader } from '@/apps/shared/HttpClient/client';
+import { getRequestKey } from '../../in/get-in-flight-request';
+import { clientWrapper } from '../../in/client-wrapper.service';
+import { DriveServerWipError, TDriveServerWipError } from '../../out/error.types';
+import { parseFolderDto } from '../../out/dto';
+
+class RenameFolderError extends DriveServerWipError {
+  constructor(
+    public readonly code: TDriveServerWipError | 'FOLDER_ALREADY_EXISTS',
+    cause: unknown,
+  ) {
+    super(code, cause);
+  }
+}
+
+export async function renameFolder(context: { uuid: string; name: string; workspaceToken: string }) {
+  const method = 'PUT';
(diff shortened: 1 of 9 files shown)


src/infra/drive-server-wip/services/workspaces.service.ts
```diff
@@ -120,37 +119,6 @@ async function getFoldersInWorkspace(context: { workspaceId: string; query: Quer
   });
 }
 
-async function createFileInWorkspace(context: { workspaceId: string; path: string; body: CreateFileInWorkspaceBody }) {
```
dajimenezriv, line 123: We have extracted this into the file `workspaces/create-file.ts`.

src/infra/drive-server-wip/services/workspaces/create-file.ts
```diff
@@ -0,0 +1,34 @@
...
+      msg: 'Create file in workspace request',
+      context,
+      attributes: {
+        method,
+        endpoint,
+      },
+    },
+  });
+
+  return parseCreateFileResponse(res);
```
dajimenezriv, line 33: Same as previous function but now we call this parse function.

src/infra/drive-server-wip/services/files/rename-folder.ts
```diff
@@ -0,0 +1,53 @@
...
+      attributes: {
+        method,
+        endpoint,
+      },
+    },
+  });
+
+  if (res.error?.code === 'UNKNOWN') {
+    switch (true) {
+      case res.error.response?.status === 409:
```
dajimenezriv, line 43: We add this typed error to the rename folder endpoint since maybe the name was already right.

dajimenezriv: There can be also the use case of what happens if the user created also a folder with that name in the previous location and so, but I don't think we will have that kind of errors and it's a very edge case.

PR #992, opened 2025-08-27
feat: sync remote changes to local

egalvis27:


src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.ts
```diff
@@ -0,0 +1,68 @@
+import { AbsolutePath } from '@/context/local/localFile/infrastructure/AbsolutePath';
```
dajimenezriv, line 1: Instead of loggin the `local` which is very big just log the relative path that you can obtain from `remote.path`. Also, logging the uuid doesn't help us at all, we cannot do anything with that so it's just noise, at least for now.

dajimenezriv: Btw, can you rename the path of local to absolutePath? Basically all relative paths are called `path` and all absolutePaths are called `absolutePath`, so we have the same structure.

src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    // When
+    await syncRemoteChangesToLocal(props);
+    // Then
+    expect(unlinkMock).not.toBeCalled();
+    expect(virtualDrive.hydrateFile).not.toBeCalled();
+  });
+
+  it('should handle errors gracefully', async () => {
+    // Given
+    virtualDrive.getPlaceholderState.mockReturnValue({ pinState: PinState.AlwaysLocal });
```
dajimenezriv: You can also put all mocks of AlwaysLocal to the beforeEach.

src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.ts
```diff
@@ -0,0 +1,68 @@
...
+
+      virtualDrive.createFileByPath({
+        itemPath: remote.path,
+        itemId: `FILE:${remote.uuid}`,
+        size: remote.size,
+        creationTime: new Date(remote.createdAt).getTime(),
+        lastWriteTime: new Date(remote.updatedAt).getTime(),
+      });
+
+      await virtualDrive.hydrateFile({ itemPath: remote.path });
```
dajimenezriv: Btw, we cannot do this directly, hydrateFile breaks if two are done at the same time, otherwise it breaks the whole process, we have a queue manager, so we need to enqueue the task. For now maybe it's easy to just skip the hydration and just create the placeholder and make it OnlineOnly.

src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.ts
```diff
@@ -0,0 +1,63 @@
...
+
+type Props = {
+  virtualDrive: VirtualDrive;
+  remote: ExtendedDriveFile;
+  local: { absolutePath: AbsolutePath; stats: Stats };
+};
+
+export async function syncRemoteChangesToLocal({ remote, local, virtualDrive }: Props) {
+  const remoteDate = new Date(remote.updatedAt);
+  const placeholderState = virtualDrive.getPlaceholderState({ path: local.absolutePath });
```
dajimenezriv: Why not { pinState } here?

src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.ts
```diff
@@ -0,0 +1,62 @@
...
+type Props = {
+  virtualDrive: VirtualDrive;
+  remote: ExtendedDriveFile;
+  local: { absolutePath: AbsolutePath; stats: Stats };
+};
+
+export async function syncRemoteChangesToLocal({ remote, local, virtualDrive }: Props) {
+  const remoteDate = new Date(remote.updatedAt);
+  const { pinState } = virtualDrive.getPlaceholderState({ path: local.absolutePath });
+  if (pinState === PinState.AlwaysLocal && remote.size !== local.stats.size && remoteDate > local.stats.mtime) {
```
dajimenezriv: We don't have to check the pinState, because if the placeholder is creating we need to change the size of it, so we need to recreate it.

PR #993, opened 2025-08-28
Fix is move event

dajimenezriv-internxt:


src/backend/features/local-sync/watcher/events/unlink/is-move-event.ts
```diff
@@ -1,34 +1,40 @@
...
-  addEvents: new Set<FileUuid>(),
-  addDirEvents: new Set<FolderUuid>(),
+  addFileEvents: new Map<FileUuid, NodeJS.Timeout>(),
+  addFolderEvents: new Map<FolderUuid, NodeJS.Timeout>(),
 };
 
-export async function trackAddEvent({ uuid }: { uuid: FileUuid }) {
-  store.addEvents.add(uuid);
-  await sleep(15_000);
-  store.addEvents.delete(uuid);
```
dajimenezriv, line 13: Previously we were deleting always the event after 15_000, so there was an issue when.
1. trackAddEvent() -> adds and event of addFile
2. isMoveEvent() -> checks that is a move event and removes the uuid from the set.
3. trackAddEvent() -> adds and event of addFile
4. the first trackAddEvent() was finishing and removing the uuid from the set.
5. isMoveEvent() -> the second move event was cheking in the set and not finding the uuid, so it was considered an unlink event.

src/backend/features/local-sync/watcher/events/unlink/is-move-event.ts
```diff
@@ -1,34 +1,40 @@
...
-}
+function trackAddEvent<T>(map: Map<T, NodeJS.Timeout>, uuid: T) {
```
dajimenezriv, line 10: Now, when we track an add event we add it to the store.addFileEvents (or Folder) and set a timeout in 5 seconds. If we track again, remove the previous timeout and create a new one, so no race condition can happen.

src/backend/features/local-sync/watcher/events/unlink/is-move-event.ts
```diff
@@ -1,34 +1,40 @@
...
+  timeoutId = setTimeout(() => map.delete(uuid), 5_000);
+  map.set(uuid, timeoutId);
 }
 
-export async function isMoveEvent({ uuid }: { uuid: FileUuid }) {
-  await sleep(5_000);
-  const isMoveEvent = store.addEvents.has(uuid);
-  if (isMoveEvent) store.addEvents.delete(uuid);
-  return isMoveEvent;
+async function isMoveEvent<T>(map: Map<T, NodeJS.Timeout>, uuid: T) {
```
dajimenezriv, line 18: Same here, if it's a move event remove the timeout and delete manually the uuid from the Map

PR #994, opened 2025-08-28
Create walk function

dajimenezriv-internxt:


src/apps/backups/Backups.ts
```diff
@@ -23,6 +23,8 @@ export class Backup {
     const local = await LocalTreeBuilder.run({ context });
     const remote = await new Traverser().run({ context });
 
+    if (context.abortController.signal.aborted) return;
```
dajimenezriv, line 26: This was missing, but if we aborted while fetching current backup state from cloud it was considering that some files were not there and it was trying to create them. We need to check the abort signal here too, before doing any backup process.

src/apps/main/antivirus/utils/get-files-from-directory.ts
```diff
@@ -17,37 +16,16 @@ export async function getFilesFromDirectory({ rootFolder }: TProps) {
...
-    const folder = folders.shift();
-    if (!folder) continue;
-
-    try {
-      const items = await readdir(folder, { withFileTypes: true });
-
-      for (const item of items) {
-        const name = item.name.toLowerCase();
-
-        if (name.includes('temp') || name.includes('tmp')) continue;
```
dajimenezriv, line 33: The only thing that we have removed now is that we don't skip tmp or temp folders. However, technically this doesn't make too much sense, since not all tmp folders are removed when we turn off the computer. Also, if we do a manual scan it will skip tmp folders and that doesn't make sense. I see the utility when we are scanning the whole user folder to skip some parts, however, the scan is extremely slow anyway and we should look for other ways, like not awaiting the whole readdir before start displaying anything to the user (which was happening before and now).

src/context/local/localTree/application/LocalTreeBuilder.ts
```diff
@@ -49,8 +49,6 @@ export default class LocalTreeBuilder {
       };
 
       tree.folders[relativePath] = folder;
-
-      await this.traverse({ context, tree, currentFolder: folder });
```
dajimenezriv, line 53: We don't need this, the readdir of backups is already doing the recursive part when obtaining the paths.

src/context/local/localTree/infrastructure/FsLocalItemsGenerator.test.ts
```diff
@@ -55,24 +43,29 @@ describe('CLSFsLocalItemsGenerator', () => {
...
-      folders: [{ path: folder1 }, { path: folder2 }],
+      folders: [{ path: folder1 }, { path: folder2 }, { path: folder3 }],
     });
   });
 
   it('If stat returns an error it should add an issue and continue', async () => {
     // Given
-    statMock.mockResolvedValue({ data: { mtime: new Date(), size: 7 } });
-    statMock.mockResolvedValueOnce({ error: { code: 'NON_EXISTS' } });
+    execSync(`icacls "${file3}" /deny "${process.env.USERNAME}":F`);
```
dajimenezriv, line 55: Now we are doing real access deny so we can be sure that the `syncWalk` is working as expected.

cg-shokworks (approved):

PR #995, opened 2025-08-29
Improve is move event

dajimenezriv-internxt:
## What

Explained in the code.

### DIFF
--- a/src/backend/features/local-sync/watcher/events/unlink/is-move-event.ts
+++ b/src/backend/features/local-sync/watcher/events/unlink/is-move-event.ts
@@ -11,19 +11,13 @@ function trackAddEvent<T>(map: Map<T, NodeJS.Timeout>, uuid: T) {
   let timeoutId = map.get(uuid);
   if (timeoutId) clearTimeout(timeoutId);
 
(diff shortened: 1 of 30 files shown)


src/backend/features/local-sync/watcher/events/unlink/is-move-event.ts
```diff
@@ -11,19 +11,13 @@ function trackAddEvent<T>(map: Map<T, NodeJS.Timeout>, uuid: T) {
...
 }
 
 async function isMoveEvent<T>(map: Map<T, NodeJS.Timeout>, uuid: T) {
-  await sleep(2_000);
-
-  const timeoutId = map.get(uuid);
-  if (!timeoutId) return false;
-
-  clearTimeout(timeoutId);
-  map.delete(uuid);
```
dajimenezriv, line 25: We don't want to delete the `add` event, because if there are two consequent moves in less than 2 seconds, the first one is going to delete the `add` event and the second `move` is going to do an unlink. Now, the problem is that since we don't delete the `add` event once we do the `move` if we unlink in less than 10s it won't be deleted. But this is scenario is better than the other.

egalvis27 (approved):

PR #998, opened 2025-08-29
feat: create read stream within limiter

cg-shokworks:


src/infra/inxt-js/file-uploader/environment-file-uploader.test.ts
```diff
@@ -2,17 +2,18 @@ import { mockDeep } from 'vitest-mock-extended';
...
 import * as processError from './process-error';
 import * as abortOnChangeSize from './abort-on-change-size';
+import * as fs from 'fs';
 
 vi.mock(import('fs'));
 
 describe('environment-file-uploader', () => {
   partialSpyOn(abortOnChangeSize, 'abortOnChangeSize');
   const processErrorMock = partialSpyOn(processError, 'processError');
+  const createReadStream = partialSpyOn(fs, 'createReadStream');
```
dajimenezriv: Here, you are not really mocking the createReadStream since `fs` cannot be spyed. This works because we were already doing above `vi.mock(import("fs"))`. So here you need to do:
```ts
const createReadStreamMock = vi.mocked(createReadStream);
```

src/infra/inxt-js/file-uploader/environment-file-uploader.test.ts
```diff
@@ -21,6 +22,9 @@ describe('environment-file-uploader', () => {
   let abortController: AbortController;
   let props: Parameters<typeof service.upload>[0];
 
+  const mockedReadable = mockProps({ destroy: vi.fn(), close: vi.fn() });
```
dajimenezriv: Here, to mock an object is easier to do `mockDeep<ReadStream>()`

src/infra/inxt-js/file-uploader/environment-file-uploader.test.ts
```diff
@@ -21,6 +22,9 @@ describe('environment-file-uploader', () => {
...
+
+  createReadStream.mockReturnValue(mockedReadable);
```
dajimenezriv: And this is better to be inside the `beforeEach`, because if for some reason we want to a `restoreMocks` globally like we do with `clearMocks: true` or similar, it's better to be mocked for every test, otherwise we will lose the value. Or let's say that we want to override the mock for just a test, after that test we want to use again the default mock.

src/infra/inxt-js/file-uploader/environment-file-uploader.test.ts
```diff
@@ -2,28 +2,31 @@ import { mockDeep } from 'vitest-mock-extended';
...
 
   const callbacks = mockDeep<FileUploaderCallbacks>();
   let abortController: AbortController;
   let props: Parameters<typeof service.upload>[0];
-
+  let mockedReadable: ReadStream;
   beforeEach(() => {
     abortController = new AbortController();
     props = mockProps<typeof service.upload>({ size: 100, callbacks, abortSignal: abortController.signal });
+    mockedReadable = mockDeep<ReadStream>();
```
dajimenezriv: This can be outside, see `mockDeep<Environment>();`

dajimenezriv-internxt (approved):

PR #1000, opened 2025-09-02
Fix refresh token error

dajimenezriv-internxt:
## What

Fix a bug that was causing the app to not start when the user had the refresh token expired.


src/apps/main/auth/handlers.ts
```diff
@@ -48,7 +48,14 @@ export async function checkIfUserIsLoggedIn() {
 
   await checkUserData();
   encryptToken();
-  await createTokenSchedule();
+
+  try {
+    await createTokenSchedule();
+  } catch (exc) {
+    if (exc instanceof RefreshError) return;
+    else throw exc;
```
dajimenezriv, line 56: We don't want to to throw an exception in case of a refresh token error, we just want to omit the `emitUserLoggedIn`.

PR #1003, opened 2025-09-02
Sync modified files infra test

dajimenezriv-internxt:


src/node-win/watcher/detect-context-menu-action.service.ts
```diff
@@ -21,27 +22,22 @@ export async function detectContextMenuAction({ ctx, self, details, absolutePath
...
 
   const { data: uuid } = NodeWin.getFileUuid({ drive: self.virtualDrive, path });
-  const status = self.virtualDrive.getPlaceholderState({ path });
+  const { pinState } = self.virtualDrive.getPlaceholderState({ path });
   const isInDevice = self.fileInDevice.has(absolutePath);
 
-  if (!uuid) return;
+  const diff = getStatsDiff({ prev, curr });
+
+  self.logger.debug({ msg: 'Change event triggered', path, pinState, diff });
```
dajimenezriv, line 30: We obtain the different values between the `prev` stats and the `curr` stats and we log them. We don't want to log all values since stats is a big object.

src/node-win/watcher/get-stats-diff.ts
```diff
@@ -0,0 +1,21 @@
...
+  prev: Stats;
+  curr: Stats;
+};
+
+export function getStatsDiff({ prev, curr }: Props) {
+  const diff: Record<string, unknown> = {};
+  const keys = Object.keys(prev) as Array<keyof Stats>;
+
+  for (const key of keys) {
+    if (prev[key] instanceof Date || curr[key] instanceof Date) continue;
```
dajimenezriv, line 13: To check stats we ignore Date fields, because we already have `atimeMs`, `mtimeMs`, `ctimeMs`, `birthtimeMs`

PR #1005, opened 2025-09-04
Remove bearer token

dajimenezriv-internxt:
## What

1. Remove `bearerToken` because it's no longer used (it was used for `drive-server`).

### DIFF
--- a/src/backend/features/auth/services/utils/are-credentials-already-reseted.test.ts
+++ b/src/backend/features/auth/services/utils/are-credentials-already-reseted.test.ts
@@ -20,7 +20,7 @@ describe('areCredentialsAlreadyReseted', () => {
 
   it('should return false if credentials are not reset', () => {
     mockConfigStore.mockImplementation((field) => {
-      if (field === 'bearerToken') {
+      if (field === 'newToken') {
         return 'some-token';
       }
       return defaults[field as keyof typeof defaults];


src/apps/main/main.ts
```diff
@@ -77,6 +76,7 @@ setupQuitHandlers();
 setupIssueHandlers();
 setupIpcDriveServerWip();
 setupIpcSqlite();
+registerUsageHandlers();
```
dajimenezriv, line 79: We can move this here to keep all handlers together, but probably they should be in a better place.

src/apps/main/auth/service.ts
```diff
@@ -2,43 +2,19 @@ import { parseAndDecryptUserKeys } from '../../../apps/shared/crypto/keys.servic
...
 type EncryptedTokenKey = `${(typeof tokensKeys)[number]}Encrypted`;
 
 type Credentials = {
   userData: User;
-  bearerToken: string;
   newToken: string;
   password: string;
 };
 
-export function encryptToken() {
```
dajimenezriv, line 20: This was just for the old `bearerToken`

src/apps/main/auth/refresh-token.ts
```diff
@@ -15,35 +15,21 @@ async function refreshToken() {
...
+  const shceduler = new TokenScheduler(5, tokens);
   const schedule = shceduler.schedule(refreshToken);
 
   if (!schedule && !refreshedTokens) {
     logger.debug({ msg: 'Refreshing tokens' });
     createTokenSchedule(await refreshToken());
   }
 }
-
-export async function checkUserData(): Promise<void> {
```
dajimenezriv, line 37: This was to migrate `root_folder_id` (number) to `rootFolderId` (uuid), but it was implemented 12 months ago, so we can delete it.

PR #1010, opened 2025-09-05
Use only new token

dajimenezriv-internxt:
## What

Since now we have removed `bearerToken` we can just use the default token `newToken` instead of working with an array of tokens.

### DIFF
--- a/src/apps/main/auth/headers.ts
+++ b/src/apps/main/auth/headers.ts
@@ -9,7 +9,7 @@ export const HEADERS = {
 };
 
 export function getAuthHeaders(): Record<string, string> {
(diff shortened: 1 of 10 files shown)


src/apps/main/auth/headers.ts
```diff
@@ -9,7 +9,7 @@ export const HEADERS = {
 };
 
 export function getAuthHeaders(): Record<string, string> {
-  const token = obtainToken('newToken');
```
dajimenezriv, line 12: Now we only have `newToken`

AlexisMora (approved):

PR #1013, opened 2025-09-05
Simplify builder

dajimenezriv-internxt:
## What

The objetive of this PRs is to simplify the dependency injection process. We join all initializations in one function and currently it's only used for download (it's still legacy implementation) and the dangled files (also legacy but for now not worth of refactoring, since it's just temporal code).


src/apps/sync-engine/build-process-container.ts
```diff
@@ -0,0 +1,52 @@
+import { RelativePathToAbsoluteConverter } from '@/context/virtual-drive/shared/application/RelativePathToAbsoluteConverter';
```
dajimenezriv, line 1: Now all dependency injection of the sync engine process is done in this file.

src/apps/sync-engine/BindingManager.ts
```diff
@@ -57,53 +47,19 @@ export class BindingsManager {
...
-    void this.polling({ ctx });
     void queueManager.processQueue();
   }
 
-  stop({ ctx }: { ctx: ProcessSyncContext }) {
+  static stop({ ctx }: { ctx: ProcessSyncContext }) {
     ctx.virtualDrive.disconnectSyncRoot();
   }
 
-  async load(ctx: ProcessSyncContext, tree: Tree): Promise<void> {
```
dajimenezriv, line 73: We have split this into a different file called `run-dangled-files.ts` to move the legacy code to a different place.

src/apps/sync-engine/BindingManager.ts
```diff
@@ -57,53 +47,19 @@ export class BindingsManager {
     void addPendingItems({ ctx });
   }
 
-  watch({ ctx }: { ctx: ProcessSyncContext }) {
+  static watch({ ctx }: { ctx: ProcessSyncContext }) {
     const { queueManager, watcher } = createWatcher({ ctx });
 
     watcher.watchAndWait({ ctx });
 
-    void this.polling({ ctx });
```
dajimenezriv, line 65: This is the only thing that changes. We have move it to here: https://github.com/internxt/drive-desktop/pull/1013/files#diff-ceaeb509204a3b3953dae6a32cf7aa2083d7a7b0095a6e9621b7b28430a01a06R56 and renamed it.

egalvis27 (approved):

PR #1016, opened 2025-09-08
Simplify delete item placeholders

dajimenezriv-internxt:
## What

1. Remove `track-refresh-item-placeholders` since it's no longer needed. Now we can just call `refresh-item-placeholders` directly.
2. Use native `rm` instead of our C++ implementation.


src/apps/sync-engine/BindingManager.ts
```diff
@@ -4,7 +4,7 @@ import { logger } from '../shared/logger/logger';
 import { Callbacks } from '@/node-win/types/callbacks.type';
 import { createWatcher } from './create-watcher';
 import { addPendingItems } from './in/add-pending-items';
-import { trackRefreshItemPlaceholders } from './track-refresh-item-placeholders';
```
dajimenezriv, line 7: This was a wrapper of `refreshing-item-placeholders`, why? Previously we were refreshing placeholders after a full sync was made (every 10 minutes or when a notifications arrives) and also in the recovery sync and we didn't want to run the `refreshItemPlaceholders` two times at the same time. However, the last execution has been removed because it had some edge cases that were not safe, so now we no longer need this. Also, because it will break once we move the default drive and the workspaces together.

src/apps/sync-engine/track-refresh-item-placeholders.ts
```diff
@@ -1,44 +0,0 @@
-import { logger } from '@internxt/drive-desktop-core/build/backend';
-import { ProcessSyncContext } from './config';
-import { refreshItemPlaceholders } from './refresh-item-placeholders';
-
-type Props = {
-  ctx: ProcessSyncContext;
-};
-
-export const store = {
```
dajimenezriv, line 9: This was a global object so it will break once the default drive and workspaces were in the same process.

src/backend/features/remote-sync/file-explorer/delete-item-placeholders.ts
```diff
@@ -21,7 +22,7 @@ export function deleteItemPlaceholders({ ctx, remotes, type }: Props) {
      * we need to be sure that the one that we are checking is the same
      */
     if (localUuid === remote.uuid) {
-      ctx.virtualDrive.deleteFileSyncRoot({ path: remote.path });
```
dajimenezriv, line 24: We had some implementation in C++, however I was checking with claude and it does the same as the native remove from node.js (I've also check and does the same). This implementation is also removed in C++.

PR #1019, opened 2025-09-10
Stop using get file by path

dajimenezriv-internxt:


src/infra/drive-server-wip/services/files/check-existance.ts
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  if (res.error) {
+    switch (true) {
+      case res.error.response?.status === 404:
+        return { error: new CheckExistenceError('PARENT_NOT_FOUND', res.error.cause) };
+      default:
+        return { error: res.error };
+    }
+  }
```
egalvis27, line 40: It would be clearer to replace the switch (true) with an if/else. In this case, you are only evaluating the status of the response, and the if makes it more readable.

```suggestion
  if (res.error) {
    if (res.error.response?.status === 404){
      return { error: new CheckExistenceError('PARENT_NOT_FOUND', res.error.cause) };
    }  
    return { error: res.error };
  }
```

src/infra/drive-server-wip/services/files/check-existance.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const fileDto = res.data.existentFiles.pop();
+
+  if (fileDto) {
+    return { data: parseFileDto({ fileDto }) };
+  } else {
+    return { error: new CheckExistenceError('FILE_NOT_FOUND', res.data) };
+  }
```
egalvis27, line 48: The else is not necessary here because you already do a return inside the if. You could simplify it by using early return for better readability.

```suggestion
  if (fileDto) {
    return { data: parseFileDto({ fileDto }) };
  }
  return { error: new CheckExistenceError('FILE_NOT_FOUND', res.data) };
```

src/infra/drive-server-wip/services/folders/check-existence.ts
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  if (res.error) {
+    switch (true) {
+      case res.error.response?.status === 404:
+        return { error: new CheckExistenceError('PARENT_NOT_FOUND', res.error.cause) };
+      default:
+        return { error: res.error };
+    }
+  }
```
egalvis27, line 40: It would be clearer to replace the switch (true) with an if/else. In this case, you are only evaluating the status of the response, and the if makes it more readable.

```suggestion
  if (res.error) {
    if (res.error.response?.status === 404){
      return { error: new CheckExistenceError('PARENT_NOT_FOUND', res.error.cause) };
    }  
    return { error: res.error };
  }
```

src/infra/drive-server-wip/services/folders/check-existence.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const folderDto = res.data.existentFolders.pop();
+
+  if (folderDto) {
+    return { data: parseFolderDto({ folderDto }) };
+  } else {
+    return { error: new CheckExistenceError('FOLDER_NOT_FOUND', res.data) };
+  }
```
egalvis27, line 48: The else is not necessary here because you already do a return inside the if. You could simplify it by using early return for better readability.

```suggestion
  if (folderDto) {
    return { data: parseFolderDto({ folderDto }) };
  }
  return { error: new CheckExistenceError('FOLDER_NOT_FOUND', res.data) };
```

src/infra/drive-server-wip/services/folders/create-folder.ts
```diff
@@ -25,25 +25,18 @@ export async function createFolder(context: { path: string; body: TCreateFolderB
...
   });
 
-  if (error?.code === 'UNKNOWN') {
+  if (res.error?.code) {
     switch (true) {
-      case error.response?.status === 409:
-        return { error: new CreateFolderError('ALREADY_EXISTS', error.cause) };
+      case res.error.response?.status === 409:
+        return { error: new CreateFolderError('ALREADY_EXISTS', res.error.cause) };
     }
```
egalvis27, line 38: Here you are using a switch (true) with a single case, which is redundant and difficult to read. This can be simplified to an if (res.error?.response?.status === 409) { … }

src/infra/drive-server-wip/services/folders/rename-folder.ts
```diff
@@ -28,26 +28,17 @@ export async function renameFolder(context: { uuid: string; name: string; worksp
...
   });
 
-  if (res.error?.code === 'UNKNOWN') {
+  if (res.error) {
     switch (true) {
       case res.error.response?.status === 409:
         return { error: new RenameFolderError('FOLDER_ALREADY_EXISTS', res.error.cause) };
+      default:
+        return { error: res.error };
     }
```
egalvis27, line 40: It would be clearer to replace the switch (true) with an if/else. In this case, you are only evaluating the status of the response, and the if makes it more readable.

```suggestion
    if (res.error.response?.status === 409) {
      return { error: new RenameFolderError('FOLDER_ALREADY_EXISTS', res.error.cause) };
    }
    return { error: res.error };
```

egalvis27 (changes requested):

cg-shokworks (approved):

src/infra/drive-server-wip/services/files/check-existance.ts
carlosandromeda: I don't have like a strong preference here, maybe we can do the if approach if we only have one case (404) in this scenario, and do a switch in the future if we start adding more cases. But that's up to the dev imo

src/infra/drive-server-wip/services/folders/check-existence.ts
cg-shokworks: I don't have like a strong preference here, maybe if we have only one case we can do the if approach, and in the future if we start adding more cases we can do the switch approach. But again in my opinion, this is up to the dev

egalvis27 (approved):

PR #1020, opened 2025-09-18
Fix abort on change size

dajimenezriv-internxt:
## What

1. Remove .npmrc from the repository, we are going to use the packages from npm. Also, migrate all @internxt deps to npm in the package-lock.json.
2. Extract the Promise creation for the upload file into a separate file `upload-file.ts`.
3. Remove recursive call on abortOnChangeSize and use instead a setInterval.


src/infra/inxt-js/file-uploader/upload-file.ts
```diff
@@ -0,0 +1,58 @@
...
+  bucket: string;
+  readable: ReadStream;
+  size: number;
+  absolutePath: AbsolutePath;
+  abortSignal: AbortSignal;
+  path: string;
+  callbacks: FileUploaderCallbacks;
+};
+
+export function uploadFile({ fn, bucket, readable, size, absolutePath, abortSignal, path, callbacks }: Props) {
```
dajimenezriv, line 22: The code is extracted into another function so it's easier to test, it does the same except that now, we create an interval and we clear it when finishedCallback is called.

src/infra/inxt-js/file-uploader/abort-on-change-size.ts
```diff
@@ -1,24 +1,20 @@
 import { AbsolutePath } from '@/context/local/localFile/infrastructure/AbsolutePath';
```
dajimenezriv, line 1: Now this function doesn't need to call itself since it's already inside a `setInterval`

src/infra/inxt-js/file-uploader/upload-file.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      return new ActionState(ActionTypes.Upload);
+    });
+    // When
+    void uploadFile(props);
+    vi.advanceTimersByTime(50);
+    // Then
+    expect(readable.destroy).toBeCalledTimes(1);
+  });
+
+  it('should call abort on change size after 5s', () => {
```
dajimenezriv, line 95: We have the same four tests as before but now we have this one too.

egalvis27 (approved):

PR #1021, opened 2025-09-18
Broadcast just to widget

dajimenezriv-internxt:


.github/workflows/check-pr-size.yml
```diff
@@ -7,7 +7,8 @@ on:
 jobs:
   check_pr_size:
     # Ignore PRs with name merge-X-X-X-release
-    if: false
+    if: |
+      !(startsWith(github.head_ref, 'merge-') && endsWith(github.head_ref, '-release') && github.base_ref == 'main')
```
dajimenezriv, line 11: This is reverted since in the last PR I had to disable it.

.github/workflows/sonar-analysis.yml
```diff
@@ -41,12 +41,6 @@ jobs:
       - name: Merge coverage reports
         run: npm run coverage:merge
 
-      - name: Run report coverage
-        if: always()
-        uses: davelosert/vitest-coverage-report-action@8ab049ff5a2c6e78f78af446329379b318544a1a # v2.8.3
-        with:
-          json-summary-path: './coverage/unit/coverage-summary.json'
```
dajimenezriv, line 48: Since now we have 3 different test suites we don't have a unified coverage summary that works for this action, so we can remove it.

src/apps/main/windows/broadcast-to-windows.d.ts
```diff
@@ -2,36 +2,17 @@ import type { Issue } from '../background-processes/issues';
...
-      | 'UPLOADED'
-      | 'UPLOADING';
-    name: string;
-    key: FileUuid | AbsolutePath;
-    progress?: number;
-  };
-};
+export type BroadcastToWidget =
+  | { name: 'sync-info-update'; data: SyncStateItem[] }
+  | { name: 'remote-sync-status-change'; data: RemoteSyncStatus };
```
dajimenezriv, line 10: These events are just broadcasted to the widget instead of all browser windows since only the widget uses them.

src/apps/renderer/hooks/useSyncInfoSubscriber.tsx
```diff
@@ -1,37 +1,14 @@
 import { useEffect, useState } from 'react';
```
dajimenezriv, line 1: Now all filtering and sorting is done in the backend.

src/backend/features/local-sync/sync-state/services/add-item.ts
```diff
@@ -0,0 +1,21 @@
...
+    ...item,
+    time: Date.now(),
+    timeout: setTimeout(() => {
+      delete store.items[item.key];
+    }, 10000),
+  };
+
+  if (!store.notifyInterval) {
+    broadcastToFrontend();
+    store.notifyInterval = setInterval(broadcastToFrontend, 1000);
```
dajimenezriv, line 19: Once we add the first item we create the interval to broadcast to the frontend every 1s.

src/backend/features/local-sync/sync-state/services/add-item.ts
```diff
@@ -0,0 +1,21 @@
...
+export function addItem(item: SyncStateItem) {
+  const previous = store.items[item.key];
+  if (previous) clearTimeout(previous.timeout);
+
+  store.items[item.key] = {
+    ...item,
+    time: Date.now(),
+    timeout: setTimeout(() => {
+      delete store.items[item.key];
+    }, 10000),
```
dajimenezriv, line 14: Also the item is deleted after seconds of being added. However if we add the same key we clear the previous timeout and we create the new one.

PR #1022, opened 2025-09-19
Backups save to sqlite asap

dajimenezriv-internxt:
## What

1. Add abort signal to `replaceFile` endpoint, so if we stop backups or we logout the requests are cancelled.
2. In backups, when create or update a file in the cloud, save the response dto to the sqlite asap as we do with sync. Why? We will have at the moment all data in local so we will be able to move backups from checking the current state in the cloud to the local database (for users with 100k files this is going from 30min to 5s).

### DIFF
--- a/src/backend/features/remote-sync/update-in-sqlite/create-or-update-file.ts
+++ b/src/backend/features/remote-sync/update-in-sqlite/create-or-update-file.ts
@@ -1,8 +1,9 @@
+import { BackupsContext } from '@/apps/backups/BackupInfo';
 import { SyncContext } from '@/apps/sync-engine/config';
 import { FileDto } from '@/infra/drive-server-wip/out/dto';
 import { SqliteModule } from '@/infra/sqlite/sqlite.module';
 
-export async function createOrUpdateFile({ context, fileDto }: { context: SyncContext; fileDto: FileDto }) {
+export async function createOrUpdateFile({ context, fileDto }: { context: SyncContext | BackupsContext; fileDto: FileDto }) {
   return await SqliteModule.FileModule.createOrUpdate({
     file: {
       ...fileDto,


src/apps/backups/BackupInfo.ts
```diff
@@ -11,6 +11,8 @@ export type BackupInfo = {
 };
 
 export type BackupsContext = BackupInfo & {
```
dajimenezriv, line 13: To match the `SyncContext` we need the userUuid and the workspaceId (which in backups for now is always empty). In the future we will add the user directly to both contexts but for that we need to improve first the auth flow.

src/apps/main/remote-sync/files/sync-remote-files.ts
```diff
@@ -40,8 +40,8 @@ export async function syncRemoteFiles({ self, from, offset = 0 }: TProps) {
     offset += FETCH_LIMIT;
 
     await Promise.all(
-      data.map(async (remoteFile) => {
-        await syncRemoteFile({ self, remoteFile });
+      data.map(async (fileDto) => {
+        await createOrUpdateFile({ context: self.context, fileDto });
```
dajimenezriv: It's the same, the previous one `syncRemoteFile` was also logging an error in case of error, however, that was already done in the sqlite function, so it was logged two times.

src/apps/main/remote-sync/folders/sync-remote-folders.ts
```diff
@@ -40,8 +40,8 @@ export async function syncRemoteFolders({ self, from, offset = 0 }: TProps) {
     offset += FETCH_LIMIT;
 
     await Promise.all(
-      data.map(async (remoteFolder) => {
-        await syncRemoteFolder({ self, remoteFolder });
+      data.map(async (folderDto) => {
+        await createOrUpdateFolder({ context: self.context, folderDto });
```
dajimenezriv: Same as before, the previous one syncRemoteFolder was also logging an error in case of error, however, that was already done in the sqlite function, so it was logged two times.

src/context/local/localFile/application/update/FileBatchUpdater.ts
```diff
@@ -18,52 +19,40 @@ type Props = {
...
-          context,
-          tracker,
-          localFile: local,
-          file: remote,
-        }),
-      ),
+      modified.map(async ({ local, remote }) => {
+        await this.process({ context, localFile: local, file: remote });
+        self.backed++;
+        tracker.currentProcessed(self.backed);
```
dajimenezriv, line 25: This was executed in a finally block inside `process` function, however here we are more sure that it's always executed since in `process` we could have throws or early returns.

src/context/local/localFile/application/update/FileBatchUpdater.ts
```diff
@@ -18,52 +19,40 @@ type Props = {
...
+          uuid: file.uuid,
+          newContentId: contentsId,
+          newSize: localFile.size,
+          modificationTime: localFile.modificationTime.toISOString(),
+        },
+        { abortSignal: context.abortController.signal },
+      );
+
+      if (fileDto) {
+        await createOrUpdateFile({ context, fileDto });
```
dajimenezriv, line 47: Once we create it in the cloud save it to our local database.

src/context/local/localFile/application/upload/FileBatchUploader.ts
```diff
@@ -19,44 +20,48 @@ type Props = {
 
 export class FileBatchUploader {
   static async run({ self, context, tracker, remoteTree, added }: Props) {
```
dajimenezriv, line 22: The functionality is the same, but we have split in two functions `run` and `process` to match the code of the `FileBatchUpdater`. The `run` function updates all backed files and sends a notification. The `process` function uploads the file contents, creates the file, saves the file to sqlite and creates the thumbnail in case of image.

PR #1023, opened 2025-09-19
Improve register sync root

dajimenezriv-internxt:
## What

1. Update node-win version.
2. Improve creation of sync root folder. Previously it was created in two ways, now there is just one and we add the code of why it's created (NON_EXISTS, NON_ACCESS).
3. Move `access` and `login` requests to not throw any exception, otherwise it breaks the login in development.

### DIFF
--- a/src/apps/sync-engine/tests/create-placeholder.infra.test.ts
+++ b/src/apps/sync-engine/tests/create-placeholder.infra.test.ts
@@ -1,6 +1,6 @@
 import { join } from 'node:path';
 import { BindingsManager } from '../BindingManager';
-import { TEST_FILES } from 'tests/vitest/mocks.helper.test';
+import { loggerMock, TEST_FILES } from 'tests/vitest/mocks.helper.test';
 import { v4 } from 'uuid';
 import { getConfig, ProcessSyncContext, setDefaultConfig } from '../config';
 import { VirtualDrive } from '@/node-win/virtual-drive';
@@ -84,6 +84,7 @@ describe('create-placeholder', () => {
       return Promise.resolve();
     });
 
+    const fileUuid = v4() as FileUuid;
     createFileMock.mockResolvedValueOnce({
       data: {
         bucket: 'bucket',
@@ -98,7 +99,7 @@ describe('create-placeholder', () => {
         size: '1',
         status: 'EXISTS',
         updatedAt: new Date().toISOString(),
-        uuid: v4() as FileUuid,
+        uuid: fileUuid,
         modificationTime: new Date().toISOString(),
         plainName: 'plainName',
         userId: 1,
@@ -117,6 +118,7 @@ describe('create-placeholder', () => {
     const container = buildProcessContainer({ ctx });
 
     // When
+    await ctx.virtualDrive.createSyncRootFolder();
     await BindingsManager.start({ ctx, container });
     BindingsManager.watch({ ctx });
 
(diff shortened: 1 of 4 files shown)


src/apps/main/background-processes/sync-engine.ts
```diff
@@ -50,7 +50,7 @@ export const stopAndClearAllSyncEngineWatcher = async () => {
   );
 };
 
-export function spawnDefaultSyncEngineWorker({ context }: { context: AuthContext }) {
+export async function spawnSyncEngineWorkers({ context }: { context: AuthContext }) {
```
dajimenezriv, line 53: Now, spawn default sync engine worker and spawn workspaces have been joined in just one function. Why? Because previously the default sync engine worker was spawned before we unregistered the old virtual drives and it was leading to errors.

src/apps/main/background-processes/sync-engine/services/unregister-virtual-drives.ts
```diff
@@ -8,9 +8,22 @@ type TProps = {
 export function unregisterVirtualDrives({ currentProviderIds = [] }: TProps) {
```
dajimenezriv, line 8: As changed in node-win, now we filter the providers in the javascript code, so before unregistering anything (like Dropbox or OneDrive) we need to be sure that the displayName or the path contains the word `internxt`.

src/apps/renderer/pages/Login/index.tsx
```diff
@@ -27,14 +27,12 @@ export default function Login() {
```
dajimenezriv, line 27: Previously we were throwing an error when the `access` or `login` request was unsuccessful, and this works in production, however in development it displays an error message because we are sending an exception through ipc. So, instead of throwing we just send the error of the request and it's the same.

src/apps/sync-engine/index.ts
```diff
@@ -68,31 +69,28 @@ async function refreshToken({ ctx }: { ctx: ProcessSyncContext }) {
   }
 }
 
-ipcRenderer.once('SET_CONFIG', (event, config: Config) => {
```
dajimenezriv, line 71: The code is the same but now we have everything wrapper inside a try catch, otherwise we were not logging an error in case something failed.

src/apps/sync-engine/index.ts
```diff
@@ -9,15 +8,17 @@ import { buildFileUploader } from '../main/background-processes/backups/build-fi
...
 
 logger.debug({ msg: 'Running sync engine' });
 
-async function ensureTheFolderExist(path: string) {
-  try {
-    await fs.access(path);
-  } catch {
-    logger.debug({ msg: `Folder <${path}> does not exists, going to  create it` });
-    await fs.mkdir(path);
+async function createRootFolder({ ctx }: { ctx: ProcessSyncContext }) {
```
dajimenezriv: We have improved the way we create the sync root folder so we append the log the error.code (NOT_EXISTS, NOT_ACCESS).

src/apps/main/background-processes/sync-engine.ts
```diff
@@ -71,24 +71,16 @@ export function spawnDefaultSyncEngineWorker({ context }: { context: AuthContext
...
-  return { providerId };
-}
-
-export async function spawnWorkspaceSyncEngineWorkers({ context, providerId }: { context: AuthContext; providerId: string }) {
   const workspaces = await getWorkspaces();
   const workspaceProviderIds = workspaces.map((workspace) => workspace.providerId);
-
   const currentProviderIds = workspaceProviderIds.concat([providerId]);
 
   unregisterVirtualDrives({ currentProviderIds });
```
dajimenezriv, line 78: Previously we were doing `spawnSyncEngineWorker` before this function which was not good

src/apps/sync-engine/index.ts
```diff
@@ -12,23 +11,14 @@ import { buildProcessContainer } from './build-process-container';
 
 logger.debug({ msg: 'Running sync engine' });
 
-async function ensureTheFolderExist(path: string) {
```
dajimenezriv, line 15: Now this function is inside `VirtualDrive` class since it was in both places.

src/infra/drive-server-wip/services/auth.service.ts
```diff
@@ -12,44 +12,26 @@ export const auth = {
   logout,
 };
 
-async function access({ email, password, tfa }: { email: string; password: string; tfa?: string }) {
```
dajimenezriv, line 15: As said, don't throw exceptions neither in `access` nor in `login` requests.

src/node-win/virtual-drive.ts
```diff
@@ -50,17 +49,16 @@ export class VirtualDrive {
...
-  }
-
   getPlaceholderState({ path }: { path: string }) {
     return this.addon.getPlaceholderState({ path: this.fixPath(path) });
   }
 
-  createSyncRootFolder() {
-    if (!fs.existsSync(this.syncRootPath)) {
-      fs.mkdirSync(this.syncRootPath, { recursive: true });
+  async createSyncRootFolder() {
```
dajimenezriv, line 56: Now this is the only function that creates the sync root folder.

src/apps/renderer/pages/Login/index.tsx
```diff
@@ -27,14 +27,12 @@ export default function Login() {
...
 
-    try {
-      const res = await accessRequest({ email, password, hashedPassword: encryptedHash, tfa: twoFA });
-      window.electron.userLoggedIn({
-        ...res,
-        password,
-      });
-    } catch (err) {
-      const { message } = err as Error;
+    const { data, error } = await accessRequest({ email, password, hashedPassword: encryptedHash, tfa: twoFA });
```
cg-shokworks, line 30: Isn't it better here to just do res = await ....

so we maintain the inferred typing?

cg-shokworks (approved):

PR #1025, opened 2025-09-23
Get file explorer state

dajimenezriv-internxt:
## What

This is the first of 3 PRs that are going to simplify the way we read the current file explorer state and perform actions based on that on the sync process start. Currently we were checking if a file or folder did not have a placeholder and in that case we were creating them. We were also checking if the size of the remote file was different from the local one and update contents id in that case. And finally we were using a queue mananger to keep track of files that were pending for hydration so we can hydrate them when the app starts again.

Now, we are going to join all 4 into the same function so we just iterate the sync folder just one time. Also, we are going to stop using the queue manager since we are going to check the PinState of a file in the explorer and the number of blocks it have to know if the file is hydrated or not and if it should.

### DIFF
--- a/src/apps/sync-engine/file-explorer-state/get-file-explorer-state.ts
+++ b/src/apps/sync-engine/file-explorer-state/get-file-explorer-state.ts
@@ -1,0 +1,43 @@
+import { logger } from '@/apps/shared/logger/logger';
+import { fileSystem } from '@/infra/file-system/file-system.module';
+import { ProcessSyncContext } from '../config';
+import { processItem } from './process-item';
+import { getExistingFiles } from '@/context/virtual-drive/items/application/RemoteItemsGenerator';
+import { FileExplorerState } from './file-explorer-state.types';
+
+type TProps = {
+  ctx: ProcessSyncContext;
+};
+
+export async function getFileExplorerState({ ctx }: TProps) {
+  const rootFolder = ctx.virtualDrive.syncRootPath;
+
+  logger.debug({ tag: 'SYNC-ENGINE', msg: 'Get file explorer state', rootFolder });
+
+  const state: FileExplorerState = {
+    createFiles: [],
+    createFolders: [],
+    hydrateFiles: [],
+    modifiedFiles: [],
+  };
+
+  const remoteFiles = await getExistingFiles({ ctx });
+  const remoteFilesMap = Object.fromEntries(remoteFiles.map((file) => [file.uuid, file]));
+
+  const localItems = await fileSystem.syncWalk({ rootFolder });
+
+  for (const localItem of localItems) {
+    processItem({ ctx, localItem, state, remoteFilesMap });
+  }
+
+  logger.debug({
+    tag: 'SYNC-ENGINE',
+    msg: 'File explorer state',
+    createFiles: state.createFiles.map((file) => file.path),
+    createFolders: state.createFolders.map((file) => file.path),
+    hydrateFiles: state.hydrateFiles.map((file) => file.path),
+    modifiedFiles: state.modifiedFiles.map((file) => file.path),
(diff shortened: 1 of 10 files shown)


src/apps/sync-engine/file-explorer-state/get-file-explorer-state.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const state: FileExplorerState = {
+    createFiles: [],
+    createFolders: [],
+    hydrateFiles: [],
+    modifiedFiles: [],
+  };
+
+  const remoteFiles = await getExistingFiles({ ctx });
+  const remoteFilesMap = Object.fromEntries(remoteFiles.map((file) => [file.uuid, file]));
```
dajimenezriv, line 25: We need to obtain also the remote files from the sqlite to check the contents id, so we know if the size is different and we need to upload the content.

src/apps/sync-engine/file-explorer-state/is-hydration-pending.ts
```diff
@@ -0,0 +1,18 @@
...
+  pinState: PinState;
+};
+
+export function isHydrationPending({ stats, pinState }: Props) {
+  if (pinState === PinState.AlwaysLocal) {
+    const expectedBlocks = Math.ceil(stats.size / 512);
+    const actualBlocks = stats.blocks;
+    const ratio = actualBlocks / expectedBlocks;
+    return ratio < 1;
+  }
```
dajimenezriv, line 15: If the file was marked as AlwaysLocal but we have less blocks than the size of the file it means that it's not fully hydrated and we need to hydrate it.

src/apps/sync-engine/file-explorer-state/is-modified.ts
```diff
@@ -0,0 +1,40 @@
...
+import { PinState } from '@/node-win/types/placeholder.type';
+import { FileExplorerFile, RemoteFilesMap } from './file-explorer-state.types';
+
+type Props = {
+  remoteFilesMap: RemoteFilesMap;
+  localFile: FileExplorerFile;
+  pinState: PinState;
+};
+
+export function isModified({ remoteFilesMap, localFile, pinState }: Props) {
```
dajimenezriv, line 11: This was as before. See `sync-modified-file.ts`

src/apps/sync-engine/file-explorer-state/process-item.ts
```diff
@@ -0,0 +1,50 @@
...
+import { FileExplorerState, RemoteFilesMap } from './file-explorer-state.types';
+
+type Props = {
+  ctx: ProcessSyncContext;
+  localItem: SyncWalkItem;
+  state: FileExplorerState;
+  remoteFilesMap: RemoteFilesMap;
+};
+
+export function processItem({ ctx, localItem, state, remoteFilesMap }: Props) {
```
dajimenezriv, line 16: Here we check if we have to create the file, the folder, hydrate the file or upload the new file content.

src/apps/sync-engine/file-explorer-state/is-hydration-pending.test.ts
```diff
@@ -0,0 +1,53 @@
...
+  beforeEach(() => {
+    props = mockProps<typeof isHydrationPending>({
+      pinState: PinState.AlwaysLocal,
+      stats: {
+        blocks: 2,
+        size: 8 * 512,
+      },
+    });
+  });
+
```
egalvis27, line 17: Shouldn't there be a true case within the tests?

dajimenezriv: Fixed

egalvis27 (approved):

PR #1026, opened 2025-09-23
Remove queue manager

dajimenezriv-internxt:
## What

This is the second of 3 PRs that are going to simplify the way we read the current file explorer state and perform actions based on that on the sync process start. Currently we were checking if a file or folder did not have a placeholder and in that case we were creating them. We were also checking if the size of the remote file was different from the local one and update contents id in that case. And finally we were using a queue mananger to keep track of files that were pending for hydration so we can hydrate them when the app starts again.

Now, we are going to join all 4 into the same function so we just iterate the sync folder just one time. Also, we are going to stop using the queue manager since we are going to check the PinState of a file in the explorer and the number of blocks it have to know if the file is hydrated or not and if it should.

## In this PR

We add a throttle to the hydrate function because we don't want to hydrate more than one files at a time (probably we can already parallelize, but we need to do more checks). Replace the old implementation with the new one.

### DIFF
--- a/src/apps/sync-engine/file-explorer-state/is-hydration-pending.test.ts
+++ b/src/apps/sync-engine/file-explorer-state/is-hydration-pending.test.ts
@@ -42,7 +42,7 @@ describe('is-hydration-pending', () => {
     expect(res).toBe(false);
   });
 
-  it('should return false if blocks are less than size', () => {
+  it('should return true if blocks are less than size', () => {
     // Given
     props.stats.blocks = 7;
     // When


src/apps/sync-engine/callbacks/handle-hydrate.ts
```diff
@@ -1,37 +1,40 @@
...
       msg: 'Error hydrating file',
-      path: task.path,
+      path,
       error,
     });
   }
 }
+
+export async function throttleHydrate(props: TProps) {
+  return await limiter.schedule(() => handleHydrate(props));
```
dajimenezriv, line 39: We still don't want to run downloads in parallel since maybe C++ throws an unexpected exception and the sync process breaks, so for now we throttle it.

src/apps/sync-engine/in/add-pending-items.ts
```diff
@@ -1,33 +1,27 @@
...
 export async function addPendingItems({ ctx }: Props) {
   try {
-    const { pendingFiles, pendingFolders } = await getPendingItems({
-      ctx,
-      path: ctx.virtualDrive.syncRootPath,
-    });
-
     const startTime = performance.now();
 
-    logger.debug({
```
dajimenezriv, line 21: This log is already inside `get-file-explorer-state.ts`

src/apps/sync-engine/file-explorer-state/is-hydration-pending.test.ts
```diff
@@ -42,7 +42,7 @@ describe('is-hydration-pending', () => {
     expect(res).toBe(false);
   });
 
-  it('should return true if blocks are less than size', () => {
+  it('should return false if blocks are less than size', () => {
```
egalvis27: It changed back from true to false.

egalvis27 (approved):

PR #1028, opened 2025-09-25
Fix 2.5.8 release

dajimenezriv-internxt:
## What

1. Add a try catch for the unregisterSyncRoot function.
2. Modify the props required to create a file or folder placeholder (now node-win requires less props).

This PR was already in qa build 3 and it's already tested by Valentyna.

### DIFF
--- a/src/node-win/virtual-drive.unit.test.ts
+++ b/src/node-win/virtual-drive.unit.test.ts
@@ -96,10 +96,9 @@ describe('VirtualDrive', () => {
         'file.txt',
         'FILE:uuid',
         1024,
-        1,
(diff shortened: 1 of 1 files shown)


src/node-win/virtual-drive.ts
```diff
@@ -32,10 +25,6 @@ export class VirtualDrive {
     this.addon.addLogger({ path: this.convertToWindowsPath({ path: loggerPath }) });
   }
 
-  private convertToWindowsTime(jsTime: number) {
```
dajimenezriv, line 35: Now, we convert the time number from javascript to windows format inside node-win.

PR #1037, opened 2025-09-29
Add recovery sync 1

dajimenezriv-internxt:
## What

First PR of 3 PRs to implement recovery sync.

1. Replace `getMockCalls` with `call` or `calls`. It's the same but the second ones are already wrapped inside `expect`.
2. Add `parseFileDto` to get files responses so that `uuid` becomes a `FileUuid`.
3. Add the function `getBetweenUuids` to obtain in the sqlite all files that are between two `uuids`.


src/apps/backups/folders/create-folders.test.ts
```diff
@@ -69,7 +69,7 @@ describe('create-folders', () => {
     await createFolders(props);
 
     // Then
-    expect(createFolderMock).not.toHaveBeenCalled();
```
dajimenezriv, line 72: `call` and `calls` is the new way we use to assert mock and spy calls. I migrated all old usages so we can remove the function `getMockCalls`, which was using the old way.

src/infra/sqlite/services/file/get-between-uuids.ts
```diff
@@ -0,0 +1,34 @@
...
+export async function getBetweenUuids({ workspaceId, firstUuid, lastUuid }: Props) {
+  try {
+    const items = await fileRepository.find({
+      order: { uuid: 'ASC' },
+      where: {
+        workspaceId,
+        status: 'EXISTS',
+        uuid: Between(firstUuid, lastUuid),
+      },
+    });
```
dajimenezriv, line 23: We need to obtain all local files that are between 2 uuids and sorted by uuids. We will do the same request to `drive-server-wip` so we can compare what we have in local and in remote.

src/apps/backups/backups.test.ts
```diff
@@ -134,7 +134,7 @@ describe('backups', () => {
 
     expect(loggerMock.error).toBeCalledTimes(0);
```
sg-gs, line 135: Why these two expects are not replaced?

src/infra/drive-server-wip/drive-server-wip.module.ts
```diff
@@ -2,17 +2,20 @@ import { auth } from './services/auth.service';
...
   auth,
   backup,
   files,
+  FileModule,
   folders,
   storage,
   user,
 };
 export const driveServerWipModule = driveServerWip;
+export const DriveServerWipModule = driveServerWip;
```
sg-gs, line 21: Why two exports of the same thing with different names? I assume you want to remove the first one which is not using PascalCase, thus not following the typical naming conventions about exported modules naming

src/node-win/watcher/watcher.unit.test.ts
```diff
@@ -27,7 +27,7 @@ describe('Watcher', () => {
       await sleep(50);
 
       // Assert
-      expect(getEvents()).toStrictEqual([{ event: 'addDir', path: syncRootPath }]);
+      getEvents().toStrictEqual([{ event: 'addDir', path: syncRootPath }]);
```
sg-gs, line 30: Are you sure that the mix of hidden `expects` + hidden `calls` is also worth it? It seems not too intuitive. I had to check what getCalls was doing to know that the calls was being used there also

src/infra/sqlite/services/file/get-between-uuids.ts
sg-gs: I assume the sort + the between is to have some kind of checkpoint, correct? As the UUIDs are completely random and do not imply any kind of hierarchy order on the folders tree

dajimenezriv: Exactly. We can't neither use `updatedAt` since there are duplicated values nor `id` since it's going to be deleted.

dajimenezriv: https://inxt.atlassian.net/browse/PB-5013?focusedCommentId=44917

src/infra/drive-server-wip/drive-server-wip.module.ts
dajimenezriv: Exactly.

src/apps/backups/backups.test.ts
dajimenezriv: We have many expects to migrate. The main point was to stop using `getMockCalls`. Maybe I migrate some expects but it was not the main purpose.

src/node-win/watcher/watcher.unit.test.ts
dajimenezriv: The `call` and `calls` now it's the default and they are easy to understand once you use them a couple of times. @egalvis27 is already using them.

About the `getEvents` maybe is not. It was just the way it was implemented before. Also, it's not something global, it only affects the tests that are from the watcher and it's repeated in many places among those tests (currently only 2 tests files), that's why it was extracted.

dajimenezriv: The main purpose it was that `watcher.unit.test.ts` was a test file too long and we wanted to split it in different files, so we created the `watcher.helper.test.ts` to reuse functionality among the new test files (currently only the one for adding files has been extracted).

PR #1039, opened 2025-10-01
Add recovery sync 2

dajimenezriv-internxt:


src/apps/main/database/entities/DriveFile.ts
```diff
@@ -25,21 +25,36 @@ export type ExtendedDriveFile = SimpleDriveFile & {
```
dajimenezriv, line 25: The fields are the same, I've just sorted them so they appear in a more useful order when opening using a sqlite viewer. This helps the developer to have in the same screen all useful information so we can develop faster.

src/apps/main/remote-sync/files/sync-remote-files.ts
```diff
@@ -26,6 +26,8 @@ export async function syncRemoteFiles({ self, from, offset = 0 }: TProps) {
       offset,
       status: from ? ('ALL' as const) : ('EXISTS' as const),
       updatedAt: from?.toISOString(),
+      sort: 'updatedAt',
+      order: 'ASC',
     };
```
dajimenezriv, line 31: Fetch to `drive-server-wip`. This is already by default but we want to ensure it.

src/backend/features/sync/recovery-sync/services/files-recovery-sync.ts
```diff
@@ -0,0 +1,66 @@
...
+  offset: number;
+};
+
+export async function filesRecoverySync({ ctx, offset }: Props) {
+  const query = {
+    limit: FETCH_LIMIT,
+    offset,
+    status: 'EXISTS' as const,
+    sort: 'uuid',
+    order: 'ASC',
```
dajimenezriv, line 21: Sort by uuid as we do in local.

src/apps/main/background-processes/sync-engine/services/spawn-sync-engine-worker.test.ts
```diff
@@ -5,7 +5,7 @@ import { BrowserWindow } from 'electron';
 import { monitorHealth } from './monitor-health';
 import { scheduleSync } from './schedule-sync';
 import { loggerMock } from 'tests/vitest/mocks.helper.test';
-import { RemoteSyncModule } from '@/backend/features/remote-sync/remote-sync.module';
+// import { RemoteSyncModule } from '@/backend/features/remote-sync/remote-sync.module';
```
sg-gs: Remove this comment

src/apps/main/background-processes/sync-engine/services/spawn-sync-engine-worker.test.ts
```diff
@@ -15,7 +15,7 @@ vi.mock(import('@/backend/features/remote-sync/remote-sync.module'));
 describe('spawn-sync-engine-worker', () => {
   const monitorHealthMock = vi.mocked(monitorHealth);
   const scheduleSyncMock = vi.mocked(scheduleSync);
-  const RemoteSyncModuleMock = vi.mocked(RemoteSyncModule);
+  // const RemoteSyncModuleMock = vi.mocked(RemoteSyncModule);
```
sg-gs: Remove this comment

src/apps/main/background-processes/sync-engine/services/spawn-sync-engine-worker.test.ts
```diff
@@ -31,7 +31,7 @@ describe('spawn-sync-engine-worker', () => {
     expect(BrowserWindow).toHaveBeenCalledTimes(1);
     expect(monitorHealthMock).toHaveBeenCalledTimes(1);
     expect(scheduleSyncMock).toHaveBeenCalledTimes(1);
-    expect(RemoteSyncModuleMock.syncItemsByFolder).toHaveBeenCalledTimes(1);
+    // expect(RemoteSyncModuleMock.syncItemsByFolder).toHaveBeenCalledTimes(1);
```
sg-gs: Remove this comment

src/backend/features/sync/recovery-sync/services/files-recovery-sync.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const { data: remotes } = ctx.workspaceId
+    ? await DriveServerWipModule.WorkspaceModule.getFilesInWorkspace({ workspaceId: ctx.workspaceId, query })
+    : await DriveServerWipModule.FileModule.getFiles({ query });
+
+  if (!remotes) return [];
+
+  const locals = await getLocalFiles({ ctx, remotes });
+
+  if (!locals) return [];
```
sg-gs, line 32: If there are remotes and no locals, why not continuing?

src/backend/features/sync/recovery-sync/services/files-recovery-sync.ts
```diff
@@ -0,0 +1,66 @@
...
+        msg: 'File does not have parentUuid and cannot be recreated',
+        uuid: file.uuid,
+      });
+
+      return Promise.resolve();
+    }
+
+    /**
+     * v2.6.0 Daniel Jiménez
+     * This should never happen. Basically if we reach this point it means that there was an
```
sg-gs: If the item is marked as "DELETED" in the remote, be sure that the content is gone, so I assume this could be a fallback, but by a re-upload, not by a movement

src/backend/features/sync/recovery-sync/services/get-local-files.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  it('should return if there are no remotes', async () => {
+    // Given
+    props.remotes = [];
+    // When
+    const res = await getLocalFiles(props);
+    // Then
+    expect(res).toBeUndefined();
+  });
+
+  it('should return if there are is an error retrieving locals', async () => {
```
sg-gs: typo here, _are is_ instead of _is_

sg-gs: If this is the behaviour, you are ignoring failures on a recovery sync or what is the idea?

src/backend/features/sync/recovery-sync/services/get-local-files.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  it('should return empty if locals is empty', async () => {
+    // Given
+    getBetweenUuidsMock.mockResolvedValue({ data: [] });
+    // When
+    const res = await getLocalFiles(props);
+    // Then
+    expect(res).toStrictEqual([]);
+  });
+
+  it('should return locals', async () => {
```
sg-gs, line 44: should return locals when? if there is at least one present or when?

src/backend/features/sync/recovery-sync/services/get-local-files.ts
```diff
@@ -0,0 +1,28 @@
+import { SyncContext } from '@/apps/sync-engine/config';
+import { ParsedFileDto } from '@/infra/drive-server-wip/out/dto';
+import { SqliteModule } from '@/infra/sqlite/sqlite.module';
+
+export async function getLocalFiles({ ctx, remotes }: { ctx: SyncContext; remotes: ParsedFileDto[] }) {
+  const first = remotes.at(0);
+  const last = remotes.at(-1);
+
+  if (!first || !last) return;
```
sg-gs, line 9: Why?

dajimenezriv: It means remote array is empty and we don't have any remote item.

src/backend/features/sync/recovery-sync/services/files-recovery-sync.ts
dajimenezriv: The only way we locals is undefined is because there were no remotes or there was an error reading the sqlite database. If we don't have locals (empty array) it doesn't enter in the if. `Boolean([]) -> true`.

src/backend/features/sync/recovery-sync/services/get-local-files.test.ts
dajimenezriv: Retrieving locals already logs when there is an error, we don't need to log it here and there is no need to do anything more. It's a recovery sync, if it fails for one 1000 items execution it will work another time.

src/backend/features/sync/recovery-sync/services/get-local-files.test.ts
dajimenezriv: It's like the previous tests are the condition to early return or throw or similar and the last test is when everything works as expected and it reaches the last return, so there is like no condition. I don't have an strong opinion about this, so I'm open to alternatives.

PR #1040, opened 2025-10-01
Add create logger

dajimenezriv-internxt:


src/apps/main/background-processes/sync-engine/services/spawn-sync-engine-worker.ts
```diff
@@ -95,26 +92,24 @@ export async function spawnSyncEngineWorker({ context }: TProps) {
         : path.join(__dirname, '..', 'sync-engine', 'index.html'),
     );
 
-    logger.debug({ msg: '[MAIN] Browser window loaded', workspaceId });
+    ctx.logger.debug({ msg: 'Browser window loaded' });
 
-    browserWindow.webContents.send('SET_CONFIG', context);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { logger, ...config } = ctx;
```
dajimenezriv, line 98: We need to extract the logger from the context because we cannot send it through the ipc since it's a function. This will be removed when move the sync to the main process.

cg-shokworks (approved):

Issue #835, opened 2025-06-18
Feature Request: Customizable Storage Drive Location


JaviertINC:
Currently, the application installs and defaults to using the `C:` drive for storage. There's no option to change this location, either during installation or within the app's settings.

This improvement would allow:
- Save space on their primary C: drive.
- Utilize a dedicated hard drive, which is ideal for large data volumes or performance needs.
- Better organize their hard drives, dedicating specific drives for application storage.

Plarwes:
Would like this too!

PR #1042, opened 2025-10-01
feat: allow users to choose custom sync folder location

JaviertINC:
- Add onboarding slide for selecting sync folder during initial setup
- Display current sync root path in General settings with change option
- Modify sync root logic to persist custom paths instead of always defaulting
- Add IPC handlers for getting and setting sync root path
- Add translations for new UI elements in EN, ES, FR

Resolves [#835](https://github.com/internxt/drive-desktop/issues/835)

### DIFF
--- a/src/apps/main/preload.ts
+++ b/src/apps/main/preload.ts
@@ -157,6 +157,9 @@ const api = {
   chooseSyncRootWithDialog(): ReturnType<typeof chooseSyncRootWithDialog> {
     return ipcRenderer.invoke('choose-sync-root-with-dialog');
   },
+  getSyncRoot(): Promise<string> {
+    return ipcRenderer.invoke('get-sync-root');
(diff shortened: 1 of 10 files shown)


filipe-silva:
Can we expect this on the next version? @dajimenezriv

dajimenezriv:
Hey @JaviertINC @filipe-silva, let me check the PR in detail and test it.

Currently, we are finishing the cleaner for the release 2.6.0 and the release 2.6.1 is going to be soon after the 2.6.0 adding a utility to create a zip containing the logs and the sqlite database so we can handle better the bugs.

Once we finish that, I will check this.

dajimenezriv:
Wtf, I think I closed this PR without knowing it.. Reopened here https://github.com/internxt/drive-desktop/pull/1102.

PR #1043, opened 2025-10-02
Remove rename operation

dajimenezriv-internxt:
## What

Now the move endpoint also accepts the name and extension for files and the name for folders, so we can move and rename with just one endpoint which makes the code simpler. Previously we were checking if an items was moved, renamed, both or none, now we just call move and that handles it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "init:ci": "npm ci --ignore-scripts && node node_modules/electron/install.js",
     "init:dev": "npm install && node node_modules/electron/install.js && npm run build:dll && npm run reload-native-deps",
     "========== Code style ==========": "",
-    "lint": "eslint . --ext .ts,.tsx --max-warnings 406",
+    "lint": "eslint . --ext .ts,.tsx --max-warnings 408",
     "lint:fix": "npm run lint -- --fix",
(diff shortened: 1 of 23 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "init:ci": "npm ci --ignore-scripts && node node_modules/electron/install.js",
     "init:dev": "npm install && node node_modules/electron/install.js && npm run build:dll && npm run reload-native-deps",
     "========== Code style ==========": "",
-    "lint": "eslint . --ext .ts,.tsx --max-warnings 406",
```
dajimenezriv, line 19: This is reverted in the next PR because I had to leave two functions without use so the check size pipelines passes.

src/backend/features/local-sync/watcher/events/rename-or-move/move-item.ts
```diff
@@ -1,8 +1,8 @@
 import { AbsolutePath, RelativePath } from '@/context/local/localFile/infrastructure/AbsolutePath';
```
dajimenezriv, line 1: Previously we had to check if the item was renamed, moved, both or none. Now we don't need to do that. We just call move and that just works.

PR #1046, opened 2025-10-03
Add recovery sync 3

dajimenezriv-internxt:
## What

Implement recovery sync. This is going to run always every 1 minute until the user logs out (abort signal is aborted).

First we retrieve 1000 items from drive-server-wip sorted by uuid. Then we retrieve the items from local that are between the first and last remote uuid sorted by uuids. We obtain the items that are in local but not in remote, the ones that are in remote and not in local and the ones that have different updatedAt.

If the item doesn't exist locally or has different updatedAt, then update it locally.
If the item doesn't exist in remote (probably is TRASHED/REMOVED) try to mark it as EXISTS, since we don't want to remove anything locally.

## Added

We don't want to analyze items that are newer that the current checkpoint, why? Because for sure they are not yet in the local database and we are still waiting to be synchronized by the default sync by checkpoint. If we don't filter by checkpoint we are going to have error logs saying that `The item doesn't exist in local`.

### DIFF
--- a/src/backend/features/sync/recovery-sync/services/get-items-to-sync.test.ts
+++ b/src/backend/features/sync/recovery-sync/services/get-items-to-sync.test.ts
@@ -12,7 +12,6 @@ describe('get-items-to-sync', () => {
 
   beforeEach(() => {
     getCheckpointMock.mockResolvedValue({ data: 'datetime' });
-    isItemToSyncMock.mockReturnValue(true);
 
     props = mockProps<typeof getItemsToSync>({
       remotes: [{ uuid: 'uuid' as FileUuid, updatedAt: 'datetime' }],
@@ -20,7 +19,7 @@ describe('get-items-to-sync', () => {
     });
   });
 
-  it('should return empty if not checkpoint', async () => {
+  it('should return empty if there is no checkpoint', async () => {
     // Given
     getCheckpointMock.mockResolvedValue({ data: undefined });
     // When
@@ -29,7 +28,7 @@ describe('get-items-to-sync', () => {
     expect(res).toHaveLength(0);
   });
 
-  it('should return empty if is not item to sync', async () => {
+  it('should return empty if it is not item to sync', async () => {
     // Given
     isItemToSyncMock.mockReturnValue(false);
     // When
@@ -38,7 +37,9 @@ describe('get-items-to-sync', () => {
     expect(res).toHaveLength(0);
   });
 
-  it('should return remotes', async () => {
+  it('should return remote if it is item to sync', async () => {
+    // Given
+    isItemToSyncMock.mockReturnValue(true);
     // When
     const res = await getItemsToSync(props);
     // Then


src/apps/main/background-processes/sync-engine/services/spawn-sync-engine-worker.ts
```diff
@@ -46,10 +46,7 @@ export async function spawnSyncEngineWorker({ ctx }: TProps) {
    * Since we can have a different status in our local database that in remote,
    * we want to run also this sync in background to update the statuses.
    */
-  void RemoteSyncModule.syncItemsByFolder({
-    rootFolderUuid: ctx.rootUuid,
-    context: ctx,
-  });
+  void RecoverySyncModule.recoverySync({ ctx });
```
dajimenezriv, line 49: Replace the old recovery sync by the new one.

src/backend/features/sync/recovery-sync/services/get-items-to-sync.ts
```diff
@@ -1,39 +1,25 @@
...
 
 type Props = FileProps | FolderProps;
 
-export function getItemsToSync(props: FolderProps): ParsedFolderDto[];
-export function getItemsToSync(props: FileProps): ParsedFileDto[];
-export function getItemsToSync({ ctx, remotes, locals }: Props) {
-  const localFilesMap = new Map(locals.map((file) => [file.uuid, file]));
-
-  const itemsToSync = remotes.filter((remote) => {
-    const local = localFilesMap.get(remote.uuid);
```
dajimenezriv, line 12: We have extracted these checks into the file `is-item-to-sync.ts`.

src/backend/features/sync/recovery-sync/services/is-item-to-sync.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ParsedFileDto, ParsedFolderDto } from '@/infra/drive-server-wip/out/dto';
+
+type Props = {
+  ctx: SyncContext;
+  checkpointDate: Date;
+  remote: ParsedFileDto | ParsedFolderDto;
+  localsMap: Map<FileUuid | FolderUuid, SimpleDriveFile | SimpleDriveFolder>;
+};
+
+export function isItemToSync({ ctx, localsMap, remote, checkpointDate }: Props) {
```
dajimenezriv, line 13: This is the code extracted from `get-items-to-sync.ts` but now we also check the `updatedAt`.

src/infra/drive-server-wip/services/files.service.ts
```diff
@@ -24,27 +24,22 @@ export const FileModule = files;
...
       params: { query: context.query },
     });
 
   const { data, error } = await clientWrapper({
     promiseFn,
     key,
-    loggerBody: {
-      msg: 'Get files request',
-      context,
-      attributes: {
```
dajimenezriv, line 43: `attributes` was for open telemetry, so it's no longer needed.

src/backend/features/sync/recovery-sync/services/files-recovery-sync.ts
```diff
@@ -13,6 +13,7 @@ type Props = {
 };
 
 export async function filesRecoverySync({ ctx, offset }: Props) {
+  const extra = { abortSignal: ctx.abortController.signal, skipLog: true };
```
sg-gs, line 16: This naming is not descriptive, why not using `configuration` or something like that?

src/backend/features/sync/recovery-sync/services/is-item-to-sync.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (new Date(remote.updatedAt) >= checkpointDate) {
+    return false;
```
sg-gs, line 15: Shouldn't this be true unless there is a strict equality?

dajimenezriv: The recovery sync is meant to recover just the items that should have been fetched but they are not in the sqlite. If an item updatedAt is equal or greater than checkpoint it means that the default sync has not reached yet that item.

dajimenezriv: The checkpointDate contains the updatedAt from the last item fetched by the default sync by checkpoint. Any item that is equal to the checkpoint or after it should be ignored.

src/backend/features/sync/recovery-sync/services/files-recovery-sync.ts
dajimenezriv: Now it's because we have been using `extra` in other parts of the code, so to keep it consistent. But yes, maybe it's not the best name. Currenty a request to `drive-server-wip` accepts `context`, which contains the props to make the request (body, query params...) and `extra`, which contains extra args to the request (currently only abortSignal and skipLog).

dajimenezriv: I think that we can leave it for now since we will have to change it in other places.

sg-gs: Consider to change that in the future as it is generic

PR #1049, opened 2025-10-07
Feat/add cleaner backend

egalvis27:


src/backend/features/cleaner/services/emit-progress.ts
```diff
@@ -0,0 +1,11 @@
+import { BrowserWindow } from 'electron';
```
dajimenezriv, line 1: Two things, do we need to send to all windows? Is this the same as `broadcast-to-windows.ts`?

PR #1058, opened 2025-10-09
Add tests for cache and storage report generation in Cleaner module

egalvis27:


src/backend/features/cleaner/reports/generate-windows-specific-file-report.test.ts
```diff
@@ -0,0 +1,58 @@
...
+import { generateWindowsSpecificFileReport } from './generate-windows-specific-file-report';
+import { pathsToClean } from './paths-to-clean';
+import * as generateReportModule from './generate-report';
+import { calls, call, partialSpyOn } from '@/tests/vitest/utils.helper.test';
+
+describe('generateWindowsSpecificFileReport', () => {
+  const scanDirectoryMock = partialSpyOn(CleanerModule, 'scanDirectory');
+  const generateReportMock = partialSpyOn(generateReportModule, 'generateReport');
+
+  const mockCleanableItems = [
```
dajimenezriv: Are we using the props `fileName` or `sizeInBytes` in the test?

src/backend/features/cleaner/reports/generate-windows-specific-file-report.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    calls(scanDirectoryMock).toMatchObject([
+      {
+        dirPath: pathsToClean.windowsSpecific.windowsUpdateCache,
+      },
+      {
+        dirPath: pathsToClean.windowsSpecific.prefetch,
+      },
+    ]);
+    calls(generateReportMock).toHaveLength(1);
+    call(generateReportMock).toMatchObject({
```
dajimenezriv, line 28: Doing `call` you already tests that it's only one call, so the line from above is not necessary.

src/backend/features/cleaner/reports/generate-windows-specific-file-report.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    // Given
+    const windowsUpdateCacheItems = [mockCleanableItems[0]];
+    const prefetchItems = [mockCleanableItems[1]];
+
+    scanDirectoryMock.mockResolvedValueOnce(windowsUpdateCacheItems).mockResolvedValueOnce(prefetchItems);
+    // When
+    const result = await generateWindowsSpecificFileReport();
+    // Then
+    calls(scanDirectoryMock).toHaveLength(2);
+    calls(scanDirectoryMock).toMatchObject([
```
dajimenezriv, line 17: Asserting agains two values of an array already checks that the length is 2, so the above line is not needed.

src/backend/features/cleaner/reports/generate-windows-specific-file-report.test.ts
egalvis27: Technically, only the file size is needed for this object, but I kept the other properties to make its purpose clearer.

src/backend/features/cleaner/reports/generate-report.test.ts
```diff
@@ -0,0 +1,67 @@
+import { generateReport } from './generate-report';
+import { CleanableItem } from '@internxt/drive-desktop-core/build/backend/features/cleaner/types/cleaner.types';
+
+describe('generateReport', () => {
+  const mockCleanableItems1: CleanableItem[] = [
+    {
+      fullPath: 'G:\\Users\\User\\file1.tmp',
```
dajimenezriv: Are we using in the function `fulPath` or `fileName`?

dajimenezriv-internxt (approved):

PR #1059, opened 2025-10-09
add tests for cleaner report generation and related modules

egalvis27:


src/backend/features/cleaner/reports/generate-cleaner-report.test.ts
```diff
@@ -0,0 +1,99 @@
...
+import { loggerMock } from '@/tests/vitest/mocks.helper.test';
+
+describe('generateCleanerReport', () => {
+  const generateAppCacheReportMock = partialSpyOn(generateAppCacheReportModule, 'generateAppCacheReport');
+  const generateLogFilesReportMock = partialSpyOn(generateLogFilesReportModule, 'generateLogFilesReport');
+  const generateWebStorageFileReportMock = partialSpyOn(generateWebStorageFileReportModule, 'generateWebStorageFileReport');
+  const generateWebCacheReportMock = partialSpyOn(generateWebCacheReportModule, 'generateWebCacheReport');
+  const generateWindowsSpecificFileReportMock = partialSpyOn(generateWindowsSpecificFileReportModule, 'generateWindowsSpecificFileReport');
+
+  beforeEach(() => {
```
dajimenezriv: This `beforeEach` can be removed, `partialSpyOn` does empty mocks by default so we don't call the real implementation if we forget to mock.

PR #1060, opened 2025-10-10
feat: add recycle bin report generation and processing

egalvis27:


src/backend/features/cleaner/reports/trash-bin/process-recycle-bin-output.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  logger.debug({
+    tag: 'CLEANER',
+    msg: `PowerShell returned ${lines.length} lines`,
+  });
+
+  for (const line of lines) {
+    const parts = line.split('|');
+    if (parts.length >= 3) {
+      const name = parts[0].trim();
```
dajimenezriv, line 28: Since currently we can't activate typescript to return `undefined`, here you can do `parts.at(0).trim()`. The same in below lines.

src/backend/features/cleaner/reports/trash-bin/process-recycle-bin-output.ts
```diff
@@ -0,0 +1,58 @@
+import { logger } from '@internxt/drive-desktop-core/build/backend';
+import { CleanableItem, CleanerSection } from '@internxt/drive-desktop-core/build/backend/features/cleaner/types/cleaner.types';
+
+export function processRecycleBinOutput(stdout: string, stderr: string): CleanerSection {
```
dajimenezriv, line 4: You can use object props: https://github.com/internxt/drive-desktop-core/blob/master/CONTRIBUTING.md#use-object-props-instead-of-multiple-props

src/backend/features/cleaner/reports/trash-bin/process-recycle-bin-output.ts
```diff
@@ -0,0 +1,58 @@
+import { logger } from '@internxt/drive-desktop-core/build/backend';
```
dajimenezriv, line 1: You can write the props for logs in different lines instead of concatenating an string: https://github.com/internxt/drive-desktop-core/blob/master/CONTRIBUTING.md#logger

src/backend/features/cleaner/reports/trash-bin/generate-recycle-bin-report.ts
```diff
@@ -0,0 +1,38 @@
+import { logger } from '@internxt/drive-desktop-core/build/backend';
```
dajimenezriv, line 1: I've split the `spawn` into it's own file so we can reuse it and it's also easier to test. Also, we need to add some tests for this function and the `process-recycle-bin-output.ts` file

PR #1062, opened 2025-10-13
Add sqlite to logs zip

dajimenezriv-internxt:
## What

Create a zip file that contains the logs, the sqlite and the lokijs db to be sent to customer support.

https://inxt.atlassian.net/browse/PB-5086

### DIFF
--- a/src/apps/shared/logger/services/open-logs.test.ts
+++ b/src/apps/shared/logger/services/open-logs.test.ts
@@ -8,12 +8,15 @@ import { INTERNXT_LOGS } from '@/core/utils/utils';
 import archiver from 'archiver';
 import { mockDeep } from 'vitest-mock-extended';
 import { EventEmitter } from 'node:stream';
+import { pipeline } from 'node:stream/promises';
 
 vi.mock(import('node:fs/promises'));
 vi.mock(import('node:fs'));
+vi.mock(import('node:stream/promises'));
 vi.mock(import('archiver'));
 
 describe('open-logs', () => {
+  vi.mocked(pipeline);
   const readdirMock = vi.mocked(readdir);
   const createWriteStreamMock = vi.mocked(createWriteStream);
   const archiverMock = vi.mocked(archiver);
@@ -43,18 +46,6 @@ describe('open-logs', () => {
   it('should catch global errors', async () => {
     // Given
     readdirMock.mockRejectedValue(new Error());
-    // When
-    await openLogs();
-    // Then
-    call(loggerMock.error).toMatchObject({ msg: 'Error creating logs zip' });
-  });
-
-  it('should catch stream errors', async () => {
-    // Given
-    archive.finalize.mockImplementation(() => {
-      writeStream.emit('error');
-      return Promise.resolve();
-    });
     // When
     await openLogs();
     // Then
--- a/src/apps/shared/logger/services/open-logs.ts
+++ b/src/apps/shared/logger/services/open-logs.ts
@@ -6,22 +6,19 @@ import { shell } from 'electron';
 import { INTERNXT_LOGS } from '@/core/utils/utils';
 import { createWriteStream } from 'node:fs';
 import archiver from 'archiver';
+import { pipeline } from 'node:stream/promises';
 
 export async function openLogs() {
   logger.debug({ msg: 'Open logs' });
+
+  void shell.openPath(PATHS.LOGS);
 
   try {
     const output = join(PATHS.LOGS, INTERNXT_LOGS);
     const writeStream = createWriteStream(output);
     const archive = archiver('zip', { zlib: { level: 9 } });
 
-    archive.pipe(writeStream);
-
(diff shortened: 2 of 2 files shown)


src/apps/main/preload.js
```diff
@@ -110,9 +110,6 @@ var api = {
   openProcessIssuesWindow() {
```
dajimenezriv, line 110: These changes are autogenerated based on the below file `preload.ts`.

src/apps/shared/logger/services/open-logs.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  it('should catch file errors', async () => {
+    // Given
+    readFileMock.mockRejectedValueOnce(new Error());
+    // When
+    await openLogs();
+    // Then
+    call(writeFileMock).toStrictEqual(`/mock/logs/internxt-drive/logs/${INTERNXT_LOGS}`);
+    call(loggerMock.error).toMatchObject({ msg: 'Error adding log file to zip' });
```
sg-gs, line 63: It would be great to assert the variable containing the message from the source code against this one, so in case that changes, this one is still valid, improving the maintainability

src/apps/shared/logger/services/open-logs.ts
```diff
@@ -0,0 +1,41 @@
...
+    const logFiles = await readdir(PATHS.LOGS);
+    const logPaths = logFiles.filter((fileName) => fileName.endsWith('.log')).map((fileName) => join(PATHS.LOGS, fileName));
+
+    const paths = [...logPaths, PATHS.SQLITE_DB, PATHS.LOKIJS_DB];
+
+    const promises = paths.map(async (path) => {
+      try {
+        const data = await readFile(path);
+        const fileName = basename(path);
+        zip.file(fileName, data);
```
sg-gs: Does JSZip accept read streams? Thus you can remove the async loop and just use a for

src/core/electron/paths.ts
```diff
@@ -1,5 +1,5 @@
 import { app } from 'electron';
-import { join } from 'node:path';
+import { join } from 'node:path/posix';
```
sg-gs, line 2: This enforces cross-platform '/' usage, as in Windows is '\' natively, is this intentional? And if so, I assume the any '\' has been removed from the code and this is retroactively working for already active apps, correct?

dajimenezriv: Yes, it's intentional and I checked and it seems that nothing is broken. The only part that we must enforce with "\\" is when we send a path to C++ and we already have a function that parses all paths to windows before reaching C++. Other than that the point is to move to posix to match with `linux` and `drive-desktop-core`.

src/apps/shared/logger/services/open-logs.test.ts
dajimenezriv: I don't understand this comment. What variable do you mean? The one for the log? It's hardcoded in the code, so there is no variable. If you mean about INTERNXT_LOGS, since it contains the version of the app we "cannot" hardcode it in the test.

dajimenezriv: Do you mean to use like variables for the log messages?

src/apps/shared/logger/services/open-logs.ts
dajimenezriv: Done, changed to `archiver` that allows streams.

src/apps/shared/logger/services/open-logs.test.ts
sg-gs: In this case, is a bit problematic, as you are not propagating it probably. If it was an error, just asserting the `instanceof` would have been enough. But as this is the logger, using a variable could be an overkill on the code readability.

src/apps/shared/logger/services/open-logs.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export async function openLogs() {
+  logger.debug({ msg: 'Open logs' });
+
+  try {
+    const output = join(PATHS.LOGS, INTERNXT_LOGS);
+    const writeStream = createWriteStream(output);
+    const archive = archiver('zip', { zlib: { level: 9 } });
+
+    archive.pipe(writeStream);
```
sg-gs: If this is the native pipe method, is preferable to use [pipeline](https://nodejs.org/api/stream.html#streampipelinesource-transforms-destination-options), which is awaitable and handles piped errors automatically. [Pipe](https://nodejs.org/api/stream.html#readablepipedestination-options) requires to add an error event handler so you will have a weird mix of try-catch, async code and events.

src/apps/shared/logger/services/open-logs.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    const completionPromise = new Promise<void>((resolve, reject) => {
```
sg-gs: Yep, here it is the issue

PR #1063, opened 2025-10-13
Typed translations

dajimenezriv-internxt:
## What

Currently the way to do translations in react is using the libraries `i18next` and `i18next-react`. However these libraries have advantages and disadvantages.

Advantages:
- wide community support.
- lazy loading of translations.
- pluralization (we are not using it, but it can be used).
- typed translations with plugin libraries.
- we can add args to the translations and type them with this plugin libraries.

Disadvantages:
- i18next weights 787kB (https://packagephobia.com/result?p=i18next) and i18next-react 18kB.
- they arey complex to implement.
- not typed translations by default (and the plugins are more complex to implement).
- can't ensure the same key is written in al languages (currently `en.json` has more translations: en.json > es.json > fr.json).

A new way of doing translations is proposed. Using zustand as an store to save the language and typescript files to store the translations.

Advantages:
- simple.
- typed translations.
- the same key must be defined in all languages.
- zustand weights 90.3kb (https://packagephobia.com/result?p=zustand) and can be used to other purposes.

Disadvantages:
- we can add args to the translations but they are not typed.
- manual pluralization.
- no lazy loading (but the weight of the i18next library is bigger than the translations by themselves).

### DIFF
--- a/src/apps/renderer/hooks/useConfig.tsx
+++ b/src/apps/renderer/hooks/useConfig.tsx
@@ -1,17 +1,6 @@
 import { useEffect, useState } from 'react';
 
-import { StoredValues } from '../../main/config/service';
 import { DEFAULT_THEME, ThemeData } from '@/apps/shared/types/Theme';
-
-export function useConfig(key: StoredValues) {
-  const [value, setValue] = useState<StoredValues | undefined>(undefined);
-
-  useEffect(() => {
-    void window.electron.getConfigKey(key).then(setValue);
-  }, []);
-
-  return value;
-}
 
 export function useTheme() {
   const [value, setValue] = useState<ThemeData>({ configTheme: DEFAULT_THEME, theme: DEFAULT_THEME });
--- a/src/apps/renderer/localize/i18n.types.ts
+++ b/src/apps/renderer/localize/i18n.types.ts
@@ -4,8 +4,12 @@ export type Language = 'es' | 'en' | 'fr';
 export type Translation = typeof en;
 export type Translations = Record<Language, Translation>;
 
+/**
+ * v2.6.0 Daniel Jiménez
+ * This type generates all possible paths from the translations.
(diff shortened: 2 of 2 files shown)


src/apps/renderer/localize/languages/en.ts
```diff
@@ -0,0 +1,41 @@
...
+  issues: {
+    errors: {
+      ABORTED: 'Aborted',
+      CREATE_FOLDER_FAILED: 'Failed to create folder',
+      DELETE_ERROR: 'Cannot delete item',
+      FILE_MODIFIED: 'File modified while uploading',
+      FILE_SIZE_TOO_BIG: 'File size too big (max 20GB)',
+      FOLDER_ACCESS_DENIED: 'The app does not have permission to access this folder',
+      FOLDER_DOES_NOT_EXIST: 'Folder does not exist',
+      INVALID_WINDOWS_NAME: String.raw`Windows does not allow names that include the characters \ / : * ? " < > |`,
```
dajimenezriv, line 31: This `String.raw` thing is because of sonarcloud: https://sonarcloud.io/organizations/internxt/rules?open=typescript%3AS7780&rule_key=typescript%3AS7780

src/apps/renderer/localize/use-i18n.ts
```diff
@@ -0,0 +1,40 @@
...
+    [language],
+  );
+
+  return { t, language };
+}
+
+// export function t(path: TranslationPath, ...args: string[]) {
+//   const language = i18nStore.getState().language;
+//   return getI18nValue(language, path, ...args);
+// }
```
egalvis27, line 40: Is it necessary to leave this here?

dajimenezriv: We need to use this function for translating things that are outside react components. However, currently is not used so it will break the `find-deadcode` pipeline.

PR #1064, opened 2025-10-13
Simplify logout

dajimenezriv-internxt:


src/apps/main/remote-sync/RemoteSyncManager.ts
```diff
@@ -9,7 +9,6 @@ import { Config } from '@/apps/sync-engine/config';
 
 export class RemoteSyncManager {
   status: RemoteSyncStatus = 'IDLE';
-  totalFilesUnsynced: string[] = [];
```
dajimenezriv, line 12: It was always empty

src/apps/renderer/pages/Widget/Logout.tsx
```diff
@@ -1,82 +1,28 @@
-// ModalLogout.tsx
```
dajimenezriv, line 1: Previously we had a warning message to show to the user when logout if we had some files that were unsync. As said before, now this variable is always empty (I'm not even sure if it worked anytime). Anyway, this was because preivously we were deleting the root folder when the user was logging out. Now, since we don't delete it we don't need this modal, the user can logout and the files will still be there.

src/apps/renderer/pages/Widget/Logout.tsx
```diff
@@ -1,82 +1,28 @@
...
-  const [isSyncPending, setIsSyncPending] = useState<boolean>(false);
-  const [isReady, setIsReady] = useState<boolean>(false);
-
-  useEffect(() => {
-    //TODO: Implement checkSyncPending function to check if sync is pending
-    const checkSyncPending = async (): Promise<boolean> => {
-      const syncPending = syncStatus === 'RUNNING';
-      const wasSyncing = await window.electron.getRecentlywasSyncing();
-
-      const existFileUnsync = await window.electron.getUnsycFileInSyncEngine();
```
dajimenezriv, line 28: As said, this array is always empty.

src/apps/renderer/pages/Widget/SyncAction.tsx
```diff
@@ -24,50 +22,44 @@ export default function SyncAction(props: { syncStatus: SyncStatus }) {
...
+          </div>
+          <span className="truncate">{translate('widget.footer.action-description.updated')}</span>
+        </>
+      );
+    }
+  }
+
   return (
     <div className="flex h-11 shrink-0 items-center space-x-2.5 border-t border-gray-10 px-2.5 text-sm font-medium text-gray-100 dark:border-gray-5">
-      <div className="flex flex-1 items-center space-x-2.5 truncate px-1.5">
```
dajimenezriv, line 29: Exact same code but extracted into a function.

src/apps/renderer/pages/Widget/SyncFailed.tsx
```diff
@@ -1,14 +1,16 @@
...
 
-interface SyncFailedProps {
-  onRetrySync: () => void;
-}
-
-export const SyncFailed: React.FC<SyncFailedProps> = (props) => {
+export function SyncFailed() {
   const { translate } = useTranslationContext();
+
+  function handleRetrySync() {
```
dajimenezriv: This was sent as prop but it's not necessary, we can create the function here.

src/apps/renderer/pages/Widget/SyncFailed.tsx
```diff
@@ -1,14 +1,16 @@
-import React from 'react';
 import Button from '../../components/Button';
 import { useTranslationContext } from '../../context/LocalContext';
 import { CloudSlash } from '@phosphor-icons/react';
 
-interface SyncFailedProps {
-  onRetrySync: () => void;
+function handleRetrySync() {
```
dajimenezriv, line 5: This was sent as prop but it's not necessary, we can create the function here.

PR #1071, opened 2025-10-14
Use new onboading images

egalvis27:
Update config files to use new images in migration and onboarding slides


src/apps/renderer/pages/Onboarding/config.tsx
```diff
@@ -49,6 +49,14 @@ export const SLIDES: OnboardingSlide[] = [
       );
     },
     image: () => {
+      const WindowsFinderImage = () => {
```
dajimenezriv: Both snippets do the same, why not extract them into a function?

src/apps/renderer/pages/Onboarding/helpers.tsx
```diff
@@ -118,6 +121,10 @@ export const getAntivirusImageSvg = (language: string, theme: 'light' | 'dark')
   return lang[theme];
 };
 
+export const fileExplorerImageSvg = (theme: 'light' | 'dark') => {
```
dajimenezriv: You can use `function` instead of `const` for functions: https://github.com/internxt/drive-desktop-core/blob/master/CONTRIBUTING.md#use-function-instead-of-arrow-functions-by-default

PR #1077, opened 2025-10-16
Improve theme listener

dajimenezriv-internxt:


src/apps/main/preload/ipc-main.ts
```diff
@@ -6,19 +6,19 @@ import { calculateUsage } from '../usage/service';
 import { getLastBackupProgress } from '../background-processes/backups/BackupsProcessTracker/BackupsProcessTracker';
```
dajimenezriv, line 6: This changes and the ones from the file below are to only return promises to the frontend. Why? Because previously the function that do not return a promise were handled using `on` and `send` and that cannot be awaited in the frontend. Now, since all of them become promises when we call them from the frontend we are going to make all of them explicitily promises so they can be awaited in the frontend.

src/apps/main/theme/theme.ts
```diff
@@ -0,0 +1,31 @@
+import { nativeTheme } from 'electron';
```
dajimenezriv, line 1: This is your previous `service.ts`, however there are already many files named `service.ts` so better to call it like the name of the module.

PR #1082, opened 2025-10-17
Improve migrate old sync root

dajimenezriv-internxt:
## What

1. We have removed from the `electron-store` the properties `lastSync` (we already control this using the checkpoint from LokiJS), `deviceId` (we use the `deviceUuid`), `clientId`, `virtualdriveWindowsLetter` and `dataIntegrityMaintenance` since they were never used in the code.

2. We have refactored the old way of migrating from the old sync root folder (C:/Users/user/InternxtDrive) to the new sync root folder (C:/Users/user/InternxtDrive - user.uuid). Now we do three steps:
- Restore the user config when login.
- Check if `syncRoot` from `electron-store` is the equal to `OLD_SYNC_ROOT`.
- Call `migrateOldSyncRoot` in case they are the same.

### DIFF
--- a/src/apps/main/event-bus.ts
+++ b/src/apps/main/event-bus.ts
@@ -4,12 +4,12 @@ import { BroadcastToWindows } from './windows/broadcast-to-windows';
 
 class EventBus extends EventEmitter {}
 
-interface Events {
+type Events = {
   USER_LOGGED_IN: () => void;
   USER_LOGGED_OUT: () => void;
   ANTIVIRUS_SCAN_PROGRESS: (progress: ProgressData & { done?: boolean }) => void;
   BROADCAST_TO_WINDOWS: (_: BroadcastToWindows) => void;
-}
+};
 
 declare interface EventBus {
   on<U extends keyof Events>(event: U, listener: Events[U]): this;
--- a/src/backend/features/auth/services/utils/save-config.test.ts
+++ b/src/backend/features/auth/services/utils/save-config.test.ts
@@ -122,7 +122,6 @@ describe('saveConfig', () => {
 
   it('should retrieve all fields specified in fieldsToSave constant', () => {
     getUserMock.mockReturnValue({ uuid: 'validation-user-uuid' });
-    const allRequiredFields = ['backupsEnabled', 'backupInterval', 'lastBackup', 'syncRoot', 'deviceUuid', 'backupList'];
 
     configGetMock.mockImplementation((key: string) => {
       if (key === 'savedConfigs') return {};
@@ -131,9 +130,9 @@ describe('saveConfig', () => {
 
     saveConfig();
 
-    allRequiredFields.forEach((field) => {
+    fieldsToSave.forEach((field) => {
       expect(configGetMock).toHaveBeenCalledWith(field);
     });
-    expect(configGetMock).toHaveBeenCalledTimes(allRequiredFields.length + 1);
+    expect(configGetMock).toHaveBeenCalledTimes(fieldsToSave.length + 1);
   });
 });


src/apps/main/virtual-root-folder/service.ts
```diff
@@ -1,102 +1,30 @@
...
-      msg: 'New root virtual drive',
-      newRoot,
-    });
-    return newRoot;
-  }
-
-  return current;
-}
-
-export function setupRootFolder(user: User): void {
```
dajimenezriv, line 52: This code is the same as we have now in `migrate-old-sync-root.ts` but the new one is a bit simplier.

src/apps/main/virtual-root-folder/service.ts
```diff
@@ -1,102 +1,30 @@
...
+import { createAbsolutePath } from '@/context/local/localFile/infrastructure/AbsolutePath';
+import { migrateOldSyncRoot, OLD_SYNC_ROOT } from './migrate-old-sync-root';
 
 function setSyncRoot(pathname: string): void {
   const pathNameWithSepInTheEnd = pathname[pathname.length - 1] === path.sep ? pathname : pathname + path.sep;
 
   configStore.set('syncRoot', pathNameWithSepInTheEnd);
 }
 
-export function getRootVirtualDrive(): string {
```
dajimenezriv, line 21: Previously we were checking if the current sync root folder contained the user.uuid, in case it doesn't we assumed that it was the old sync root folder and we needed to migrate it using `setupRootFolder`. Now what we do is compare the current one with the old sync root and if they are the same it means that we need to migrate to the new format.

src/apps/main/virtual-root-folder/service.ts
```diff
@@ -105,7 +33,6 @@ export async function chooseSyncRootWithDialog(): Promise<string | null> {
     const chosenPath = result.filePaths[0];
 
     setSyncRoot(chosenPath);
-    eventBus.emit('SYNC_ROOT_CHANGED', chosenPath);
```
dajimenezriv, line 108: This event doesn't exist.

src/apps/main/windows/widget.ts
```diff
@@ -57,7 +56,6 @@ const createWidget = async () => {
   });
 
   await widgetLoaded;
-  eventBus.emit('WIDGET_IS_READY');
```
dajimenezriv, line 60: This event didn't exist.

src/apps/main/config.ts
```diff
@@ -3,63 +3,20 @@ import { AppStore } from '@/core/electron/store/app-store.interface';
 import { defaults } from '@/core/electron/store/defaults';
 
 const schema: Schema<AppStore> = {
```
dajimenezriv, line 5: We have removed the properties `lastSync` (we already control this using the checkpoint from LokiJS), `deviceId` (we use the `deviceUuid`), `clientId`, `virtualdriveWindowsLetter` and `dataIntegrityMaintenance` since they were never used in the code.

src/apps/main/event-bus.ts
```diff
@@ -5,18 +5,8 @@ import { BroadcastToWindows } from './windows/broadcast-to-windows';
 class EventBus extends EventEmitter {}
 
 interface Events {
```
egalvis27: Maybe this would be a good time to switch it to Type?

src/backend/features/auth/services/utils/save-config.test.ts
```diff
@@ -128,16 +122,7 @@ describe('saveConfig', () => {
...
-      'backupsEnabled',
-      'backupInterval',
-      'lastBackup',
-      'syncRoot',
-      'lastSync',
-      'deviceId',
-      'deviceUuid',
-      'backupList',
-    ];
+    const allRequiredFields = ['backupsEnabled', 'backupInterval', 'lastBackup', 'syncRoot', 'deviceUuid', 'backupList'];
```
egalvis27: This same array is already being exported in fields-to-save. Couldn’t you import it from there instead?

PR #1083, opened 2025-10-17
Create migrate old sync root

dajimenezriv-internxt:
## What

This is the first PR of two PRs to refactor the old migrate sync root so it can handle also the option to change the sync root folder. Previously we had the sync root folder `C:/Users/dajim/InternxtDrive` and when the user was logging out we deleted it. However if the files were not synced the user was losing all files. So we decided to not delete the sync root folder but now we needed to add the `userId` to the sync root folder to handle multiple accounts in the same device. Now we have extracted that into it's own function.

Also, we had the function `createRelativePath` to ensure that all relatives paths had the same format. We are doing the same with `createAbsolutePath`.

### DIFF
--- a/src/apps/main/virtual-root-folder/migrate-old-sync-root.ts
+++ b/src/apps/main/virtual-root-folder/migrate-old-sync-root.ts
@@ -5,7 +5,7 @@ import { existsSync, renameSync } from 'node:fs';
 import { User } from '../types';
 import { configStore } from '../config';
 
-export const OLD_SYNC_ROOT = createAbsolutePath(PATHS.HOME_FOLDER_PATH, 'InternxtDrive');
+const OLD_SYNC_ROOT = createAbsolutePath(PATHS.HOME_FOLDER_PATH, 'InternxtDrive');
 
 type Props = { user: User };
 


src/apps/main/virtual-root-folder/migrate-old-sync-root.ts
```diff
@@ -0,0 +1,41 @@
...
+   * login called "InternxtDrive - {user.uuid}."
+   * So, we need to rename "InternxtDrive" to "InternxtDrive - {user.uuid}".
+   */
+
+  try {
+    if (existsSync(newSyncRoot)) {
+      logger.debug({ msg: 'New sync root already exists, skiping' });
+    } else if (existsSync(OLD_SYNC_ROOT)) {
+      logger.debug({ msg: 'Migrate old sync root' });
+      renameSync(OLD_SYNC_ROOT, newSyncRoot);
```
dajimenezriv, line 31: If `newSyncRoot` does not exist but `OLD_SYNC_ROOT` exists then mv the old one name to the new one name (containing the `userId`).

src/context/local/localFile/infrastructure/AbsolutePath.ts
```diff
@@ -1,22 +1,31 @@
 import { posix, win32 } from 'node:path';
-import { Brand } from '../../../shared/domain/Brand';
+import { AbsolutePath as CoreAbsolutePath, RelativePath as CoreRelativePath } from '@internxt/drive-desktop-core/build/backend';
```
dajimenezriv, line 2: We use branded types `RelativePath` and `AbsolutePath` from `drive-desktop-core`.

src/apps/main/config.ts
```diff
@@ -62,6 +62,5 @@ const schema: Schema<AppStore> = {
   },
 } as const;
 
-const configStore = new Store({ schema, defaults });
-
+export const configStore = new Store({ schema, defaults });
 export default configStore;
```
egalvis27, line 66: Is it correct to export the ConfigStore twice?

src/apps/main/virtual-root-folder/migrate-old-sync-root.ts
```diff
@@ -0,0 +1,41 @@
...
+import { configStore } from '../config';
+
+export const OLD_SYNC_ROOT = createAbsolutePath(PATHS.HOME_FOLDER_PATH, 'InternxtDrive');
+
+type Props = { user: User };
+
+export function migrateOldSyncRoot({ user }: Props) {
+  const newSyncRoot = createAbsolutePath(PATHS.HOME_FOLDER_PATH, `InternxtDrive - ${user.uuid}`);
+
+  logger.debug({ msg: 'Check migrate old sync root', oldSyncRoot: OLD_SYNC_ROOT, newSyncRoot });
```
egalvis27: There is no tag for this logger?

src/apps/main/config.ts
dajimenezriv: The default export is the legacy one, so technically we should only have the `export const configStore`. For now it's to be able to use `configStore` without the default until we fully remove the default export.

PR #1084, opened 2025-10-17
Refactor CleanerProvider to use useMemo for context value optimization

egalvis27:


src/apps/renderer/context/cleaner-context.tsx
```diff
@@ -74,20 +74,20 @@ export function CleanerProvider({ children }: Props) {
...
-        diskSpace,
-        generateReport,
-        startCleanup,
-        stopCleanup,
-        setInitialCleaningState,
-        sectionKeys: cleanerSectionKeys,
-      }}>
-      {children}
-    </CleanerContext.Provider>
+  const value = useMemo(
```
dajimenezriv: Why this is necessary?

PR #1087, opened 2025-10-18
Move i18n to object args

dajimenezriv-internxt:


src/apps/main/device/handlers.ts
```diff
@@ -10,7 +10,6 @@ import {
   getOrCreateDevice,
```
dajimenezriv, line 10: This PR migrates some `useI18n` and also deletes a screen in onboarding that was hidden about backups. Basically it was a screen for selecting the backup folders in the onboarding, however it was legacy.

src/apps/renderer/localize/use-i18n.ts
```diff
@@ -7,15 +7,17 @@ import { fr } from './languages/fr';
 
 const translations: Translations = { en, es, fr };
 
-export function getI18nValue(language: Language, path: TranslationPath, ...args: string[]) {
```
dajimenezriv, line 10: Make args a dictionary to keep the same functionality as with `i18next`

src/apps/renderer/pages/Onboarding/index.tsx
```diff
@@ -97,22 +59,12 @@ export default function Onboarding() {
...
-          onSetupBackups={setupBackups}
           onGoNextSlide={nextSlide}
           onSkipOnboarding={finish}
           currentSlide={slideIndex}
           totalSlides={totalSlides}
         />
       </div>
-      <div
-        className={`backups-modal-overlay w- absolute h-full w-full py-11 transition-all duration-300 ease-in-out ${
-          backupsModalOpen ? 'pointer-events-auto translate-y-0 opacity-100' : 'pointer-events-none translate-y-2 opacity-0'
```
dajimenezriv, line 110: This was always hidden.

PR #1091, opened 2025-10-21
Update package version

dajimenezriv-internxt:


.github/workflows/check-pr-size.yml
```diff
@@ -7,7 +7,8 @@ on:
 jobs:
   check_pr_size:
     # Ignore PRs with name merge-X-X-X-release
-    if: false
+    if: |
+      !(startsWith(github.head_ref, 'merge-') && endsWith(github.head_ref, '-release') && github.base_ref == 'main')
```
dajimenezriv, line 11: We disabled this for the last PR of translations since deleting all translation files were so many PRs, so we need to enable now.

sg-gs (approved):

PR #1092, opened 2025-10-21
Feat: added foundation functions for the new identification mechanism

AlexisMora:
Here we have foundation methods that will be used in the upcoming methods for the new identification mechanism for the devices

### DIFF
--- a/src/backend/features/device/services/decrypt-device-name.test.ts
+++ b/src/backend/features/device/services/decrypt-device-name.test.ts
@@ -1,3 +1,4 @@
+import { call, calls, mockProps } from '@/tests/vitest/utils.helper.test';
 import { decryptDeviceName } from './decrypt-device-name';
 import { aes } from '@internxt/lib';
 
@@ -6,21 +7,16 @@ vi.mock('@internxt/lib');
 describe('decrypt-device-name', () => {
   const aesDecryptMock = vi.mocked(aes.decrypt);
   const decryptedName = 'decrypted-device-name';
-  const mockDevice = {
-    id: 1,
-    uuid: 'device-uuid-123',
-    name: 'encrypted-device-name',
-    bucket: 'test-bucket',
-    removed: false,
-    hasBackups: true,
-    lastBackupAt: '2024-01-01T00:00:00Z',
-  };
+  const encryptedName = 'encrypted-device-name';
+  const bucket = 'bucket';
+
+  const props = mockProps<typeof decryptDeviceName>({ name: encryptedName, bucket });
 
   it('should decrypt device name using bucket key', () => {
     aesDecryptMock.mockReturnValue(decryptedName);
-    const device = decryptDeviceName(mockDevice);
-    expect(device).toStrictEqual({ ...mockDevice, name: decryptedName });
-    expect(aesDecryptMock).toHaveBeenCalledWith(mockDevice.name, `${process.env.NEW_CRYPTO_KEY}-${mockDevice.bucket}`);
+    const device = decryptDeviceName(props);
+    expect(device).toMatchObject({ name: decryptedName, bucket });
+    call(aesDecryptMock).toStrictEqual([encryptedName, `${process.env.NEW_CRYPTO_KEY}-${bucket}`]);
   });
 
   it('should fallback to null bucket key when decryption with bucket fails', () => {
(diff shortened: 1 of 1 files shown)


src/core/utils/exec-async.ts
```diff
@@ -0,0 +1,3 @@
+import { exec } from 'node:child_process';
```
AlexisMora, line 1: I created this file basically because it was giving me a headache to mock it inside get machine guid. I also saw that this was already done in `is-windows-defender-available.ts`

src/backend/features/device/utils/map-device-as-folder-to-device.ts
```diff
@@ -0,0 +1,19 @@
+import { components } from '@/apps/shared/HttpClient/schema';
+import { Device } from '../../../../apps/main/device/service';
+
+/**
+ * Maps a DeviceAsFolder from the API to the internal Device type
+ * @param deviceAsFolder - The device-as-folder object from the API
+ * @returns Device - The mapped device object
+ */
+export function mapDeviceAsFolderToDevice(deviceAsFolder: components['schemas']['DeviceAsFolder']): Device {
```
AlexisMora: I did this mapper to map the responses of the infra calls that are going to be used for the new device identification mechanism, so that they are mapped our our domain object `Device` since some responses come with extra data and object we dont currenyly need

src/infra/device/get-machine-guid.ts
```diff
@@ -0,0 +1,25 @@
...
+    public readonly code: 'NON_EXISTS' | 'NO_ACCESS' | 'UNKNOWN',
+    cause?: unknown,
+  ) {
+    super(code, { cause });
+  }
+}
+
+export async function getMachineGuid() {
+  try {
+    const { stdout } = await execAsync(String.raw`reg query HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography /v MachineGuid`, {
```
AlexisMora, line 14: I did it async just to be sure that this is not going to block anything.
`HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography /v MachineGuid` is unique enough so that if the user reinstall the app it will give the same id, but if the user reinstalls the whole os, this command will give a new id so we have to keep in mind that is unique but up to a certain level.
Regarding permissions this command should be a pretty basic

src/infra/drive-server-wip/services/backup/create-backup-folder.ts
```diff
@@ -0,0 +1,28 @@
+import { client } from '@/apps/shared/HttpClient/client';
+import { getRequestKey } from '../../in/get-in-flight-request';
+import { clientWrapper } from '../../in/client-wrapper.service';
+
+type Props = { parentFolderUuid: string; plainName: string };
+
+export async function createBackupFolder(context: Props) {
```
AlexisMora: I created this request because i saw that in `device/service.ts` `postBackup` it was still being used like this ` await client.POST('/folders', {...[]`

src/infra/drive-server-wip/services/backup/create-backup-folder.ts
```diff
@@ -0,0 +1,28 @@
...
+    loggerBody: {
+      msg: 'Create backup folder request',
+      context,
+      attributes: {
+        method,
+        endpoint,
+      },
+    },
+  });
+  return { data, error };
```
AlexisMora: Regarding the response, should this request also propagate the 409 response?

src/infra/drive-server-wip/services/backup/create-backup-folder.ts
```diff
@@ -0,0 +1,28 @@
...
+    client.POST(endpoint, {
+      body: { ...context },
+    });
+  const { data, error } = await clientWrapper({
+    promiseFn,
+    key,
+    loggerBody: {
+      msg: 'Create backup folder request',
+      context,
+      attributes: {
```
dajimenezriv: Attributes was only for `open-telemetry`, it can be removed

src/infra/drive-server-wip/services/backup/create-backup-folder.ts
dajimenezriv: Can we reuse the endpoint of creating a folder?

src/backend/features/device/utils/map-device-as-folder-to-device.ts
dajimenezriv: Since `Device` contains the same fields as `DeviceAsFolder` you don't need a mapper. You can just do `as Device` when you return the data, or use `Device` in the prop and it will already do the map in types.

src/infra/drive-server-wip/services/backup/create-backup-folder.ts
AlexisMora: Yup totally, didnt see it at first

dajimenezriv-internxt (approved):

PR #1094, opened 2025-10-22
Fix language and theme bug

dajimenezriv-internxt:


src/apps/renderer/pages/Widget/InfoBanners/Banners/DiscoverBackups.tsx
```diff
@@ -6,15 +6,12 @@ import { useI18n } from '@/apps/renderer/localize/use-i18n';
 
 export function DiscoverBackups() {
   const { hasDiscovered, discover } = useDiscoverBackups();
+  const { translate } = useI18n();
```
dajimenezriv, line 9: This was causing an error because of conditional render.

src/apps/renderer/features/config/use-theme-setup.tsx
```diff
@@ -0,0 +1,15 @@
+import { useEffect } from 'react';
```
dajimenezriv, line 1: This functionality was used before in the file below `getConfig`. However, it was doing a suscription to `listenToConfigKeyChange` everytime we used `useTheme`. Now we have a `useThemeSetup` that is loaded only one time and stores the value in `configStore`, the same as language.

Now, the function `useTheme` from below just reads the value from the store.

egalvis27 (approved):

PR #1102, opened 2025-10-28
Allow users to choose custom sync folder location

dajimenezriv-internxt:
## What

Original PR: https://github.com/internxt/drive-desktop/pull/1042

Allow user to change virtual drive folder. When a user changes the virtual drive, we need to stop the sync engine worker, unregister the cloud provider, wait 2s until the cloud provider is fully unregistered, migrate the sync root (move the folder) and spawn the sync engine worker again.

### DIFF
--- a/src/apps/sync-engine/tests/create-placeholder.infra.test.ts
+++ b/src/apps/sync-engine/tests/create-placeholder.infra.test.ts
@@ -60,7 +60,7 @@ describe('create-placeholder', () => {
   });
 
   afterAll(() => {
-    BindingsManager.stop({ ctx });
+    ctx.virtualDrive.disconnectSyncRoot();
     VirtualDrive.unregisterSyncRoot({ providerId });
   });
 


src/apps/main/virtual-root-folder/migrate-sync-root.ts
```diff
@@ -0,0 +1,25 @@
+import { AbsolutePath, logger } from '@internxt/drive-desktop-core/build/backend';
```
dajimenezriv, line 1: This does the same as the old `migrate-old-sync-root.ts`, however now it accepts the props of the `oldSyncRoot` and the `newSyncRoot` so it can be use also to change the sync root.

src/apps/main/virtual-root-folder/service.ts
```diff
@@ -23,25 +23,59 @@ export function getRootVirtualDrive() {
...
+
+  logger.debug({ msg: 'Choose sync root with dialog', chosenPath });
+
+  const newSyncRoot = createAbsolutePath(chosenPath, `InternxtDrive - ${user.uuid}`);
+  const oldSyncRoot = createAbsolutePath(configStore.get('syncRoot'));
+
+  if (newSyncRoot === oldSyncRoot) return;
+
+  try {
+    const worker = workers.get('');
```
dajimenezriv, line 60: We get the worker with no workspace (empty string).

src/apps/sync-engine/index.ts
```diff
@@ -22,11 +22,12 @@ async function setUp({ ctx }: { ctx: ProcessSyncContext }) {
 
   const container = buildProcessContainer({ ctx });
 
+  await BindingsManager.start({ ctx, container });
```
dajimenezriv, line 25: In case there is an error registering the sync root we want to throw before setting the ipc listener.

egalvis27 (approved):

PR #1103, opened 2025-10-29
Improve create file and folder placeholders

dajimenezriv-internxt:
## What

We have improved and simplified how we create file and folder placeholders in C++. We have also renamed the functions from `createPlaceholderFile` and `createEntry` to `createFilePlaceholder` and `createFolderPlaceholder`. This PR changes these things from C++ in drive-desktop.

### DIFF
--- a/src/apps/main/remote-sync/handlers.ts
+++ b/src/apps/main/remote-sync/handlers.ts
@@ -56,31 +56,23 @@ export const deleteFileInBatch = async (itemsIds: string[]) => {
   });
 };
 
-export async function getUpdatedRemoteItems(workspaceId: string) {
-  try {
-    const promise = Promise.all([
-      SqliteModule.FileModule.getByWorkspaceId({ workspaceId }),
-      SqliteModule.FolderModule.getByWorkspaceId({ workspaceId }),
-    ]);
+ipcMainSyncEngine.handle('FIND_EXISTING_FILES', async (_, { userUuid, workspaceId }) => {
+  const { data: files = [] } = await SqliteModule.FileModule.getByWorkspaceId({ userUuid, workspaceId });
+  return files;
+});
 
-    const [{ data: files = [] }, { data: folders = [] }] = await promise;
+ipcMainSyncEngine.handle('GET_UPDATED_REMOTE_ITEMS', async (_, { userUuid, workspaceId }) => {
+  const [{ data: files = [] }, { data: folders = [] }] = await Promise.all([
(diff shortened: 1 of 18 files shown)


src/apps/sync-engine/tests/create-placeholder.infra.test.ts
```diff
@@ -139,11 +137,11 @@ describe('create-placeholder', () => {
       { msg: 'onReady' },
       { msg: 'Create file', path: '/file.txt' },
       { tag: 'SYNC-ENGINE', msg: 'File uploaded', path: '/file.txt', contentsId: '012345678901234567890123', size: 7 },
-      { msg: 'Convert to placeholder succeeded', itemPath: '/file.txt', id: `FILE:${fileUuid}` },
+      { tag: 'SYNC-ENGINE', msg: 'Convert to placeholder succeeded', itemPath: '/file.txt', id: `FILE:${fileUuid}` },
       {
         msg: 'Change event triggered',
         path: '/file.txt',
-        pinState: 1,
+        pinState: PinState.Unspecified,
```
dajimenezriv, line 144: Since now we are not enforcing a pin state using `UpdatePinState` in C++, instead of having AlwaysLocal it will have Unspecified, however the file explorer displays it well.

src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.infra.test.ts
```diff
@@ -51,8 +50,6 @@ describe('sync-remote-changes-to-local', () => {
 
     await writeFile(filePath, 'content');
     virtualDrive.convertToPlaceholder({ itemPath: filePath, id: 'FILE:uuid' });
-    let status = virtualDrive.getPlaceholderState({ path: filePath });
```
dajimenezriv, line 54: Since now we are not enforcing a pin state using `UpdatePinState` in C++, instead of having AlwaysLocal it will have Unspecified, however the file explorer displays it well.

src/node-win/virtual-drive.ts
```diff
@@ -139,90 +86,76 @@ export class VirtualDrive {
...
-      });
-    } else {
-      logger.error({
-        msg: 'Convert to placeholder failed',
-        itemPath,
-        id,
-        error: result.errorMessage,
-      });
+    try {
+      this.addon.convertToPlaceholder({ path: this.fixPath(itemPath), id });
```
dajimenezriv, line 154: Now it throws an exception in case of error and it doesn't return anything

PR #1105, opened 2025-10-29
Sqlite filter by user uuid

dajimenezriv-internxt:
## What

Filter in sqlite also by user uuid. We have a bug that for multi accounts we were retrieving also all items from another accounts.

### DIFF
--- a/src/backend/features/sync/recovery-sync/common/get-deleted-items.ts
+++ b/src/backend/features/sync/recovery-sync/common/get-deleted-items.ts
@@ -7,7 +7,7 @@ type Props = FileProps | FolderProps;
 export function getDeletedItems(props: FolderProps): SimpleDriveFolder[];
 export function getDeletedItems(props: FileProps): SimpleDriveFile[];
 
-export function getDeletedItems({ ctx, remotes, locals }: Props) {
+export function getDeletedItems({ ctx, type, remotes, locals }: Props) {
   const remotesMap = new Map(remotes.map((item) => [item.uuid, item]));
 
   const itemsToDelete = locals.filter((local) => {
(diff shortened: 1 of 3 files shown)


src/apps/main/remote-sync/handlers.ts
```diff
@@ -93,16 +85,6 @@ ipcMain.handle('UPDATE_FIXED_FILES', async (_, inputData) => {
   await deleteFileInBatch(inputData.toDelete);
 });
 
-ipcMain.handle('GET_UPDATED_REMOTE_ITEMS', (_, workspaceId: string) => {
```
dajimenezriv, line 96: We move this to `ipcMainSyncEngine` since it's typed.

src/apps/main/remote-sync/handlers.ts
```diff
@@ -56,31 +56,23 @@ export const deleteFileInBatch = async (itemsIds: string[]) => {
   });
 };
 
-export async function getUpdatedRemoteItems(workspaceId: string) {
-  try {
-    const promise = Promise.all([
```
dajimenezriv, line 61: This cannot throw an exception

sg-gs: Why? If this is a kind of controller, don't you prefer to centralise error management here?

dajimenezriv: Since it follows the pattern { data, error }. All functions that are from infra return { data, error } instead of throwing an error, so we can handle in the business layer the different types of error. In this case the functions `getByWorkspaceId` already log the error in case something happens.

sg-gs: Golang style haha

sg-gs (approved):
Is this retroactive for already errored users? 

If affected users have other users items right now, does this fix that also?

PR #1106, opened 2025-10-30
Simplify spawn sync engine worker

dajimenezriv-internxt:
## What

This PR simplifies the way in which we spawn a sync engine worker (for drive or workspace), how we schedule the sync by checkpoint and how we stop the sync engine worker. This is a PR needed to allow the change of a sync root folder, since we need to kill the current sync engine worker and start a new one with the new sync root folder. Also, it solves an inefficiency that could lead to multiple syncs by checkpoint happening at the same time.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@internxt/drive-desktop-core": "0.1.8",
         "@internxt/inxt-js": "2.2.9",
         "@internxt/lib": "1.3.1",
-        "@internxt/node-win": "1.0.25",
+        "@internxt/node-win": "1.0.26",
         "@internxt/scan": "1.0.7",
         "@internxt/sdk": "1.11.10",
         "@phosphor-icons/react": "2.0.9",
@@ -3317,9 +3317,9 @@
       "license": "MIT"
     },
     "node_modules/@internxt/node-win": {
-      "version": "1.0.25",
-      "resolved": "https://registry.npmjs.org/@internxt/node-win/-/node-win-1.0.25.tgz",
-      "integrity": "sha512-ac8Ai0SRmm5IPl4zLqmzJD8SPp4jGKdcEl9q0mrHFsCRJr5Z3809Aknn/tIpJcdxo6bGaN7QRJbnq9oSS1GZcQ==",
+      "version": "1.0.26",
+      "resolved": "https://registry.npmjs.org/@internxt/node-win/-/node-win-1.0.26.tgz",
+      "integrity": "sha512-HhXTSAbrLMxTUnGq/YYjf6oejslY79mkRYh3rrN6f43Gn5q1xEnmJ/iSC9aALo+lMw7Ap5rqp1HJIYXrll/cxA==",
       "hasInstallScript": true,
       "license": "MIT"
     },
--- a/package.json
+++ b/package.json
@@ -164,7 +164,7 @@
     "@internxt/drive-desktop-core": "0.1.8",
     "@internxt/inxt-js": "2.2.9",
     "@internxt/lib": "1.3.1",
-    "@internxt/node-win": "1.0.25",
+    "@internxt/node-win": "1.0.26",
     "@internxt/scan": "1.0.7",
     "@internxt/sdk": "1.11.10",
     "@phosphor-icons/react": "2.0.9",
--- a/src/apps/sync-engine/in/add-pending-folders.ts
+++ b/src/apps/sync-engine/in/add-pending-folders.ts
@@ -16,7 +16,7 @@ export async function addPendingFolders({ ctx, createFolders }: TProps) {
         path: absolutePath,
       });
 
-      await createFolder({ ctx, path, absolutePath });
+      await createFolder({ ctx, path });
     }),
   );
 }
--- a/src/apps/sync-engine/in/add-pending-items.ts
+++ b/src/apps/sync-engine/in/add-pending-items.ts
@@ -19,7 +19,7 @@ export async function addPendingItems({ ctx }: Props) {
     await Promise.all([
       addPendingFiles({ ctx, createFiles }),
       addPendingFolders({ ctx, createFolders }),
-      hydrateFiles.map((file) => throttleHydrate({ ctx, path: file.path })),
+      hydrateFiles.map((file) => throttleHydrate({ ctx, ...file })),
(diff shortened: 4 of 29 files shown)


src/apps/main/background-processes/sync-engine/services/schedule-sync.ts
```diff
@@ -1,16 +1,15 @@
...
-  worker.syncSchedule?.cancel(false);
-  worker.syncSchedule = nodeSchedule.scheduleJob('*/10 * * * *', async () => {
+type Props = { workspaceId: string };
+
+export function scheduleSync({ workspaceId }: Props) {
+  void updateRemoteSync({ workspaceId });
+
+  return setInterval(async () => {
     logger.debug({ tag: 'SYNC-ENGINE', msg: 'Start scheduled sync' });
-    await debouncedSynchronization();
```
dajimenezriv, line 14: Previously the `scheduleSync` was calling `debouncedSynchronization` which was a function that executed the sync for the drive and all workspaces, so technically every worker was calling the sync by checkpoint for all. This wasn't leading to any problem because the sync by checkpoint is debounced by 5s so all calls were joined into one. Now, we just run the sync by checkpoint of this drive/workspace and we schedule also the sync by checkpoint only of this drive/workspace.

src/apps/main/background-processes/sync-engine/services/spawn-sync-engine-worker.ts
```diff
@@ -2,43 +2,18 @@ import { SyncContext } from '@/apps/sync-engine/config';
 import { BrowserWindow } from 'electron';
```
dajimenezriv, line 2: Now we have simplified the way in which we handle the Map of workers so we cannot instanciate two sync engine workers for the same drive. There are only two ways of spawning a sync engine worker. When we log in (we stop all sync engine workers when we logout so we cannot have two at the same time) and when the sync engine stops working we first clear the previous worker and then start a new one.

src/apps/main/background-processes/sync-engine/services/spawn-sync-engine-worker.ts
```diff
@@ -2,43 +2,18 @@ import { SyncContext } from '@/apps/sync-engine/config';
...
 import { scheduleSync } from './schedule-sync';
 import { addRemoteSyncManager } from '@/apps/main/remote-sync/handlers';
 import { RecoverySyncModule } from '@/backend/features/sync/recovery-sync/recovery-sync.module';
+import { stopSyncEngineWorker } from './stop-sync-engine-worker';
 
 type TProps = {
   ctx: SyncContext;
 };
 
-let hasPrinted = false;
```
dajimenezriv, line 16: We move this variable inside the function since we want it to false in every spawn sync engine worker, not just for the first one.

src/apps/main/background-processes/sync-engine/services/spawn-sync-engine-worker.ts
```diff
@@ -94,18 +86,6 @@ export async function spawnSyncEngineWorker({ ctx }: TProps) {
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     const { logger, ...config } = ctx;
     browserWindow.webContents.send('SET_CONFIG', config);
-
-    monitorHealth({
```
dajimenezriv, line 98: We move this code to above the `await browserWindow.loadFile` so we start performing the sync by checkpoint instead of waiting for the process to load.

src/apps/main/background-processes/sync-engine/services/unregister-virtual-drives.ts
```diff
@@ -20,7 +20,6 @@ export function unregisterVirtualDrives({ currentProviderIds = [] }: TProps) {
 
   internxtSyncRoots.forEach((syncRoot) => {
     if (!currentProviderIds.includes(syncRoot.id)) {
-      logger.debug({ msg: 'Unregistering sync root', syncRoot });
```
dajimenezriv, line 23: We move this log inside the function that unregisters since now this function will be called also when we change the sync root folder.

src/apps/main/quit.ts
```diff
@@ -1,10 +1,8 @@
 import { app, ipcMain } from 'electron';
-import { stopAndClearAllSyncEngineWatcher } from './background-processes/sync-engine';
-import { unregisterVirtualDrives } from './background-processes/sync-engine/services/unregister-virtual-drives';
+import { stopSyncEngineWorkers } from './background-processes/sync-engine/services/stop-sync-engine-worker';
 
 export async function quitApp() {
-  await stopAndClearAllSyncEngineWatcher();
-  unregisterVirtualDrives({});
+  await stopSyncEngineWorkers();
```
dajimenezriv, line 5: We already unregister all virtual drives inside this function.

src/apps/main/background-processes/sync-engine/services/stop-sync-engine-worker.ts
```diff
@@ -0,0 +1,26 @@
...
+  worker.browserWindow.destroy();
+  workers.delete(ctx.workspaceId);
+}
+
+export async function stopSyncEngineWorkers() {
+  workers.forEach((worker) => {
+    stopSyncEngineWorker({ worker });
+  });
+
+  await sleep(2000);
```
sg-gs, line 24: Does this value respond to some specific reason ?

src/apps/main/background-processes/sync-engine/services/schedule-sync.ts
sg-gs: Why it was running before on all the drives and workspaces and now it does not?

dajimenezriv: Now each worker just perform it's own update with `updateRemoteSync`, when the worker is stopped it clears it's own interval

src/apps/main/background-processes/sync-engine/services/stop-sync-engine-worker.ts
dajimenezriv: Nop, we need to wait until the sync engine worker is stopped before trying to unregister the virtual drives, otherwise sometimes it doesn't register because it's still opened by the sync engine worker. Probably one second is enough, but wanted to be sure.

sg-gs (approved):

PR #1108, opened 2025-10-31
Add get file and folder info

dajimenezriv-internxt:
## What

This the first PR of 2 PRs that apply the changes of removing the `getFileIdentity` and joining it with `getPlaceholderState`  in C++.

This PR has the infra tests failing since I cannot make the full migration in just one PR, so this PR doesn't contain yet the new version of C++ and the infra tests for these new functions `getFileInfo` and `getFolderInfo` are not working. You can check that the tests are passing in this second PR which completes the migration: https://github.com/internxt/drive-desktop/pull/1107.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,7 +49,7 @@
         "reflect-metadata": "^0.1.13",
         "regenerator-runtime": "^0.13.9",
         "socket.io-client": "^4.4.1",
-        "typeorm": "^0.3.16",
+        "typeorm": "^0.3.26",
         "uuid": "^8.3.2",
         "zustand": "^4.5.7"
       },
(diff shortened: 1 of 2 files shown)


src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.infra.test.ts
```diff
@@ -27,16 +25,13 @@ describe('sync-remote-changes-to-local', () => {
   const rootUuid = v4();
   const providerId = `{${rootUuid.toUpperCase()}}`;
   const virtualDrive = new VirtualDrive({ loggerPath: '', providerId, rootPath });
-  const callbacks = mockDeep<Callbacks>();
 
   beforeEach(async () => {
     await virtualDrive.createSyncRootFolder();
     virtualDrive.registerSyncRoot({ providerName });
-    virtualDrive.connectSyncRoot({ callbacks });
```
dajimenezriv, line 35: This is not needed, we just need to `registerSyncRoot` and `unregisterSyncRoot`.

src/infra/node-win/services/item-identity/get-folder-info.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+}
+
+type TProps = {
+  ctx: ProcessSyncContext;
+  path: string;
+};
+
+export function getFolderInfo({ ctx, path }: TProps) {
+  if (path === '/' || path === ctx.virtualDrive.syncRootPath || `${path}\\` === ctx.virtualDrive.syncRootPath) {
```
dajimenezriv, line 22: This is the same check that we have in `get-folder-uuid.ts` since we cannot obtain the placeholder info of the root folder since it's not a placeholder.

PR #1111, opened 2025-11-01
Improve download file

dajimenezriv-internxt:
## What

This is the second of 2 PRs that implement the fetchData and cancelFetchData changes from node-win. Now:
1. We get the path of the file instead of the placeholderId from C++ when we hydrate a file.
2. We obtain the uuid of that file and then we obtain the file info from the sqlite.
3. We start the download and instead of writing it to a local file and sending the tmpPath to C++ now we just send the chunks.

### DIFF
--- a/src/node-win/types/callbacks.type.ts
+++ b/src/node-win/types/callbacks.type.ts
@@ -1,6 +1,6 @@
 export type CallbackDownload = (buffer: Buffer, offset: number) => void;
 
 export type Callbacks = {
-  fetchDataCallback: (path: string, callback: CallbackDownload) => void;
+  fetchDataCallback: (path: string, callback: CallbackDownload) => Promise<void>;
   cancelFetchDataCallback: (path: string) => void;
 };


src/context/virtual-drive/files/application/FileOverwriteContent.ts
```diff
@@ -42,36 +40,8 @@ export class FileOverwriteContent {
     void this.processErrorQueue();
   }
 
-  private async registerEvents(input: {
```
dajimenezriv, line 45: We move the `on('progress')` and `on('error')` below.

src/context/virtual-drive/contents/application/ContentsDownloader.ts
```diff
@@ -33,7 +33,8 @@ export class ContentsDownloader {
 
     logger.debug({ msg: '[Server] Stopping download 2' });
     this.downloaderIntance.forceStop();
-    void this.downloaderIntanceCB(false, '');
+    // @ts-ignore
+    this.downloaderIntanceCB(false, '');
```
dajimenezriv, line 37: This is just to allow the migration since this file won't be used anymore but we cannot delete everything because we will reach the PR size. The same for the above test, which is the test of this file.

src/apps/sync-engine/callbacks/fetchData.service.ts
```diff
@@ -1,67 +1,30 @@
...
+    if (error2) throw error2;
 
-    logger.debug({ msg: '[Fetch Data Callback] Preparing begins', tmpPath });
-
-    let finished = false;
-    let progressBuffer = 0;
-
-    try {
-      while (!finished) {
-        const result = await callback(true, tmpPath);
```
dajimenezriv, line 36: Now callback doesn't return anything and we don't need to perform any of this logic. We just obtain the uuid using `getFileInfo` from the path. Then we obtain the file from the sqlite using the uuid.

src/apps/sync-engine/download-contents/download-contents.ts
```diff
@@ -0,0 +1,62 @@
...
+      callback(chunk, offset);
+
+      offset += chunk.length;
+    }
+
+    ctx.logger.debug({ msg: 'File downloaded', path });
+
+    ipcRendererSyncEngine.send('FILE_DOWNLOADED', { path });
+  } catch (error) {
+    if (error instanceof Error && error.message !== 'The operation was aborted') {
```
dajimenezriv, line 54: When we abort the `ActionState` in inxt-js it throws an exception in the for loop of the readable. The message is this one. We want to ignore since it's not an error.

src/apps/sync-engine/download-contents/download-contents.ts
```diff
@@ -0,0 +1,62 @@
...
+        ctx.logger.debug({
+          msg: 'Last chunk received',
+          path,
+          offset,
+          chunk: chunk.length,
+          size: file.size,
+        });
+      }
+
+      callback(chunk, offset);
```
dajimenezriv, line 45: This is the callback that sends the chunk to C++.

src/apps/main/app-info/handlers.ts
sg-gs, line 1: What does this change has to do with the PR?

src/apps/main/background-processes/backups/build-file-uploader.ts
```diff
@@ -21,6 +22,7 @@ export function buildFileUploader({ bucket }: Props) {
   });
 
   const fileUploader = new EnvironmentFileUploader(environment, bucket);
+  const contentsDownloader = new InxtJs.ContentsDownloader(environment, bucket);
```
sg-gs: Why not fileDownloader?

src/apps/sync-engine/index.ts
```diff
@@ -48,13 +46,15 @@ ipcRenderer.once('SET_CONFIG', async (event, config: Config) => {
   try {
     setConfig(config);
 
-    const { fileUploader } = buildFileUploader({ bucket: config.bucket });
+    const { fileUploader, contentsDownloader } = buildFileUploader({ bucket: config.bucket });
```
sg-gs: The `buildFileUploader` seems to be building more than the `fileUploader` haha

src/apps/main/app-info/handlers.ts
dajimenezriv: This file was setting up handlers in main process so we could obtain a tmp path for intermidiate file from the sync engine process. Since they were handlers I preferred delete it now to not forget it because they won't appear in the `find-deadcode` script.

src/apps/main/background-processes/backups/build-file-uploader.ts
dajimenezriv: It was the first option, however, I think that makes more sense to rename `fileUploader` to `contentsUploader`. Why? The `fileId` of `drive-server-wip` is called `contentsId` in windows app so to make it more consistent I think that makes more sense to name all things related to the file contents with `contents` so we don't have any doubt if `fileUploader` is a `create` file or an `upload` file contents. The same when we want to update the `contents`, what would be `fileUpdater` (which currently we have in backups and it's for creating/replacing files)?

src/context/virtual-drive/files/application/FileOverwriteContent.ts
```diff
@@ -104,11 +72,18 @@ export class FileOverwriteContent {
...
     for (const file of files) {
       if (filesWithContentLocally[file.path]) {
-        const downloader = downloaderManger.downloader();
-        void this.registerEvents({ downloader, file, callback: asynchronousFixingOfDangledFiles });
-
         logger.debug({ msg: 'Trying to download file ', uuid: file.uuid, name: file.name });
-        await downloader.download(file);
+
+        const { error } = await ctx.contentsDownloader.download({
+          onProgress: () => ctx.contentsDownloader.forceStop(),
```
sg-gs, line 78: Why the progress callback is forcing the stop of the download?

dajimenezriv: You can see that it was the previous behaviour. Basically it is for dangled files. We don't need the full content, just to know that file exists, so once we obtain the first chunk we cancel the download.

src/apps/sync-engine/index.ts
dajimenezriv: Oops, true, changed.

sg-gs (approved):

PR #1112, opened 2025-11-03
Fix logout

dajimenezriv-internxt:
## What

1. Stop using many listeners for logout and centralize all inside just one listener to the event. The listener is set just `once` when we login so it cannot be called multiple times because the logout event is emitted more than once.
2. Use a default sync root when the app starts first time because it is empty at the beginning.

### DIFF
--- a/src/apps/main/virtual-root-folder/services.test.ts
+++ b/src/apps/main/virtual-root-folder/services.test.ts
@@ -1,16 +1,32 @@
-import { partialSpyOn } from '@/tests/vitest/utils.helper.test';
+import { calls, partialSpyOn } from '@/tests/vitest/utils.helper.test';
 import { configStore } from '../config';
 import { getRootVirtualDrive } from './service';
 import * as getUserOrThrowModule from '../auth/service';
 import * as migrateOldSyncRootModule from './migrate-old-sync-root';
 import { createAbsolutePath } from '@/context/local/localFile/infrastructure/AbsolutePath';
+import { PATHS } from '@/core/electron/paths';
 
 describe('getRootVirtualDrive', () => {
-  partialSpyOn(getUserOrThrowModule, 'getUserOrThrow');
+  const getUserOrThrowMock = partialSpyOn(getUserOrThrowModule, 'getUserOrThrow');
   const getMock = partialSpyOn(configStore, 'get');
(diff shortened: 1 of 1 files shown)


src/apps/main/auth/handlers.ts
```diff
@@ -29,7 +29,6 @@ export function getIsLoggedIn() {
 export function onUserUnauthorized() {
   logger.debug({ tag: 'AUTH', msg: 'User has been logged out because it was unauthorized' });
   eventBus.emit('USER_LOGGED_OUT');
-  setIsLoggedIn(false);
```
dajimenezriv, line 32: We don't need to do this here since inside the event `USER_LOGGED_OUT` we already doing it.

src/apps/main/virtual-root-folder/service.ts
```diff
@@ -15,7 +17,9 @@ function setSyncRoot(pathname: string): void {
 
 export function getRootVirtualDrive() {
   const user = getUserOrThrow();
-  const syncRoot = createAbsolutePath(configStore.get('syncRoot'));
+
+  const defaultSyncRoot = join(PATHS.HOME_FOLDER_PATH, `InternxtDrive - ${user.uuid}`);
+  const syncRoot = createAbsolutePath(configStore.get('syncRoot') || defaultSyncRoot);
```
dajimenezriv, line 22: `configStore.get('syncRoot')` can be empty, so we need to set a default sync root.

src/backend/features/auth/services/utils/reset-config.ts
```diff
@@ -1,9 +1,8 @@
-import ConfigStore from '@/apps/main/config';
-import { defaults } from '@/core/electron/store/defaults';
-import { fieldsToSave } from '@/core/electron/store/fields-to-save';
+import { configStore } from '@/apps/main/config';
+import { defaults, fieldsToReset, fieldsToSave } from '@/core/electron/store/defaults';
 
 export function resetConfig() {
-  for (const field of fieldsToSave) {
-    ConfigStore.set(field, defaults[field]);
+  for (const field of [...fieldsToSave, ...fieldsToReset]) {
```
dajimenezriv, line 5: After saved the `fieldsToSave` in `saveConfig`, we want to reset the `fieldsToSave` and the `fieldsToReset`.

PR #1114, opened 2025-11-06
fix: avoid error popup when upload fails

cg-shokworks:


src/infra/inxt-js/file-uploader/upload-file.ts
```diff
@@ -21,38 +22,40 @@ type Props = {
...
+      interval = setInterval(() => abortOnChangeSize({ path, absolutePath, size, resolve, stopUpload, state }), 5000);
+      abortSignal.addEventListener('abort', () => {
+        logger.debug({ msg: 'Aborting upload', path });
+        stopUpload(state);
+      });
+    } catch (err: unknown) {
+      clearInterval(interval);
+      readable.close();
+      const error = err instanceof Error ? err : new Error(String(err));
+      resolve({ error: processError({ path, err: error, callbacks }) });
```
dajimenezriv: The errors that `processError` handles are the ones returned from `finishCallback`, here you can just return a `new EnvironmentFileUploaderError("UNKNOWN", error)`.

src/infra/inxt-js/file-uploader/upload-file.ts
```diff
@@ -21,38 +22,40 @@ type Props = {
...
+          callbacks.onProgress({ progress });
+        },
+      });
+
+      interval = setInterval(() => abortOnChangeSize({ path, absolutePath, size, resolve, stopUpload, state }), 5000);
+      abortSignal.addEventListener('abort', () => {
+        logger.debug({ msg: 'Aborting upload', path });
+        stopUpload(state);
+      });
+    } catch (err: unknown) {
```
dajimenezriv: Here, you don't need to manually set `unknown`. Also, sonarcloud is going to tell to name it as `error` to make it consistent across the whole codebase.

cg-shokworks: done

src/infra/inxt-js/file-uploader/upload-file.ts
cg-shokworks: it gives an error "not all code paths return a value"

cg-shokworks: I implementeda similar solution

PR #1115, opened 2025-11-06
Remove c++ logger and use path for notifications

dajimenezriv-internxt:
## What

This is the first of 2 PRs that implement the fetchData and cancelFetchData changes from node-win. This one has this test `contents-downloader.test.ts` failing and it will be fixed in the next PR (https://github.com/internxt/drive-desktop/pull/1111).

1. Update node-win version to 1.0.28. This one doesn't use a logger, so we need to remove all node-win loggers. Also, connectSyncRoot returns undefined and fetchData and cancelFetchData return the path in the callbacks.
2. Now that we have the paths in the callbacks we use it instead of the placeholderId in fetchData.
3. Now that we have the paths in the callbacks we can stop using different keys (uuid or path) when notifying the progress of an upload, move, rename, download, etc., now we have always the path so we can just use the path.

### DIFF
--- a/src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.ts
+++ b/src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.ts
@@ -13,16 +13,15 @@ type Props = {
 
 export async function syncRemoteChangesToLocal({ remote, local, virtualDrive }: Props) {
   const remoteDate = new Date(remote.updatedAt);
-
-  if (remote.size !== local.stats.size && remoteDate >= local.stats.mtime) {
+  if (remote.size !== local.stats.size && remoteDate > local.stats.mtime) {
     logger.debug({
       tag: 'SYNC-ENGINE',
       msg: 'Syncing remote changes to local',
       path: remote.path,
       remoteSize: remote.size,
       localSize: local.stats.size,
-      remoteDate,
-      localDate: local.stats.mtime,
+      remoteDate: remoteDate.toISOString(),
+      localDate: local.stats.mtime.toISOString(),
     });
 
     try {


src/node-win/virtual-drive.ts
```diff
@@ -52,14 +51,11 @@ export class VirtualDrive {
   }
 
   connectSyncRoot({ callbacks }: { callbacks: Callbacks }) {
-    const connectionKey = this.addon.connectSyncRoot({ callbacks });
-
-    logger.debug({ msg: 'connectSyncRoot', connectionKey });
-    return connectionKey;
+    return this.addon.connectSyncRoot({ callbacks });
```
dajimenezriv-internxt: Now the `connectionSyncRoot` does not return anything.

src/apps/sync-engine/callbacks/fetchData.service.ts
```diff
@@ -1,25 +1,30 @@
...
-    const tmpPath = await container.downloadFile.execute(filePlaceholderId, callback);
+    const { data: fileInfo, error } = NodeWin.getFileInfo({ ctx, path });
 
-    const uuid = NodeWin.getFileUuidFromPlaceholder({ placeholderId: filePlaceholderId });
-    const file = await container.downloadFile.fileFinderByUuid({ uuid });
+    if (error) throw error;
+
+    const tmpPath = await container.downloadFile.execute(fileInfo.placeholderId, callback);
+
+    // const file = await container.downloadFile.fileFinderByUuid({ uuid: fileInfo.uuid });
```
dajimenezriv, line 27: This is just temporal. This file will be used in the second PR of refactoring the fetch data (https://github.com/internxt/drive-desktop/pull/1111).

sg-gs:
The path is open to be changed while to placeholderId - I assume - it's unique, what is the advantage of that change @dajimenezriv-internxt ?

src/context/virtual-drive/files/application/file-creator.test.ts
```diff
@@ -45,8 +45,7 @@ describe('File Creator', () => {
     await expect(promise).rejects.toThrowError(FolderNotFoundError);
 
     expect(ipcRendererSyncEngineMock.send).toBeCalledWith('FILE_UPLOAD_ERROR', {
-      key: 'C:\\Users\\user\\InternxtDrive\\cat.png',
-      nameWithExtension: 'cat.png',
+      path: '/folder/cat.png',
```
sg-gs, line 48: Should you keep the Windows-based format or the current implementation is agnostic to the path-specific details of the underlying fs?

src/node-win/virtual-drive.ts
sg-gs, line 54: I have accidentally deleted a comment here made by you @dajimenezriv-internxt about that the connectSyncRoot now does not return anything

dajimenezriv: does not `return` anything.

src/context/virtual-drive/files/application/file-creator.test.ts
dajimenezriv: It's agnostic, we use the path to differenciate two different notifications in the frontend (UPLOAD /file.txt or UPLOAD /folder/file.txt), but we just display the basename of the path (file.txt), so it doesn't matter if windows or posix. Also, we are migrating everything to posix paths since I discovered that C++ also works with posix paths and we don't need to do the parse. This will allow to have just one type of path and match the functionality of `linux` and `core`.

dajimenezriv:
@sg-gs 
I don't know why but I can't replay (or I don't know how to) you directly when you write like a global comment. The path in this case it's also unique during the moment of fetching data/cancel fetch data. Also it allows for us to display it in the logs the path which is more useful to readability and to be able to unify how we display the notifications in the frontend (now we have the path everywhere, so we don't need to use path/uuid, we can just stick with the path). Anyway, the first thing that we do when fetching data is obtaining the uuid from the path (the placeholderId is `FILE:${uuid}`). So technically, now we still have the placeholderId but also the path.

src/context/virtual-drive/contents/application/ContentsDownloader.ts
```diff
@@ -16,58 +13,12 @@ export class ContentsDownloader {
   private downloaderIntanceCB: CallbackDownload | null = null;
   private downloaderFile: SimpleDriveFile | null = null;
```
sg-gs, line 15: Could you explain the removal of this part ? If this related to the temp file used to share content between desktop and node-win?

dajimenezriv: This is just to display notifications in the frontend (FILE_DOWNLOADING, FILE_DOWNLOAD_ERROR). We are no longer using events to display these notifications.  Now, the environment file uploader will have a callback called `onProgress` (https://github.com/internxt/drive-desktop/pull/1111/files#diff-b82ae857e5f7763081cf818137b131e3435d23d607e931cb1f76ee1fe6ce0a74R27)

dajimenezriv: https://github.com/internxt/drive-desktop/pull/1111/files#diff-6abe8dcf608d175e5b6d08bc1422273882577b661001c4c6bf9bc5519eda2fe2R24

sg-gs:
Okay @dajimenezriv-internxt, as long as file movements do not produce weird side-effects (as the path changes, but the placeholder does not - if in the same volume, as long as I have seen -), LGTM. Did you have that case taken care of?

dajimenezriv:
@sg-gs I'm not sure which case should we take care of. All paths (files and folders) in the cloud provider have a placeholderId. When the user double clicks in a file we get in C++ a struct `callbackInfo` that contains the path and the placeholderId. Previously we were sending the placeholderId to javascript, however, we cannot obtain the path from the placeholderId. Now, we send the path and we can obtain the placeholderId from the path. Which is what we do. If the path changes because we move the file, it doesn't matter, the placeholderId is still the same.

We don't rely in the path for anything important. We just use for logs, obtaining the uuid/placeholderId and now to be used as a key for displaying notifications in the frontend (which was already used partially).

sg-gs:
> **We don't rely in the path for anything important**. We just use for logs, obtaining the uuid/placeholderId and now to be used as a key for displaying notifications in the frontend (which was already used partially).

This is what I was looking for, perfect then @dajimenezriv-internxt

sg-gs (approved):

sg-gs:
The tests seem to be broken @dajimenezriv-internxt

PR #1117, opened 2025-11-08
Make electron store type safe

dajimenezriv-internxt:
## What

Instead of creating multiple instances of `electron-store` use just one and declare there all keys that we use so they are typed and have default values.

### DIFF
--- a/src/infra/inxt-js/file-uploader/abort-on-change-size.ts
+++ b/src/infra/inxt-js/file-uploader/abort-on-change-size.ts
@@ -3,21 +3,23 @@ import { fileSystem } from '@/infra/file-system/file-system.module';
 import { EnvironmentFileUploaderError } from './process-error';
 import type { TResolve } from './environment-file-uploader';
 import { logger } from '@internxt/drive-desktop-core/build/backend';
+import { ActionState } from '@internxt/inxt-js/build/api';
 
 type Props = {
   path: string;
   absolutePath: AbsolutePath;
   size: number;
   resolve: TResolve;
-  stopUpload: () => void;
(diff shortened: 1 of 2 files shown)


src/apps/main/config.ts
```diff
@@ -3,21 +3,38 @@ import { AppStore } from '@/core/electron/store/app-store.interface';
...
+const configStore = new Store({ schema, defaults });
+
+function get<T extends keyof AppStore>(key: T) {
+  return configStore.get(key);
+}
+
+function set<T extends keyof AppStore>(key: T, value: AppStore[T]) {
+  return configStore.set(key, value);
+}
+
```
dajimenezriv, line 38: We expose this new functions to make sure that we can only call the `electronStore` with typed keys.

src/apps/main/config/service.ts
```diff
@@ -8,10 +8,6 @@ import { ConfigTheme } from './theme.types';
 
 export type StoredValues = keyof AppStore;
 
-export const getConfigKey = <T extends StoredValues>(key: T): AppStore[T] => {
-  return store.get(key);
-};
-
 type SetConfigKeyProps = { key: 'preferedLanguage'; value: Language } | { key: 'preferedTheme'; value: ConfigTheme };
 
 export const setConfigKey = ({ key, value }: SetConfigKeyProps): void => {
```
egalvis27, line 13: With the new system, isn't this method redundant?

dajimenezriv: Nop, because it also broadcast theme or language to frontend when we modify one of those keys.

PR #1121, opened 2025-11-11
Use absolute paths for file progress

dajimenezriv-internxt:
## What

1. Stop using dot notation in `electron.store`. Basically if a key is `migrations.whatever-migration` we want to store all together instead of separate by dots like this:

```
"migrations": {
"whatever-migration"
}
```

2. Enforce absolute paths when displaying file uploads and actions. We were having an error because something we were using relative paths and other times it was absolute paths.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "internxt-drive",
-  "version": "2.6.1",
+  "version": "2.6.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "internxt-drive",
-      "version": "2.6.1",
+      "version": "2.6.2",
       "license": "MIT",
       "dependencies": {
         "@headlessui/react": "^1.4.2",


src/apps/main/config/language.ts
```diff
@@ -1,9 +1,8 @@
 import { broadcastToWindows } from '../windows';
-import { DEFAULT_LANGUAGE } from './language.types';
 import { electronStore } from '../config';
 
 export function getLanguage() {
-  return electronStore.get('preferedLanguage') || DEFAULT_LANGUAGE;
```
dajimenezriv, line 6: We already have the default values in the store.

src/backend/features/local-sync/watcher/events/rename-or-move/move-item.test.ts
```diff
@@ -20,7 +20,7 @@ describe('move-item', () => {
       uuid: 'uuid' as FileUuid,
       item: { parentUuid: 'parentUuid' },
       itemName: 'name',
-      path: createRelativePath('folder', 'newName'),
+      path: '/folder/newName' as AbsolutePath,
```
egalvis27, line 23: Isn't it better to use CreateAbsolutePath?

egalvis27 (approved):

PR #1122, opened 2025-11-11
Backups save to sqlite asap

dajimenezriv-internxt:
## What

Currently we search for the current state of backups in `drive-server-wip` using get files/folders by folder request. However, this request has a limit of 50 items, so in backups of 100k items this is so slow. This is the first of 2 PRs to migrate to look for the current state in the sqlite database instead of remote.

1. We assume that the local sqlite matches the remote database (now we have the sync by checkpoint and the recovery sync).
2. We also want to save to sqlite as soon as we upload the items in backups instead of waiting until the sync is performed to have this data locally.

The objetive of this PR is the second point. We create some common functions to be used in backups and sync.

### DIFF
--- a/src/apps/backups/folders/create-folders.ts
+++ b/src/apps/backups/folders/create-folders.ts
@@ -39,12 +39,9 @@ export async function createFolders({ self, context, added, tree, tracker }: TPr
         parentPath,
       });
 
-      /**
-       * v2.5.3 Daniel Jiménez
-       * TODO: add issue
-       */
+      context.addIssue({ error: 'CREATE_FOLDER_FAILED', name: localFolder.absolutePath });
     } else {
-      const { data: folder } = await createFolder({
+      const { data: folder, error } = await createFolder({
         path: localFolder.relativePath,
         plainName: basename(localFolder.relativePath),
         parentUuid: parent.uuid,
@@ -58,11 +55,8 @@ export async function createFolders({ self, context, added, tree, tracker }: TPr
           path: localFolder.relativePath,
           absolutePath: localFolder.absolutePath,
         };
-      } else {
-        /**
-         * v2.5.3 Daniel Jiménez
-         * TODO: add issue
-         */
+      } else if (error.code !== 'ABORTED') {
+        context.addIssue({ error: 'CREATE_FOLDER_FAILED', name: localFolder.absolutePath });
       }
     }
 
--- a/src/apps/backups/folders/create-folders.test.ts
+++ b/src/apps/backups/folders/create-folders.test.ts
@@ -16,6 +16,7 @@ describe('create-folders', () => {
       currentProcessed: vi.fn(),
     },
     context: {
+      addIssue: vi.fn(),
       abortController: {
         signal: {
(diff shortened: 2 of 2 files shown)


src/context/virtual-drive/files/infrastructure/HttpRemoteFileSystem.ts
```diff
@@ -53,15 +57,11 @@ export class HttpRemoteFileSystem {
       if (data) return data;
     }
 
-    if (error.code === 'FILE_ALREADY_EXISTS') {
```
dajimenezriv, line 56: We move this to the above function.

src/infra/drive-server-wip/out/ipc-main.ts
```diff
@@ -1,40 +1,19 @@
...
-      LocalSync.SyncState.addItem({ action: 'DELETE_ERROR', path });
-    } else {
-      LocalSync.SyncState.addItem({ action: 'DELETED', path });
-      await SqliteModule.FolderModule.updateByUuid({ uuid, payload: { status: 'TRASHED' } });
-    }
-
-    return res;
-  });
+  ipcMainDriveServerWip.handle('storageDeleteFileByUuid', (_, props) => deleteFileByUuid(props));
+  ipcMainDriveServerWip.handle('storageDeleteFolderByUuid', (_, props) => deleteFolderByUuid(props));
```
dajimenezriv, line 15: We extract these functions so we can be reused also in backups. But `storageDeleteByFileUuid` and `storageDeleteByFolderUuid` are still the same.

src/apps/backups/folders/create-folders.ts
```diff
@@ -45,17 +44,17 @@ export async function createFolders({ self, context, added, tree, tracker }: TPr
        * TODO: add issue
```
egalvis27: Since you're working on this file, don't you think it's a good time to resolve this TODO?

dajimenezriv: Done

src/infra/drive-server-wip/out/ipc-main.ts
```diff
@@ -64,3 +43,39 @@ export function setupIpcDriveServerWip() {
...
+
+export async function createFolder({ plainName, parentUuid, path, userUuid, workspaceId }: CreateFolderProps) {
+  const res = await HttpRemoteFolderSystem.persist({ plainName, parentUuid, path, workspaceId });
+
+  if (res.error) return res;
+
+  return await SqliteModule.FolderModule.createOrUpdate({
+    folder: { ...res.data, userUuid, workspaceId },
+  });
+}
```
egalvis27, line 81: Don't you think it's very strange that these functions are imported from the IPC-Main file?

dajimenezriv: Yes, it's unusual for them to be exported from ipc-main. However, the idea is to eventually combine the drive and backups, so I'm still not entirely sure where to put them. It's also because the PR is quite large and those functions were already there, so it would mean making more changes.

PR #1123, opened 2025-11-12
Backups take state from sqlite

dajimenezriv-internxt:
## What

This is the second part of 2 PRs. Now we stop fetching the current state of backups to the `drive-server-wip` and we just retrieve it from our local sqlite. So we can remove all endpoints of fetch files/folder by folder since they are no longer used.

### DIFF
--- a/.github/workflows/check-pr-size.yml
+++ b/.github/workflows/check-pr-size.yml
@@ -7,8 +7,9 @@ on:
 jobs:
   check_pr_size:
     # Ignore PRs with name merge-X-X-X-release
-    if: |
-      !(startsWith(github.head_ref, 'merge-') && endsWith(github.head_ref, '-release') && github.base_ref == 'main')
+    if: false
+    # if: |
+    #   !(startsWith(github.head_ref, 'merge-') && endsWith(github.head_ref, '-release') && github.base_ref == 'main')
     runs-on: ubuntu-latest
     timeout-minutes: 1
 


src/apps/main/remote-sync/handlers.ts
```diff
@@ -129,17 +129,11 @@ ipcMain.handle('SYNC_MANUALLY', async () => {
...
-  const { data: folders = [] } = await driveServerWip.folders.getFoldersByFolder({
-    folderUuid,
-    query: {
-      limit: FETCH_LIMIT_50,
-      offset: 0,
-      sort: 'updatedAt',
-      order: 'DESC',
-    },
-  });
+  const { data: folder } = await driveServerWip.backup.fetchFolder({ folderUuid });
```
dajimenezriv, line 132: This function does the same but was explicitly created for backups. Also it's used in `getBackupsFromDevice` so we stick with the same implementation as there.

src/apps/backups/remote-tree/traverser.ts
```diff
@@ -56,11 +56,15 @@ export class Traverser {
...
-    });
+    const [{ data: files = [] }, { data: folders = [] }] = await Promise.all([
+      SqliteModule.FileModule.getByWorkspaceId({ userUuid: context.userUuid, workspaceId: '' }),
+      SqliteModule.FolderModule.getByWorkspaceId({ userUuid: context.userUuid, workspaceId: '' }),
+    ]);
+
+    const items = {
+      files: files.filter((file) => file.status === 'EXISTS'),
+      folders: folders.filter((folder) => folder.status === 'EXISTS'),
+    };
```
dajimenezriv, line 67: This is the main part. Instead of fetch the items we just retrieve them from our local sqlite.

.github/workflows/check-pr-size.yml
```diff
@@ -7,8 +7,9 @@ on:
 jobs:
   check_pr_size:
     # Ignore PRs with name merge-X-X-X-release
-    if: |
-      !(startsWith(github.head_ref, 'merge-') && endsWith(github.head_ref, '-release') && github.base_ref == 'main')
+    if: false
+    # if: |
+    #   !(startsWith(github.head_ref, 'merge-') && endsWith(github.head_ref, '-release') && github.base_ref == 'main')
```
dajimenezriv, line 12: I disable the check pr size since this PR mainly removes all fetchs to retrieve the state from `drive-server-wip`.

PR #1125, opened 2025-11-12
Fix posix path issue

dajimenezriv-internxt:
## What

1. Stop using posix basename/dirname, and use the default basename/dirname that already handles win32 and posix.
2. Centralize how we stop and clear sync engine workers.
3. Add issue when file is bigger than MAX_FILE_SIZE.

### DIFF
--- a/src/backend/features/remote-sync/file-explorer/update-file-placeholder.test.ts
+++ b/src/backend/features/remote-sync/file-explorer/update-file-placeholder.test.ts
@@ -1,7 +1,7 @@
 import { mockDeep } from 'vitest-mock-extended';
 import { FilePlaceholderUpdater } from './update-file-placeholder';
 import VirtualDrive from '@/node-win/virtual-drive';
-import { mockProps, partialSpyOn } from '@/tests/vitest/utils.helper.test';
+import { call, mockProps, partialSpyOn } from '@/tests/vitest/utils.helper.test';
 import * as validateWindowsName from '@/context/virtual-drive/items/validate-windows-name';
 import { AbsolutePath, createRelativePath } from '@/context/local/localFile/infrastructure/AbsolutePath';
 import { FileUuid } from '@/apps/main/database/entities/DriveFile';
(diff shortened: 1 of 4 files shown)


src/context/virtual-drive/shared/application/PlatformPathConverter.ts
```diff
@@ -1,23 +1,7 @@
-import path, { posix } from 'node:path';
+import path from 'node:path';
 
 export class PlatformPathConverter {
-  static winToPosix(win: string): string {
```
dajimenezriv, line 4: This function and `getFatherPathPosix` were only used in the above test.

src/context/virtual-drive/contents/application/ContentsUploader.ts
```diff
@@ -23,7 +23,7 @@ export class ContentsUploader {
 
     if (contentsId) return { id: contentsId, size: stats.size };
 
-    if (error && error.code !== 'UNKNOWN' && error.code !== 'FILE_MODIFIED') {
+    if (error && error.code === 'NOT_ENOUGH_SPACE') {
```
dajimenezriv, line 26: Here we have `UNKNOWN`, `FILE_MODIFIED`, `ABORTED` and `NOT_ENOUGH_SPACE`. Only `NOT_ENOUGH_SPACE` is a problem.

src/apps/main/background-processes/sync-engine/services/stop-sync-engine-worker.ts
```diff
@@ -16,7 +13,24 @@ export function stopSyncEngineWorker({ worker }: Props) {
   workers.delete(ctx.workspaceId);
 }
 
-export async function stopSyncEngineWorkers() {
+export async function cleanSyncEngineWorker({ worker }: { worker: WorkerConfig }) {
+  const { ctx } = worker;
+
+  stopSyncEngineWorker({ worker });
+  await sleep(2000);
```
cg-shokworks, line 20: why do we need this sleep?

dajimenezriv: We need to wait until the sync engine worker is stopped to unregister the virtual drive, otherwise it can has problems unregistering. I set 2s because it works with that time, maybe it could be less, but it's better to be sure that the operation completes.

PR #1137, opened 2025-11-18
Refactor dangled files

dajimenezriv-internxt:


src/apps/sync-engine/dangled-files/check-dangled-files.ts
```diff
@@ -0,0 +1,38 @@
...
+    const { data: fileInfo } = NodeWin.getFileInfo({ ctx, path: file.absolutePath });
+
+    if (!fileInfo) return false;
+
+    if (fileInfo.pinState !== PinState.AlwaysLocal) {
+      ctx.logger.warn({ msg: 'Dangled file is not hydrated', path: file.absolutePath });
+      return false;
+    }
+
+    return true;
```
sg-gs: Being inside the problematic date is not the same as being dangled. It could be or not, careful with the naming

dajimenezriv: In the end I renamed both functions to `checkDangledFiles` and `checkDangleFile` so we don't say explicitely in any moment that it's a dangled file, only in the logs.

src/infra/inxt-js/contents-downloader/contents-downloader.ts
```diff
@@ -18,19 +20,21 @@ export class ContentsDownloader {
...
-  download({ contentsId, onProgress }: { contentsId: ContentsId; onProgress: (progress: number) => void }) {
+  download({ path, contentsId }: { path: AbsolutePath; contentsId: ContentsId }) {
     return new Promise((resolve: Resolve) => {
       this.state = this.environment.download(
         this.bucket,
         contentsId,
         {
-          progressCallback: (progress) => onProgress(progress),
+          progressCallback: (progress) => {
+            ipcRendererSyncEngine.send('FILE_DOWNLOADING', { path, progress });
```
dajimenezriv, line 30: `onProgress` callback was extracted so it didn't display any progress in dangled files. However, this is never executed in dangled files since we are not consuming the readable stream, so we can move the on progress here.

src/infra/inxt-js/contents-downloader/contents-downloader.ts
```diff
@@ -18,19 +20,21 @@ export class ContentsDownloader {
...
+          },
           finishedCallback: (error, stream) => {
             if (stream) {
               return resolve({ data: stream });
             }
 
-            return resolve({ error });
+            return resolve({ error: error ?? new Error('UNKNOWN') });
```
dajimenezriv, line 37: Technically error should never be null if we have the `stream`, but typescript says that it can be null.

src/context/virtual-drive/files/application/FileOverwriteContent.ts
```diff
@@ -75,7 +75,7 @@ export class FileOverwriteContent {
         logger.debug({ msg: 'Trying to download file ', uuid: file.uuid, name: file.name });
 
         const { error } = await ctx.contentsDownloader.download({
-          onProgress: () => ctx.contentsDownloader.forceStop(),
+          path: file.absolutePath,
           contentsId: file.contentsId,
         });
```
dajimenezriv, line 80: Ignore this. It's the old implementation of dangled files.

src/apps/sync-engine/dangled-files/check-dangled-file.ts
```diff
@@ -0,0 +1,45 @@
...
+      path: file.absolutePath,
+      contentsId: file.contentsId,
+    });
+
+    if (data) {
+      ctx.logger.debug({ msg: 'Not dangled file', path: file.absolutePath });
+
+      await ipcRendererSqlite.invoke('fileUpdateByUuid', { uuid: file.uuid, payload: { isDangledStatus: false } });
+
+      ctx.contentsDownloader.forceStop();
```
dajimenezriv, line 23: Previously we were stopping the download when progress was made, however, since we don't consume the stream the download was never stopped (it was also never started, but better to abort the download). Now we stop it once we receive the stream.

src/apps/sync-engine/dangled-files/check-dangled-file.ts
```diff
@@ -0,0 +1,45 @@
...
+
+      return;
+    }
+
+    if (error.message.includes('not found')) {
+      ctx.logger.warn({ msg: 'Dangled file found', path: file.absolutePath, error });
+
+      const stats = await FileSystemModule.statThrow({ absolutePath: file.absolutePath });
+
+      await updateContentsId({
```
sg-gs, line 33: What does this function do?

dajimenezriv: It's the same we use when a file has been modified locally. Basically it uploads the file again, obtains the new `fileId` and performs the request `replaceFile` against `drive-server-wip` to update the `fileId`.

src/apps/sync-engine/dangled-files/check-dangled-files.ts
```diff
@@ -0,0 +1,33 @@
...
+const startDate = new Date('2025-02-19T12:40:00.000Z').getTime();
+const endDate = new Date('2025-03-04T14:00:00.000Z').getTime();
+
+export async function checkDangledFiles({ ctx }: { ctx: ProcessSyncContext }) {
+  const { files } = await Traverser.run({ ctx });
+
+  ctx.logger.debug({ msg: 'Checking dangled files', total: files.length });
+
+  for (const file of files) {
+    if (file.isDangledStatus === false) return;
```
sg-gs: You should `continue`, not `return`, otherwise the loop ends at any time. This works for .map etc as they affect a function as an individual handler, but this is not the case

src/apps/sync-engine/dangled-files/check-dangled-files.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const createdAt = new Date(file.createdAt).getTime();
+
+    if (createdAt < startDate || createdAt > endDate) return;
```
sg-gs: Same here

src/apps/sync-engine/dangled-files/check-dangled-files.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const { data: fileInfo } = NodeWin.getFileInfo({ ctx, path: file.absolutePath });
+
+    if (!fileInfo) return;
```
sg-gs: Same here

src/apps/sync-engine/dangled-files/check-dangled-files.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    if (fileInfo.pinState !== PinState.AlwaysLocal) {
```
sg-gs, line 26: Same here. I will add a test for a case where there is more than 1 file, so this case is not missed by error

src/apps/sync-engine/dangled-files/check-dangled-files.ts
dajimenezriv: Shit, I made the change but I forgot to commit. Changing and creating the test.

src/apps/sync-engine/dangled-files/check-dangled-files.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    // Then
+    call(checkDangledFileMock).toMatchObject({ file: { absolutePath: '/file.txt' } });
+  });
+
+  it('should check through multiple files', async () => {
+    // Given
+    getFileInfoMock.mockReturnValue({ data: { pinState: PinState.AlwaysLocal } });
+    traverserMock.mockResolvedValue({
+      files: [
+        { absolutePath: abs('/file1.txt'), isDangledStatus: true, createdAt: '2025-03-01T00:00:00.000Z' },
```
sg-gs, line 72: You should add a false one, and should not be the last, so you enforce the continue to check if the entire loop is still completed

dajimenezriv: True, done

sg-gs: It does not appear to me. I may have the previous version cached, confirm it to me please

PR #1138, opened 2025-11-18
Refactor dangled files

dajimenezriv-internxt:


src/migrations/v2.5.1/add-user-uuid-to-database.ts
```diff
@@ -9,14 +10,8 @@ export class AddUserUuidToDatabase {
...
-        driveFilesCollection.updateInBatch({
-          where: { userUuid: '' },
-          payload: { userUuid: user.uuid },
-        }),
-        driveFoldersCollection.updateInBatch({
-          where: { userUuid: '' },
-          payload: { userUuid: user.uuid },
-        }),
+        fileRepository.update({ userUuid: '' }, { userUuid: user.uuid }),
+        folderRepository.update({ userUuid: '' }, { userUuid: user.uuid }),
```
dajimenezriv, line 14: This does the same, but it was the last usage of the collections since we deleted dangled files, so we just replace with the operation directly to the repository and it's easier. This only happens for migrations. To access the sqlite from the code we use the `SqliteModule`.

AlexisMora (approved):

PR #1139, opened 2025-11-18
Migrate relative to absolute paths

dajimenezriv-internxt:


src/node-win/virtual-drive.ts
```diff
@@ -78,77 +64,59 @@ export class VirtualDrive {
   }
 
   createFileByPath({
```
dajimenezriv, line 66: We can safely throw exceptions from `createFileByPath`, `createFolderByPath` and `convertToPlaceholder` because they are captured in the function that calls them.

src/node-win/addon-wrapper.ts
```diff
@@ -1,14 +1,17 @@
...
 import { Callbacks } from './types/callbacks.type';
 import { logger } from '@/apps/shared/logger/logger';
 import { FilePlaceholderId } from '@/context/virtual-drive/files/domain/PlaceholderId';
 import { FolderPlaceholderId } from '@/context/virtual-drive/folders/domain/FolderPlaceholderId';
+import { posix, win32 } from 'node:path';
 
-export class Addon {
-  syncRootPath!: string;
+function toWin32(path: AbsolutePath) {
+  return path.replaceAll(posix.sep, win32.sep) as Win32Path;
```
dajimenezriv, line 11: We parse to windows paths in the addon instead of the virtual drive since it's the last step before C++. Also, maybe we reach a moment in which we can remove virtual drive and use the addon directly.

src/node-win/addon.ts
```diff
@@ -5,7 +5,9 @@ import { Callbacks } from './types/callbacks.type';
 import { addon as rawAddon } from '@internxt/node-win/dist';
 import { FilePlaceholderId } from '@/context/virtual-drive/files/domain/PlaceholderId';
 import { FolderPlaceholderId } from '@/context/virtual-drive/folders/domain/FolderPlaceholderId';
-import { AbsolutePath } from '@internxt/drive-desktop-core/build/backend';
+import { Brand } from '@internxt/drive-desktop-core/build/backend/core/utils/brand.types';
+
+export type Win32Path = Brand<string, 'Win32Path'>;
```
dajimenezriv, line 10: We use `AbsolutePath` everywhere except in the last moment that we are going to send them to C++ using the `addon-wrapper.ts` from above.

src/infra/node-win/services/item-identity/get-folder-info.ts
```diff
@@ -15,11 +16,11 @@ export class GetFolderInfoError extends Error {
 
 type TProps = {
   ctx: ProcessSyncContext;
-  path: string;
+  path: AbsolutePath;
 };
 
 export function getFolderInfo({ ctx, path }: TProps) {
-  if (path === '/' || path === ctx.virtualDrive.syncRootPath || `${path}\\` === ctx.virtualDrive.syncRootPath || path === ctx.rootPath) {
+  if (path === ctx.rootPath) {
```
dajimenezriv, line 23: We need to check if we are trying to obtain the placeholder info of the root since is not a placeholder.

.github/workflows/check-pr-size.yml
```diff
@@ -7,9 +7,8 @@ on:
 jobs:
   check_pr_size:
     # Ignore PRs with name merge-X-X-X-release
-    if: false
-    # if: |
-    #   !(startsWith(github.head_ref, 'merge-') && endsWith(github.head_ref, '-release') && github.base_ref == 'main')
+    if: |
+      !(startsWith(github.head_ref, 'merge-') && endsWith(github.head_ref, '-release') && github.base_ref == 'main')
```
dajimenezriv, line 11: We disabled the check pr size in the last PR, now we enable it again.

src/infra/node-win/services/item-identity/get-folder-info.ts
sg-gs: I assume that the removed checks are being run - when applicable, as you are using Node.js paths now - when setting the `rootPath` value?

src/node-win/addon-wrapper.ts
sg-gs: Now, this function is pretty important, I would consider adding tests for it, it is the only translation point between Node.js paths and the Win32 ones

src/infra/node-win/services/item-identity/get-folder-info.ts
dajimenezriv: Basically here we could receive a string path (it could be C:/posix/absolute/path, C:\win32\absolute\path or /posix/relative/path). Now we are sending an AbsolutePath that is already parsed (C:/posix/absolute/path) and rootPath is also a branded `AbsolutePath`. We can never have here `/`, and `syncRootPath` now is the same as `rootPath` so it will be removed soon.

src/node-win/addon-wrapper.ts
dajimenezriv: It's already tested indirectly in `virtual-drive.test.ts`, since we test in which position we send the props to the addon (because it's the only place in which we can't use object props) and we check that the prop has win32 path format.

It's not the best, but as said, we are going to join `virtual-drive.ts` and `addon-wrapper.ts` soon. In this PR (the next one) https://github.com/internxt/drive-desktop/pull/1140 we already merging them and we will add a test to each addon function.

sg-gs (approved):
ℹ️ My approval does not mean that the other reviewers should skip the code review process

PR #1140, opened 2025-11-19
Move to packages addon

dajimenezriv-internxt:
## What

1. Refactor virtual drive and addon, so in the end we can fully remove the virtual drive.
2. Rename package from @internxt/node-win to @packages/addon, since it makes more sense.
3. Replace the `gyp.config.py` to `gyp.config.js` about how to build the C++ addon, so we don't have code in python and everything is javascript or C++.
4. Now `create_file_placeholder.cpp` and `create_folder_placeholder.cpp` accept a path to make it consistent across all addon functions.

### DIFF
--- a/src/apps/sync-engine/BindingManager.ts
+++ b/src/apps/sync-engine/BindingManager.ts
@@ -57,7 +57,7 @@ export class BindingsManager {
   }
 
   static watch({ ctx }: { ctx: ProcessSyncContext }) {
-    const { watcher } = createWatcher({ ctx });
+    const { watcher } = createWatcher();
 
     watcher.watchAndWait({ ctx });
   }
--- a/src/apps/sync-engine/create-watcher.ts
+++ b/src/apps/sync-engine/create-watcher.ts
@@ -1,12 +1,7 @@
 import { Watcher } from '@/node-win/watcher/watcher';
-import { ProcessSyncContext } from './config';
 
-type TProps = {
-  ctx: ProcessSyncContext;
-};
-
-export function createWatcher({ ctx }: TProps) {
-  const watcher = new Watcher(ctx.rootPath, {
+export function createWatcher() {
+  const watcher = new Watcher({
     depth: undefined,
     followSymlinks: true,
     ignored: /(^|[/\\])\../,
--- a/src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.infra.test.ts
+++ b/src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.infra.test.ts
@@ -36,9 +36,8 @@ describe('sync-remote-changes-to-local', () => {
 
   it('should sync remote changes to local', async () => {
     // Given
-    const createWatcherProps = mockProps<typeof createWatcher>({ ctx: { virtualDrive } });
-    const { watcher } = createWatcher(createWatcherProps);
-    const watcherProps = mockProps<typeof watcher.watchAndWait>({ ctx: { virtualDrive } });
+    const { watcher } = createWatcher();
+    const watcherProps = mockProps<typeof watcher.watchAndWait>({ ctx: { virtualDrive, rootPath } });
     watcher.watchAndWait(watcherProps);
     await sleep(100);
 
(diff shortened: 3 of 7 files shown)


src/node-win/addon-wrapper.ts
```diff
@@ -11,23 +11,24 @@ function toWin32(path: AbsolutePath) {
...
+      fn,
+      error: result.error,
+    });
   }
 
-  registerSyncRoot({
+  return data;
+}
+
+export class Addon {
```
dajimenezriv, line 30: We can make all functions static since the C++ addon doesn't need to be constructed. This is good since we won't need to send the `virtualDrive` doing prop drilling with the context.

packages/addon/gyp.config.js
```diff
@@ -0,0 +1,43 @@
+const { readFileSync, readdirSync, writeFileSync } = require('fs');
```
dajimenezriv, line 1: This is a script to create the `binding.gyp`, previously it was created in python, now it's done in javascript.

packages/addon/native-src/virtual_drive/create_file_placeholder.cpp
```diff
@@ -31,7 +32,7 @@ napi_value create_file_placeholder_impl(napi_env env, napi_callback_info info)
     cloudEntry.FsMetadata.BasicInfo.FileAttributes = FILE_ATTRIBUTE_NORMAL;
     cloudEntry.FsMetadata.BasicInfo.CreationTime = creationTime;
     cloudEntry.FsMetadata.BasicInfo.LastWriteTime = lastWriteTime;
-    cloudEntry.FsMetadata.BasicInfo.LastAccessTime = lastAccessTime;
+    cloudEntry.FsMetadata.BasicInfo.LastAccessTime = lastWriteTime;
```
sg-gs, line 35: Why?

packages/addon/native-src/virtual_drive/create_folder_placeholder.cpp
```diff
@@ -30,7 +31,7 @@ napi_value create_folder_placeholder_impl(napi_env env, napi_callback_info info)
     cloudEntry.FsMetadata.BasicInfo.FileAttributes = FILE_ATTRIBUTE_DIRECTORY;
     cloudEntry.FsMetadata.BasicInfo.CreationTime = creationTime;
     cloudEntry.FsMetadata.BasicInfo.LastWriteTime = lastWriteTime;
-    cloudEntry.FsMetadata.BasicInfo.LastAccessTime = lastAccessTime;
+    cloudEntry.FsMetadata.BasicInfo.LastAccessTime = lastWriteTime;
```
sg-gs, line 34: Why?

packages/addon/gyp.config.js
sg-gs: I assume this has been executed and works the same way?

packages/addon/package.json
```diff
@@ -1,17 +1,10 @@
 {
-  "name": "@internxt/node-win",
-  "version": "1.0.28",
-  "author": "Internxt <hello@internxt.com>",
-  "description": "Drive desktop node addon",
+  "name": "@packages/addon",
+  "license": "MIT",
```
sg-gs, line 3: Don't you prefer to keep the version in case you need to keep the track of something or log the add-on version?

packages/addon/native-src/virtual_drive/create_file_placeholder.cpp
dajimenezriv: In the javascript code we were sending Date.now() as `lastAccessTime`, however, talking with claude said that the most usual thing when we don't have a real access time (since the file comes from the cloud) is to use the `lastWriteTime`. But we can keep as before. There is no real answer.

packages/addon/package.json
dajimenezriv: Now the version of the addon is the version of `drive-desktop` since it's a package that only exists on windows. If we move to a previous commit, the current code of the `addon` is going to match the version of the current code of `drive-desktop`. They are the same. So we are setting the version of the addon in the `drive-desktop` package. Also, it's something less to mantain.

dajimenezriv: I would remove the `license` too since we already have it in the root `package.jon`. However, if we don't put the license it emits a warning everytime we execute a script from the package.json.

packages/addon/gyp.config.js
dajimenezriv: Exactly

packages/addon/native-src/virtual_drive/create_file_placeholder.cpp
sg-gs: Mmm, I see. Yeah, that seems like the best option.

PR #1143, opened 2025-11-20
Refactor download backup

dajimenezriv-internxt:
## What

Ticket https://inxt.atlassian.net/browse/PB-5369.

Migrate the old implementation of downloading a backup using the legacy shards download to the new one that we already use in drive. The process is as follow:
1. Create the ipc listener in main to `downloadBackup`, that can receive some specific folder uuids or the whole device uuid.
2. Setup the ipc listener to abort, create the environment for `inxt-js` and iterate through each folder uuid (or device uuid).
3. For each folder we use the already existing traverser to obtain all backup items that belong to that folder uuid. After that we download sequentially each file (it will be parallelized in the next PR) and update the progress to the frontend.
4. For each file we create the writable stream and using the `contents-downloader.ts` that we used in drive we download the file contents.

### DIFF
--- a/src/apps/renderer/hooks/backups/useBackupDownloadProgress.tsx
+++ b/src/apps/renderer/hooks/backups/useBackupDownloadProgress.tsx
@@ -16,7 +16,7 @@ export function useBackupDownloadProgress(): BackupDownloadContextProps {
   const [backupDownloadProgress, setBackupDownloadProgress] = useState<DownloadBackupProgress>({});
 
   useEffect(() => {
-    return window.electron.onBackupDownloadProgress(({ id, progress }) =>
+    return globalThis.window.electron.onBackupDownloadProgress(({ id, progress }) =>
       setBackupDownloadProgress((prevState) => {
         return { ...prevState, [id]: Math.round(progress) };
       }),
--- a/src/backend/features/backups/download/download-backup.test.ts
+++ b/src/backend/features/backups/download/download-backup.test.ts
@@ -51,7 +51,7 @@ describe('download-backup', () => {
     call(removeListenerMock).toStrictEqual(['abort-download-backups-deviceUuid', expect.any(Function)]);
   });
 
-  it('should use deviceUuid as rootUuids if no folderUuids are provided', async () => {
+  it('should use the device uuid as the root uuid if no folder uuids are provided', async () => {
     // When
     await downloadBackup(props);
     // Then
@@ -62,7 +62,7 @@ describe('download-backup', () => {
     });
   });
 
-  it('should use folderUuids as rootUuids if are provided', async () => {
+  it('should use folder uuids as the root uuids if they are provided', async () => {
     // Given
     props.folderUuids = ['folderUuid' as FolderUuid];
     // When
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -69,6 +69,7 @@ module.exports = {
     'sonarjs/slow-regex': 'off',
     'sonarjs/todo-tag': 'off',
     'unicorn/filename-case': ['warn', { case: 'kebabCase' }],
+    'unicorn/prefer-global-this': 'warn',
     'unicorn/prefer-node-protocol': 'error',
     'unicorn/prefer-string-raw': 'warn',
     'padding-line-between-statements': [
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "init:ci": "npm ci --ignore-scripts && node node_modules/electron/install.js",
     "init:dev": "npm install && node node_modules/electron/install.js && npm run build:dll && npm run reload-native-deps",
     "========== Code style ==========": "",
-    "lint": "eslint src --ext .ts,.tsx --max-warnings 398",
+    "lint": "eslint src --ext .ts,.tsx --max-warnings 443",
     "lint:fix": "npm run lint -- --fix",
     "format": "prettier src tests --check",
     "format:fix": "prettier src tests --write",


src/apps/backups/remote-tree/traverser.ts
```diff
@@ -27,44 +26,38 @@ export class Traverser {
     };
   }
 
-  private traverse(context: BackupsContext, tree: RemoteTree, items: Items, currentFolder: ExtendedDriveFolder) {
-    if (!items) return;
-    if (context.abortController.signal.aborted) return;
```
dajimenezriv, line 32: We don't really need to check the aborted signal here since this doesn't perform any complex iterations and finishes in milliseconds.

src/apps/sync-engine/download-contents/download-contents.ts
```diff
@@ -20,7 +20,9 @@ export async function downloadContents({
   try {
     const { data: readable, error } = await ctx.contentsDownloader.download({
       contentsId: file.contentsId,
-      path,
+      onProgress: (progress) => {
+        ipcRendererSyncEngine.send('FILE_DOWNLOADING', { path, progress });
+      },
```
dajimenezriv, line 25: We extract the emit progress to the frontend from the `contents-downloader` since this part in drive is executed in the `sync-engine` process, but for backups is in the `main` process, so it will break when trying to send the event.

src/apps/renderer/hooks/backups/useBackupDownloadProgress.tsx
```diff
@@ -17,12 +16,11 @@ export function useBackupDownloadProgress(): BackupDownloadContextProps {
   const [backupDownloadProgress, setBackupDownloadProgress] = useState<DownloadBackupProgress>({});
 
   useEffect(() => {
-    const removeListener = window.electron.onBackupDownloadProgress(({ id, progress }: { id: string; progress: number }) =>
+    return window.electron.onBackupDownloadProgress(({ id, progress }) =>
```
sg-gs: Why not `globalThis`?

src/apps/renderer/pages/Settings/Backups/Selector/DownloadSelector.tsx
```diff
@@ -81,24 +81,16 @@ export default function DownloadFolderSelector({ onClose }: DownloadFolderSelect
...
-        abortDownloadBackups(selected!);
-        onClose();
-      } catch {
-        // error while aborting (aborting also throws an exception itself)
-      } finally {
-        setTimeout(() => {
-          clearBackupDownloadProgress(selected!.uuid);
-        }, 600);
+  async function handleDownloadBackup() {
+    if (selected) {
```
sg-gs, line 85: Why?

src/backend/features/backups/download/download-folder.ts
```diff
@@ -0,0 +1,48 @@
...
+    });
+  }
+
+  const tree = await Traverser.run({ rootPath, rootUuid, userUuid: user.uuid });
+
+  const files = Object.values(tree.files);
+
+  let downloadedItems = 0;
+
+  updateProgress(1);
```
sg-gs, line 37: Should this be 0 if the download is aborted?

src/backend/features/backups/download/download-folder.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  for (const file of files) {
+    if (abortController.signal.aborted) return;
+
+    await downloadFile({ file, contentsDownloader });
+
+    downloadedItems += 1;
+    const progress = (downloadedItems / files.length) * 100;
```
sg-gs, line 45: In Web I recall using 50% items count, 50% items size, thus, you make the progress feel faster on corner cases like a lot of folders and only a few big files or a lot of small files only

src/backend/features/backups/download/download-backup.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  it('should set listeners', async () => {
+    // When
+    await downloadBackup(props);
+    // Then
+    call(onMock).toStrictEqual(['abort-download-backups-deviceUuid', expect.any(Function)]);
+    call(removeListenerMock).toStrictEqual(['abort-download-backups-deviceUuid', expect.any(Function)]);
+  });
+
+  it('should use deviceUuid as rootUuids if no folderUuids are provided', async () => {
```
sg-gs: I would avoid referencing variable / function names on the tests, as if they change, you have to change this part, but no one will remember you so at this is a legend without any validation. It's preferable to talk about what should happen, like: 

_Should use the device id as the root id if no folder ids are provided_

src/backend/features/backups/download/download-backup.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    await downloadBackup(props);
+    // Then
+    call(downloadFolderMock).toMatchObject({
+      device: { name: 'device', uuid: 'deviceUuid' },
+      rootPath: '/backup/Backup_20250101120000',
+      rootUuid: 'deviceUuid',
+    });
+  });
+
+  it('should use folderUuids as rootUuids if are provided', async () => {
```
sg-gs: Same here

src/backend/features/backups/download/download-file.ts
```diff
@@ -0,0 +1,35 @@
...
+
+type Props = {
+  file: ExtendedDriveFile;
+  contentsDownloader: InxtJs.ContentsDownloader;
+};
+
+export async function downloadFile({ file, contentsDownloader }: Props) {
+  logger.debug({ tag: 'BACKUPS', msg: 'Download file', path: file.absolutePath });
+
+  await mkdir(dirname(file.absolutePath), { recursive: true });
```
sg-gs, line 17: What is the behaviour if those subfolders already exist?

src/backend/features/backups/download/download-file.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  try {
+    const writeStream = createWriteStream(file.absolutePath);
+
+    const { data: readStream, error } = await contentsDownloader.download({ contentsId: file.contentsId });
+
+    if (error) throw error;
+
+    await pipeline(readStream, writeStream);
```
sg-gs, line 26: Nice

src/backend/features/backups/download/download-folder.ts
dajimenezriv: Exactly, or when it's finished. It's like the old implementation. In the frontend if the value is greater than 0 it means that it's in progress.

src/backend/features/backups/download/download-folder.ts
dajimenezriv: I know, here we have that ticket from long time ago for the upload, but it was never prioritized. Also because we added the current file being uploaded in the bottom right panel (the same as in the drive) and I think we can do the same for downloads. However, both are big implementations for this PR.

src/apps/renderer/pages/Settings/Backups/Selector/DownloadSelector.tsx
dajimenezriv: Technically `selected` can be null and in the function it was enforced using `!`. I prefer to be safe and do the condition. Anyway, most frontend code (this included) has never been touched since I came here and this `selected` condition happens a lot inside backups, so probably there are better ways to do it.

src/backend/features/backups/download/download-file.ts
dajimenezriv: It does nothing.

sg-gs (approved):

PR #1144, opened 2025-11-20
Clean legacy download backup

dajimenezriv-internxt:


dajimenezriv:
@sg-gs The approval was removed because I had to reinstall the `bip39`. I had deleted it since we weren't validating the mnemonic anymore, but since now we have added it in the login we still need the library (it was already working because it's a subdependency but better to have the direct dependency in the package.json).

sg-gs (approved):

PR #1145, opened 2025-11-22
Parallelize download backup

dajimenezriv-internxt:
## What

1. Simplify in frontend how we get the `downloadProgress` and check if it's in progress (`downloadProgress > 0`).
2. Add effect library for error management (https://inxt.atlassian.net/browse/PB-5369?focusedCommentId=50869).
3. Allow parallel downloads in javascript by using a map of paths that are being download and store their `ActionState` so we can stop them later.
4. Add concurrency to download files. This needs to be improved and joined with the limiters of uploading files and so, but they are still split in different processes.


src/infra/inxt-js/contents-downloader/contents-downloader.ts
```diff
@@ -37,10 +42,14 @@ export class ContentsDownloader {
           label: 'Dynamic',
           params: {
             useProxy: false,
-            chunkSize: 4 * 1024 * 1024,
+            chunkSize: 256 * 1024,
```
dajimenezriv, line 45: This is explained here: https://inxt.atlassian.net/browse/PB-5369?focusedCommentId=50868

src/infra/inxt-js/contents-downloader/contents-downloader.ts
```diff
@@ -1,26 +1,31 @@
 import { ActionState } from '@internxt/inxt-js/build/api';
 import { Environment } from '@internxt/inxt-js';
 import { ContentsId } from '@/apps/main/database/entities/DriveFile';
+import { AbsolutePath, logger, throwWrapper } from '@internxt/drive-desktop-core/build/backend';
+
+const downloads = new Map<AbsolutePath, ActionState>();
```
dajimenezriv, line 6: Now we are oficially able in javascript of download multiple files in parallel and stop them. Previously we were losing `state` references and only keeping the last one. This will be used also for drive downloads.

src/apps/renderer/hooks/backups/useBackupDownloadProgress.tsx
```diff
@@ -16,30 +16,25 @@ export function useBackupDownloadProgress(): BackupDownloadContextProps {
...
 
-  const [thereIsDownloadProgress, setThereIsDownloadProgress] = useState<boolean>(false);
-  const [downloadProgress, setDownloadProgress] = useState<number>(0);
+  const downloadProgress = useMemo(() => {
+    if (!selected) return 0;
 
-  useEffect(() => {
-    if (!selected?.uuid) return;
     const downloadProgress = backupDownloadProgress[selected.uuid];
-    if (downloadProgress && downloadProgress < 100) {
```
dajimenezriv, line 32: Now we consider `downloadProgress === 100` also as progress, why? If we have multiple folders selected for the download, every one has it's own bar of progress, so it's going to reach multiple times the 100 and we don't want the UI to not show the progress. After all folders are downloaded we send the ipc event of progress `0`.

Basically, let's say we have `folder1` and `folder2`:

```ts
updateProgress(1); // to start showing progress
updateProgress(downloadedFiles / totalFilesOfFolder1); // it will reach 100
updateProgress(1); // reset the progress
updateProgress(downloadedFiles / totalFilesOfFolder2); // it will reach 100
updateProgress(0);
```

src/backend/features/backups/download/download-file.ts
```diff
@@ -1,35 +1,42 @@
...
-      path: file.absolutePath,
-      error,
-    });
-  }
+    yield* pipeline({ readable, writable });
+  }).pipe(
+    Effect.catchTag('PipelineAborted', () => {
+      return Effect.void;
+    }),
+    Effect.catchAllCause((error) => {
```
dajimenezriv, line 35: We capture all other exceptions.

src/backend/features/backups/download/download-folder.ts
```diff
@@ -22,27 +24,44 @@ export async function downloadFolder({ user, device, rootUuid, rootPath, abortCo
...
+    });
+  });
 
-    downloadedItems += 1;
-    const progress = (downloadedItems / files.length) * 100;
-    updateProgress(Math.trunc(progress));
+  try {
+    await Promise.all(promises);
+  } catch (error) {
+    if (error instanceof Bottleneck.BottleneckError && error.message === 'This limiter has been stopped.') {
```
dajimenezriv, line 59: The point of `Effect` is to not do these things, since `instanceof` can be applied to anything and we don't have any type safety. Here is not applied yet, but it will be.

src/backend/features/backups/download/download-file.ts
```diff
@@ -1,35 +1,40 @@
...
-    logger.error({
-      tag: 'BACKUPS',
-      msg: 'Error downloading file',
-      path: file.absolutePath,
-      error,
-    });
-  }
+    yield* pipeline({ readable, writable });
+  }).pipe(
+    Effect.catchTag('PipelineAborted', () => Effect.void),
```
dajimenezriv, line 34: This is typed, if the code inside stops throwing a `PipelineAborted` we will have an error.

src/backend/features/backups/download/download-folder.ts
```diff
@@ -22,27 +24,44 @@ export async function downloadFolder({ user, device, rootUuid, rootPath, abortCo
...
   }
 
   const tree = await Traverser.run({ rootPath, rootUuid, userUuid: user.uuid });
-
   const files = Object.values(tree.files);
 
   let downloadedItems = 0;
 
+  const limiter = new Bottleneck({ maxConcurrent: 3 });
+  const runningFiles = new Set<AbsolutePath>();
```
dajimenezriv, line 36: We want to keep track of all files that the download has started but not finished so we can stop them.

src/apps/renderer/hooks/backups/useBackupDownloadProgress.tsx
sg-gs: I assume the 100% will be shown on the UI as 'downloaded'?

src/backend/features/backups/download/download-folder.ts
sg-gs: Have you considered using async.queue which manages the concurrency, abort state support and even adding an error callback or a drain method to wait for the queue to be ready for more?

src/apps/renderer/hooks/backups/useBackupDownloadProgress.tsx
dajimenezriv: Not really, the `updateProgress(1)` happens so fast so it will start again from the beginning. Maybe we can place here the name of the folder being downloaded instead of `Download Backup...`, but I will leave that to the next release since we have to do some changes in backend and frontend.

image

src/backend/features/backups/download/download-folder.ts
dajimenezriv: I didn't consider it. But we are already used `async.queue` for the antivirus and it gave some problems when the queue had 0 items (maybe it was also because we didn't appliy it well). But `bottleneck` works really well. It allows priorities, stop the queue and concurrency, which are the things that we need for now.

PR #1146, opened 2025-11-22
Async await create file and folder placeholder

dajimenezriv-internxt:


packages/addon/include/helpers/async_wrapper.h
```diff
@@ -0,0 +1,75 @@
...
+        napi_value error;
+        napi_create_string_utf8(env, asyncWork->error.c_str(), NAPI_AUTO_LENGTH, &error);
+        napi_reject_deferred(env, asyncWork->deferred, error);
+    }
+
+    napi_delete_async_work(env, asyncWork->work);
+}
+
+template <typename Func, typename... Args>
+napi_value run_async(napi_env env, const char* resource_name, Func&& fn, Args&&... args)
```
sg-gs, line 54: Which is the benefit of this?

sg-gs: Oh, I see, this is the same that was on hydrate file, but renamed and moved to a diff file

sg-gs (approved):

PR #1147, opened 2025-11-22
Remove virtual drive class

dajimenezriv-internxt:
## What

The C++ addon provides an object with some functions that can be used in the javascript code. Currently we were encapsulating that logic inside the `VirtualDrive` class and that had the downside of prop drilling with the `virtualDrive` variable. However, this is not necessary. Now use directly the `Addon` class with static functions. This and making the functions `async` will start removing the barrier of two different processes since it will be easier to run C++ code in the main process (we were already doing that but only with 2 functions that were already static).


src/backend/features/remote-sync/file-explorer/delete-item-placeholders.ts
```diff
@@ -12,9 +12,7 @@ type Props = { ctx: ProcessSyncContext } & (
 export async function deleteItemPlaceholders({ ctx, remotes, type }: Props) {
   for (const remote of remotes) {
     const { data: info } =
-      type === 'folder'
-        ? NodeWin.getFolderInfo({ path: remote.absolutePath, ctx })
-        : NodeWin.getFileInfo({ path: remote.absolutePath, ctx });
+      type === 'folder' ? NodeWin.getFolderInfo({ path: remote.absolutePath, ctx }) : NodeWin.getFileInfo({ path: remote.absolutePath });
```
sg-gs, line 15: I assume the context is still needed on `getFolderInfo` as it has not been still migrated

sg-gs:
I would consider renaming `Addon` to something more self-explanatory in the future @dajimenezriv-internxt

src/backend/features/remote-sync/file-explorer/delete-item-placeholders.ts
dajimenezriv: Not that, the `getFolderInfo` needs the `rootUuid` and the `rootPath`, because we can try to obatin the placeholder of the root folder `C:/Users/user/InternxtDrive - {uuid}` and we need to manually return the "placeholder". This happens when try to obtain the parent uuid of an item to perform some operations.

dajimenezriv:
@sg-gs Technically I like the name of `Addon` since it's the default name for Node C++ addons and it's so generic that if we need to use it for something else, the name works (like obtaining the items from the trash for the cleaner).

PR #1148, opened 2025-11-22
Migrate addon functions to async await

dajimenezriv-internxt:
## What

Ticket https://inxt.atlassian.net/browse/PB-5370.

1. Use the `run_async` wrapper to make most of the functions of C++ async.
2. Migrate those functions from `sync` to `async` in javascript.


packages/addon/include/placeholders_interface/Placeholders.h
```diff
@@ -1,20 +1,19 @@
...
+struct FileState {
     std::string placeholderId;
     CF_PIN_STATE pinState;
 };
 
 class Placeholders
 {
-public:
-    static winrt::file_handle OpenFileHandle(const std::wstring &path, DWORD accessRights, bool openAsPlaceholder);
-    static void UpdateSyncStatus(const std::wstring &path);
```
dajimenezriv, line 17: We extract this function into it's own file as we are doing with all entry point functions.

packages/addon/native-src/placeholders_interface/Planceholders.cpp
```diff
@@ -37,30 +38,14 @@ winrt::file_handle Placeholders::OpenFileHandle(const std::wstring &path, DWORD
...
-    if (!fileHandle)
-    {
+    if (!fileHandle) {
         throw std::runtime_error("Failed to open file handle: " + std::to_string(GetLastError()));
     }
 
     return fileHandle;
 }
 
-void Placeholders::UpdateSyncStatus(const std::wstring &path)
```
dajimenezriv, line 48: As said we move this function to `update_sync_status_wrapper.cpp`.

PR #1149, opened 2025-11-23
Skip duplicate dehydrate file

dajimenezriv-internxt:
## What

1. When the user was clicking on “Free up space“ on the file explorer the dehydrate event was triggered twice. We fix it by checking the current blocks of the file to see if it’s already dehydrated.
2. We create the `package-lock.json` from 0 since we had 3 security issues that didn't have direct solutions, so instead of overrinding manually we let npm decide which versions to use for subdependencies.

### DIFF
--- a/src/apps/main/token-scheduler/token-scheduler.test.ts
+++ b/src/apps/main/token-scheduler/token-scheduler.test.ts
@@ -62,7 +62,7 @@ describe('Token Scheduler', () => {
     call(loggerMock.debug).toMatchObject({
       msg: 'Token renew date',
       expiresAt: new Date('1970-02-01'),
-      renewAt: new Date('1970-01-27'),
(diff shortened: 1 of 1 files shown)


src/apps/main/main.ts
```diff
@@ -70,6 +70,7 @@ const gotTheLock = app.requestSingleInstanceLock();
 
 if (!gotTheLock) {
   app.quit();
+  process.exit(0);
```
dajimenezriv, line 73: When running deeplinks it opens a second instance of electron. `app.quit()` quits the app but technically it doesn't end the execution. We do `process.exit(0)` so we don't continue the execution. I was checking because it was duplicating logs. With this everything works perfectly.

AlexisMora (approved):

AlexisMora (approved):

PR #1154, opened 2025-11-25
Cleanup legacy login

dajimenezriv-internxt:


src/apps/main/remote-sync/files/sync-remote-files.ts
```diff
@@ -21,10 +22,10 @@ export async function syncRemoteFiles({ self, from, offset = 0 }: TProps) {
      * However, if no checkpoint is provided it means that we don't have a local state yet.
      * In that situation, fetch only EXISTS files.
      */
-    const query = {
+    const query: GetFilesQuery = {
```
dajimenezriv, line 25: We have updated the swagger types, that's why we need to make these changes.

src/apps/main/types.ts
```diff
@@ -1,34 +1,4 @@
...
-    kyber: {
-      privateKey: string;
-      publicKey: string;
-    };
-  };
-};
+import { components } from '../shared/HttpClient/schema';
+
+type UserResponseDto = components['schemas']['UserResponseDto'];
+export type User = Omit<UserResponseDto, 'privateKey'> & { needLogout?: boolean; privateKey: string };
```
dajimenezriv, line 4: We need to omit the privateKey and then add it again because the type from swagger says that privateKey is Record<string, unknown>.

package.json
```diff
@@ -181,7 +181,6 @@
     "bytes": "^3.1.1",
     "chokidar": "^3.6.0",
     "clsx": "^2.1.1",
-    "crypto-js": "4.1.1",
```
sg-gs, line 184: Nice

PR #1156, opened 2025-11-26
Fix refresh token issue

dajimenezriv-internxt:
## What

There was an issue when a user didn’t open the app in some time and the refresh token was expired. Basically it was emitting the event of `USER_LOGGED_OUT` but we only listen to that event after the user has logged in. Since the token was expired we never entered inside the logged in block and we were never listeting to the `USER_LOGGED_OUT` to display the login window.

Solution. When the app opens we call `checkIfUserIsLoggedIn` that checks if we have the user object in `electron-store`. In case we don't have it we return false. In case we have it we perform a `refreshToken` that returns true or false based in the refresh was successful. If `checkIfUserIsLoggedIn` return false we display the login window. In case it returns `true` we start the login block (basically we set the token scheduler and we start all other services).

### DIFF
--- a/src/apps/main/token-scheduler/token-scheduler.test.ts
+++ b/src/apps/main/token-scheduler/token-scheduler.test.ts
@@ -52,7 +52,7 @@ describe('Token Scheduler', () => {
     call(loggerMock.error).toMatchObject({ msg: 'Error scheduling refresh token' });
   });
 
-  it('should not schedule if token is invalid', () => {
+  it('should schedule if token is valid', () => {
     // Given
     obtainTokenMock.mockReturnValue(createToken('31 day'));
     // When


sg-gs:
@dajimenezriv-internxt add to the description how you solved it and why is that robust enough.

src/apps/main/auth/handlers.ts
```diff
@@ -33,21 +34,17 @@ export function onUserUnauthorized() {
...
 
-  if (!isLoggedIn) return;
-
-  try {
-    await createTokenSchedule();
-  } catch (exc) {
-    if (exc instanceof RefreshTokenError) return;
-    else throw exc;
+  if (user.needLogout === undefined) {
+    logger.debug({ tag: 'AUTH', msg: 'User needs logout' });
```
sg-gs, line 43: Why being undefined means that needs logout?

src/apps/main/auth/handlers.ts
```diff
@@ -61,11 +58,15 @@ export function setupAuthIpcHandlers() {
 }
 
 export async function emitUserLoggedIn() {
+  const scheduler = new TokenScheduler();
```
sg-gs, line 61: This is the responsible of refreshing the token before it expires, correct?

dajimenezriv: Exactly

src/apps/main/token-scheduler/TokenScheduler.ts
```diff
@@ -1,66 +1,49 @@
...
-
-    if (expiration >= Date.now()) {
-      logger.warn({ msg: '[TOKEN] TOKEN IS EXPIRED' });
-      onUserUnauthorized();
-
-      return;
+  schedule() {
+    try {
+      const expirationDate = this.getExpiration();
+      const renewDate = expirationDate - DAYS_BEFORE * 24 * 60 * 60 * 1000;
```
sg-gs, line 21: Consider using dayjs someday, as dealing with dates manually could be problematic

src/apps/main/token-scheduler/TokenScheduler.ts
```diff
@@ -1,66 +1,49 @@
...
+
+      logger.debug({
+        tag: 'AUTH',
+        msg: 'Token renew date',
+        expiresAt: new Date(expirationDate),
+        renewAt: new Date(renewDate),
+      });
+
+      this.timeout = setTimeout(async () => {
+        const isRefreshed = await refreshToken();
```
sg-gs, line 32: Add a try catch here, otherwise an error refreshing the token could provoke an unhandled exception as the callback is not in the same context.

src/apps/main/auth/handlers.ts
dajimenezriv: Some months ago, like 7 months or so, we needed to force the logout. Basically we added that variable to the user object to check if we have already forced it. I don't remember very well why we did that because Jonathan did it, but we talked about it in that moment and it was necessary.

src/apps/main/token-scheduler/TokenScheduler.ts
```diff
@@ -1,66 +1,49 @@
...
+        if (isRefreshed) {
```
sg-gs, line 33: If not refreshed, which should be the behaviour?

src/apps/main/token-scheduler/TokenScheduler.ts
dajimenezriv: It cannot happen, the `refreshToken` function cannot throw an exception. Basically it uses the `infra` layer that returns { data, error }.

src/apps/main/auth/refresh-token.ts
```diff
@@ -1,35 +1,13 @@
...
-
-  const shceduler = new TokenScheduler(token);
-  const schedule = shceduler.schedule(refreshToken);
-
-  if (!schedule && !refreshedToken) {
-    logger.debug({ msg: 'Refreshing token' });
-    createTokenSchedule(await refreshToken());
+  if (data) {
+    updateCredentials({ newToken: data.newToken });
+    return true;
```
sg-gs, line 9: What does this flag indicate? Shouldn't this function return be void?

src/apps/main/token-scheduler/TokenScheduler.ts
dajimenezriv: We try to refresh when we start the app or with the scheduler 5 days before it expires. So, if it fails we ignore it for now since we can still use it for 5 days more and maybe it will be refreshed successfully in the next app start. In case it doesn't the first request that receives a 401 emits a `USER_LOGGED_OUT`.

src/apps/main/token-scheduler/TokenScheduler.ts
dajimenezriv: I'm not used to `dayjs`, so for me currently this is clearer. For example, I don't know if dayjs returns seconds, milliseconds or what. Also we have the tests to check that is done right.

src/apps/main/auth/refresh-token.ts
dajimenezriv: Has been the refresh token successful? Then true, otherwise false.

src/apps/main/token-scheduler/TokenScheduler.ts
sg-gs: dayjs hides those details, that's the thing of using it haha. Give it a check, I bet you will like it, pretty semantic.

PR #1158, opened 2025-11-27
Create addon tarball and spawn sync engine before fetch

dajimenezriv-internxt:
## What

This PR addresses some small fixes and improvements for the QA build.

1. Instead of referencing the addon package with a relative path we create the tarball. Why? Because otherwise everytime we reload the native deps (which happens frequently) it creates a new `addon.node` since we are not really installing a static addon. Now, we will have a static addon and we need to start managing versions.
2. Change in which order we spawn the sync engine worker and when we start the sync by checkpoint. We need the sync engine worker to happen before so it get's reads the sqlite before any new changes have been written.

### DIFF
--- a/src/apps/shared/logger/services/open-logs.test.ts
+++ b/src/apps/shared/logger/services/open-logs.test.ts
@@ -1,5 +1,4 @@
 import { openLogs } from './open-logs';
-import { readdir } from 'node:fs/promises';
 import { call, calls } from '@/tests/vitest/utils.helper.test';
 import { loggerMock } from '@/tests/vitest/mocks.helper.test';
 import { shell } from 'electron';
@@ -9,14 +8,12 @@ import archiver from 'archiver';
 import { mockDeep } from 'vitest-mock-extended';
 import { pipeline } from 'node:stream/promises';
 
-vi.mock(import('node:fs/promises'));
 vi.mock(import('node:fs'));
 vi.mock(import('node:stream/promises'));
 vi.mock(import('archiver'));
 
 describe('open-logs', () => {
   vi.mocked(pipeline);
-  const readdirMock = vi.mocked(readdir);
   const createWriteStreamMock = vi.mocked(createWriteStream);
   const archiverMock = vi.mocked(archiver);
   const openPathMock = vi.mocked(shell.openPath);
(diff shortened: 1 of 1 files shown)


src/apps/main/background-processes/sync-engine/services/spawn-sync-engine-worker.ts
```diff
@@ -57,12 +57,24 @@ export async function spawnSyncEngineWorker({ ctx }: TProps) {
...
+    const { logger, ...config } = ctx;
+    browserWindow.webContents.send('SET_CONFIG', config);
+
+    const manager = addRemoteSyncManager({ context: ctx });
 
     const worker: WorkerConfig = {
       ctx,
       browserWindow,
-      syncSchedule: scheduleSync({ ctx }),
+      syncSchedule: scheduleSync({ ctx, manager }),
```
dajimenezriv, line 77: It's the same code but now the `scheduleSync` is executed after we have loaded the worker.

src/apps/renderer/pages/Login/index.tsx
```diff
@@ -9,32 +10,36 @@ export function Login() {
...
-          variant="primary"
-          size="lg"
-          data-automation-id="buttonLogin"
-          onClick={() => openUrl('https://drive.internxt.com/login?universalLink=true')}
-          customClassName="font-medium w-7/12">
-          {t('login.signInBrowser')}
-        </Button>
-      </div>
+    <div className="relative flex h-screen flex-col">
+      <WindowTopBar />
```
dajimenezriv, line 14: We add this top bar so we have the close button.

src/apps/shared/logger/services/open-logs.ts
```diff
@@ -18,10 +18,7 @@ export async function openLogs() {
 
     const pipelinePromise = pipeline(archive, writeStream);
 
-    const logFiles = await readdir(PATHS.LOGS);
-    const logPaths = logFiles.filter((fileName) => fileName.endsWith('.log')).map((fileName) => join(PATHS.LOGS, fileName));
-
-    const paths = [...logPaths, PATHS.SQLITE_DB, PATHS.LOKIJS_DB];
+    const paths = [join(PATHS.LOGS, 'drive.log'), join(PATHS.LOGS, 'drive-important.log'), PATHS.SQLITE_DB, PATHS.LOKIJS_DB];
```
dajimenezriv, line 21: We don't want to bring old format logs, just `drive.log` and `drive-important.log` since otherwise in people that has big log files it takes too much to build the zip and we are not going to use them. (It doesn't throw an error if the file doesn't exist, it just ignores it).

PR #1159, opened 2025-11-27
Migrate lokijs to sqlite

dajimenezriv-internxt:
## What

Ticket https://inxt.atlassian.net/browse/PB-5430

This PR and the next one revert a decision I made 4 months ago.

First, we were obtaining the checkpoint from the file and folder tables (last updatedAt). However, this was a problem because we couldn't modify the database from the recovery sync and similar because it would break the checkpoint. So I decided to store the checkpoint in a separate place. However, typeorm is so strict and running migrations is a problem in a client app, also it couldn't be accessed from the sync engine process and since it's a native dependency (because of better-sqlite3) it cannot be run on tests.

So I decided to use LokiJS. An NoSQL database, that can be accessed from any process and can be used on tests. The plan was to migrate at some point everything to LokiJS, mainly because of the migrations.

Now, after 4 months I'm reverting it. Why?
- I discovered a new way of running migrations better to sqlite before loading the tables of typeorm and breaking the data.
- Having the database unaccesible from the renderer process was not so bad. This is because LokiJS is an in memory database with persistence every 5s, so it's not the same memory in both processes. Also, we don't know if another process is overwritting the json file.
- SQLite already works very well and it's solid.
- For debugging bugs, the sqlite format is much more useful and readable that NoSQL json.
- Valentyna was having problems because the checkpoint was not being updated because some issue with LokiJS.

Typeorm still has the problems of migrations (they are still a problem) and that cannot be run on tests, but it's a price to pay.

### DIFF
--- a/src/apps/utils/icon.ts
+++ b/src/apps/utils/icon.ts
@@ -2,4 +2,4 @@ import path from 'node:path';
 import { cwd } from 'node:process';
 
 export const iconPath =
-  process.env.NODE_ENV !== 'production' ? path.join(cwd(), 'assets', 'icon.ico') : path.join(process.resourcesPath, 'assets', 'icon.ico');
+  process.env.NODE_ENV === 'production' ? path.join(process.resourcesPath, 'assets', 'icon.ico') : path.join(cwd(), 'assets', 'icon.ico');


sg-gs:
This is a big one. Isn't there a SQLite in-memory tool to test the migration with a more realistic environment ? @dajimenezriv-internxt

dajimenezriv:
@sg-gs technically that didn't matter, because the important thing was that `better-sqlite3` cannot be run on tests. However, I've been researching again and found `sql.js`, which is a provider that is not native. Now, we have this provider for tests (and in memory database) and `better-sqlite3` for development and production.

sg-gs:
> @sg-gs technically that didn't matter, because the important thing was that `better-sqlite3` cannot be run on tests. However, I've been researching again and found `sql.js`, which is a provider that is not native. Now, we have this provider for tests (and in memory database) and `better-sqlite3` for development and production.

Is this not runnable by using a GitHub Action that: 
- Installs the dependencies
- Starts a Docker container with a SQLite instance and runs the migrations
- Runs those tests

dajimenezriv:
@sg-gs With this last change we are already running the tests in a Github Action against an in memory database. Maybe we can do what you say, but the complexity is too high, the tests will be slower also the tests will fails always when running on a local machine. Basically when we install `better-sqlite3` we need to rebuild it against electron node version, which maybe is not the same as our local node version and I wasn't able to rebuild it to the same version or similar solutions.

sg-gs:
> With this last change we are already running the tests in a Github Action against an in memory database

It works for me then

PR #1160, opened 2025-11-27
Migrate lokijs to sqlite 2

dajimenezriv-internxt:
## What

Ticket [inxt.atlassian.net/browse/PB-5430](https://inxt.atlassian.net/browse/PB-5430)

### DIFF
--- a/src/migrations/v2.5.6/move-checkpoint-to-lokijs.ts
+++ b/src/migrations/v2.5.6/move-checkpoint-to-lokijs.ts
@@ -1,0 +1,98 @@
+/*
+v2.6.3 Daniel Jiménez
+https://github.com/internxt/drive-desktop/pull/1159
+
+import { logger } from '@/apps/shared/logger/logger';
+import { CheckpointsModule } from '@/infra/lokijs/databases/checkpoints/checkpoints.module';
+import { fileRepository } from '@/infra/sqlite/services/drive-file';
+import { folderRepository } from '@/infra/sqlite/services/drive-folder';
+import { z } from 'zod';
+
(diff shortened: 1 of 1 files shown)


src/migrations/v2.5.6/move-checkpoint-to-lokijs.test.ts
sg-gs, line 1: The idea is not to remove the migration to precisely keep it as a history of the things that have been done in the past

src/migrations/v2.5.6/move-checkpoint-to-lokijs.ts
sg-gs, line 1: Same here

src/migrations/v2.5.6/move-checkpoint-to-lokijs.test.ts
dajimenezriv: The problem about keeping this migration is that it will imply to keep the dependency of `lokijs` forever and to initialize the database everytime so the migration can run, otherwise it will break. Also, this migration is not critical. Most users have their checkpoint already migrated because it was 4 months ago and in case it's not migrated, the checkpoint is undefined which means a sync from the beginning.

Apart from that, I'm not a big fan of keeping migrations forever. Since the migrations are running inside typescript it will imply having a code that no longer has a purpose (if the migration is 1 year old for example), that can lead to typescript errors and we have to mantain it.

sg-gs: Could we use another path to move the older, no longer used migrations so that is not included on the build and you can leave the history + remove lokijs ?

PR #1162, opened 2025-12-01
Fix download backups

dajimenezriv-internxt:
## What

1. Remove the secret `CRYPTO_KEY`. Now only the `NEW_CRYPTO_KEY` is used.
2. When there are many files being downloaded in backups the division was close to 0 so when we were doing `Math.trunc` we were getting a 0 and hiding the download progress. Now we do `Math.max(1, progress)`.
3. Create all folders in download backups so we create empty folders.
4. Add device name to downloaded backup.

### DIFF
--- a/src/node-win/addon-wrapper.ts
+++ b/src/node-win/addon-wrapper.ts
@@ -22,6 +22,7 @@ function parseAddonZod<T>(fn: keyof typeof addonZod, data: T) {
       tag: 'SYNC-ENGINE',
       msg: 'Error parsing addon result',
       fn,
+      data,
       error: result.error,
     });
   }


.erb/scripts/validate-process-env.ts
```diff
@@ -4,7 +4,6 @@ import { z } from 'zod';
 const ENV_SCHEMA = z.object({
   ANALYZE: z.union([z.literal('true'), z.literal('false')]),
   BRIDGE_URL: z.string(),
-  CRYPTO_KEY: z.string(),
```
sg-gs, line 7: Which is the difference between CRYPTO_KEY and the NEW one?

dajimenezriv: The `NEW_CRYPTO_KEY` is used for decrypting the name of files, folders and devices. The `CRYPTO_KEY` doesn't have any occurrences. Maybe it was used for the login? I don't know. The values were different.

dajimenezriv: Confirmed. It was used to hash the password.

sg-gs: Why should we remove it ? It is not the login broken after this change? This could be fatal if not approached entirely

dajimenezriv: Now for the login we have the SSO so we don't handle the user password.

PR #1163, opened 2025-12-01
Migrate universal link to http server

dajimenezriv-internxt:
## What

Universal link was giving issues in Opera in Windows, so we have decided to use the same approach as in cli (https://github.com/internxt/cli/pull/415/files#diff-0a4996726e042ea090d39c0640695b39e30333982974469ed8fbf7d6609c9286R69).


src/apps/main/electron/deeplink/process-login.ts
```diff
@@ -1,50 +1,42 @@
 import { DriveServerWipModule } from '@/infra/drive-server-wip/drive-server-wip.module';
```
dajimenezriv, line 1: We remove the try catch because the exception to reach the parent function.

src/apps/main/electron/deeplink/process-login.test.ts
```diff
@@ -20,12 +19,9 @@ describe('process-login', () => {
...
     refreshMock.mockResolvedValue({ data: { user: { uuid: 'uuid' } } });
     // When
-    await processLogin({ search: '?mnemonic=bW5lbW9uaWM=&newToken=bmV3VG9rZW4=&privateKey=cHJpdmF0ZUtleQ==' });
+    const promise = processLogin({ search: '?mnemonic=bW5lbW9uaWM=&newToken=bmV3VG9rZW4=&privateKey=cHJpdmF0ZUtleQ==' });
     // Then
-    call(loggerMock.error).toStrictEqual({
-      msg: 'Cannot process login deeplink',
-      error: new Error('Invalid mnemonic: mnemonic'),
-    });
+    await expect(promise).rejects.toThrowError('Invalid mnemonic: mnemonic');
```
sg-gs, line 24: Are you sure you want to only assert the message and not the error instance itself ?

sg-gs (approved):

src/apps/main/auth/open-login-url.ts
```diff
@@ -0,0 +1,44 @@
+import http from 'node:http';
+import { AddressInfo } from 'node:net';
+import { logger } from '@internxt/drive-desktop-core/build/backend';
+import { processLogin } from '../electron/deeplink/process-login';
+import { shell } from 'electron';
+import { DRIVE_WEB_URL } from '@/core/utils/utils';
+
+const HOST = '127.0.0.1';
+
+export let server: http.Server | undefined;
```
larryrider, line 10: check sonarcloud issues, this should be const

src/apps/main/electron/deeplink/process-login.test.ts
dajimenezriv: We have the eslint rule `no-throw-literal: error` so we can only throw erros. Anyway, I will change it for the next PR.

src/apps/main/auth/open-login-url.ts
dajimenezriv: The problem is because we are exporting a `let` variable because we are using it inside a test. We can wrap it inside a const store, but we lose a bit of readability in the code.

PR #1167, opened 2025-12-03
Throw error on create placeholder

dajimenezriv-internxt:
## What

1. Replace flag `CF_CREATE_FLAG_NONE` to `CF_CREATE_FLAG_STOP_ON_ERROR` since we want to throw an error in case the create placeholder fails.
2. Improve how we create the `packages/addon` tarball so it can be installed in just one command.


sg-gs:
@dajimenezriv-internxt should something be added to the README.md considering the package.json command changes? If not, merge

PR #1168, opened 2025-12-03
Improve delete item placeholders

dajimenezriv-internxt:
## What

1. Migrate `ProcessSyncContext` to `SyncContext` since it's the context used in the main process so it will allow the migration from sync process to the main process.
2. Make delete item placeholders more efficient and secure. First, we do all `await rm` inside a `Promise.all`. Second, instead of accessing to the placeholder state every time, we get it from the `InMemory` which already has that info. Third, check if the path matches, otherwise it means that we have an error. We shouldn't get that third case, but I'm adding the log just in case. Should we delete the path anyway? @sg-gs

### DIFF
--- a/src/apps/sync-engine/tests/create-placeholder.infra.test.ts
+++ b/src/apps/sync-engine/tests/create-placeholder.infra.test.ts
@@ -123,7 +123,7 @@ describe('create-placeholder', () => {
     calls(loggerMock.debug).toStrictEqual([
       { tag: 'SYNC-ENGINE', msg: 'Create sync root folder', code: 'NON_EXISTS' },
       { msg: 'Register sync root', rootPath },
-      { msg: 'Tree built', files: 0, folders: 0, trashedFiles: 0, trashedFolders: 0 },
+      { msg: 'Tree built', files: 0, folders: 1, trashedFiles: 0, trashedFolders: 0 },
       { tag: 'SYNC-ENGINE', msg: 'Load in memory paths', rootPath },
       { msg: 'onReady' },
       { msg: 'Create file', path: file },


src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
```diff
@@ -47,10 +49,9 @@ export class FolderPlaceholderUpdater {
     }
   }
 
-  static async run({ ctx, remotes, folders }: { ctx: ProcessSyncContext; remotes: ExtendedDriveFolder[]; folders: InMemoryFolders }) {
+  static async run({ ctx, remotes, folders }: { ctx: SyncContext; remotes: ExtendedDriveFolder[]; folders: InMemoryFolders }) {
     await Promise.all(
       remotes.map(async (remote) => {
-        if (remote.absolutePath === ctx.rootPath) return;
```
dajimenezriv, line 53: We are no longer adding the `rootPath` to the traverser so we don't need this check.

sg-gs:
If the error is because the path is no longer there, just ignore the error. Otherwise, log it, we do not know which cases could appear. Do you have a list of the possible errors that can happen?

src/context/virtual-drive/items/application/Traverser.test.ts
```diff
@@ -41,7 +41,7 @@ describe('Traverser', () => {
     const tree = await Traverser.run(props);
 
     expect(extractPaths(tree.files)).toStrictEqual(['/drive/file.txt']);
-    expect(extractPaths(tree.folders)).toStrictEqual(['/drive']);
+    expect(extractPaths(tree.folders)).toStrictEqual([]);
```
sg-gs: I get the point of removing the rootPath from the traverser but why is it useful overall ? Is it due to the root folder having a name is never displayed on the Virtual Drive?

dajimenezriv: We needed to have the rootPath in the traverser because sometimes we need to obtain the parent info. Now we are obtaining it with `NodeWin.getFolderInfo`. However, as I'm saying this it doesn't make sense to execute the `NodeWin.getFolderInfo` because we can obtain that for the traverser and it will be more efficient. I'm reverting this.

PR #1169, opened 2025-12-03
Create async wrapper with result

dajimenezriv-internxt:
## What

1. Improve the `async_wrapper.h` file in C++. Currently the `run_async` could only wrap functions that returned void. Now, we are able to wrap function with return values.
2. Move `getPlaceholderState` in C++ from sync to async.
3. Move `getPlaceholderState` in javascript from sync to async.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "init:ci": "npm ci --ignore-scripts && node node_modules/electron/install.js",
     "init:dev": "npm install && node node_modules/electron/install.js && npm run build:dll && npm run reload-native-deps",
     "========== Code style ==========": "",
-    "lint": "eslint src --ext .ts,.tsx --max-warnings 383",
+    "lint": "eslint src --ext .ts,.tsx --max-warnings 382",
     "lint:fix": "npm run lint -- --fix",
     "format": "prettier src tests --check",
     "format:fix": "prettier src tests --write",


packages/addon/include/helpers/async_wrapper.h
```diff
@@ -1,27 +1,44 @@
...
 
-template <typename... Args>
+template <typename R>
 struct AsyncWorkWrapper {
+    napi_async_work work;
+    napi_deferred deferred;
+    std::function<R()> fn;
+    std::string error;
+    bool success;
+    R result;
```
dajimenezriv, line 18: Same but with `R result at the end`.

packages/addon/include/helpers/async_wrapper.h
```diff
@@ -1,27 +1,44 @@
...
-template <typename... Args>
+template <typename R>
 void execute_work(napi_env env, void* data)
 {
-    auto* asyncWork = static_cast<AsyncWorkWrapper<Args...>*>(data);
+    auto* asyncWork = static_cast<AsyncWorkWrapper<R>*>(data);
 
     try {
-        asyncWork->fn();
+        if constexpr (std::is_void_v<R>) {
```
dajimenezriv, line 36: We check here and in other places if the return type `R` is void. In that case we don't need to store the return value in `asyncWork->result`.

packages/addon/include/helpers/async_wrapper.h
```diff
@@ -53,20 +76,26 @@ void complete_work(napi_env env, napi_status status, void* data)
 template <typename Func, typename... Args>
 napi_value run_async(napi_env env, const char* resource_name, Func&& fn, Args&&... args)
 {
+    using R = std::invoke_result_t<Func, Args...>;
```
dajimenezriv, line 79: Get the return type.

packages/addon/include/placeholders_interface/Placeholders.h
```diff
@@ -10,6 +11,25 @@ struct FileState {
     CF_PIN_STATE pinState;
 };
 
+template <>
+struct NapiSerializer<FileState> {
```
dajimenezriv, line 15: We need to add a serializer for each return type. `getPlaceholderState` returns a `FileState`. It's the same code as in below file, however here we have replaced `state.placeholderId.length()` with `NAPI_AUTO_LENGTH` to make it consistent with other parts of the C++ code. In javascript the result is the same.

egalvis27: Is there a way to handle this similarly to how function parameters are serialized, to avoid having to create one for each function with a return value, or am I misunderstanding how it works?

dajimenezriv: Maybe, but for now it doesn't matter. We have just 2 functions that return value,.

PR #1170, opened 2025-12-04
Simplify logs and use map in backups

dajimenezriv-internxt:
## What

1. Remove `Create file placeholder`, `Create folder placeholder` and `Convert to placeholder` logs since they take too much space in the logs and we can more or less obtain the state of those items from the sqlite.
2. Migrate backups tree from Record to Map. This is because when we access a property in a record it doesn't mark it as undefined, but with `Map.get` it does. Basically we were getting so many log errors for this. We can change the rule in tsconfig to return undefined with Record, but it will break too much code.
3. Remove a log that was happening when `FolderNotFound` was thrown.


src/apps/backups/folders/create-folders.ts
```diff
@@ -19,46 +18,39 @@ type TProps = {
...
-    const parent = tree.folders[parentPath];
-
-    if (!parent) {
-      logger.error({
-        tag: 'BACKUPS',
-        msg: 'Parent folder does not exist',
-        path: localFolder.absolutePath,
-      });
-
-      context.addIssue({ error: 'CREATE_FOLDER_FAILED', name: localFolder.absolutePath });
```
dajimenezriv, line 41: It's the same code except for this. If the parent does not exist it means that it had an error when creating, we don't need to log an error and add an issue for every child.

src/context/local/localFile/application/update/FileBatchUpdater.ts
```diff
@@ -16,59 +16,38 @@ type Props = {
...
-        }),
-      ),
-    );
-  }
+export async function replaceFiles({ self, context, tracker, modified }: Props) {
+  await Promise.all(
+    modified.map(async ({ local, remote }) => {
+      await replaceFile({ context, localFile: local, file: remote });
+      self.backed++;
+      tracker.currentProcessed(self.backed);
```
dajimenezriv, line 24: Same code, but we move the `self.backed` and `tracker.currentProcessed` here so we don't skip them because of doing `return` in the below function. Technically it didn't because this was inside a `finally` block, but it's not so clear.

src/context/local/localFile/application/upload/FileBatchUploader.ts
```diff
@@ -18,49 +18,52 @@ type Props = {
...
-  static async run({ self, context, tracker, remoteTree, added }: Props) {
-    const promises = added.map(async (localFile) => {
-      try {
-        const contentsId = await uploadFile({ context, localFile });
+export async function createFiles({ self, context, tracker, remoteTree, added }: Props) {
+  await Promise.all(
+    added.map(async (localFile) => {
+      await createFile({ context, localFile, remoteTree });
+      self.backed++;
+      tracker.currentProcessed(self.backed);
```
dajimenezriv, line 26: Same code, but we move the `self.backed` and `tracker.currentProcessed` here so we don't skip them because of doing `return` in the below function. Technically it didn't because this was inside a `finally` block, but it's not so clear.

src/context/virtual-drive/files/application/FileCreator.ts
```diff
@@ -19,47 +17,34 @@ type Props = {
...
-      if (error) throw error;
-
-      return fileDto;
-    } catch (error) {
-      logger.error({
-        tag: 'SYNC-ENGINE',
-        msg: 'Error in file creator',
-        path,
-        exc: error,
-      });
```
dajimenezriv, line 58: Here we were logging many times the expected exception of `FolderNotFoundError`. However, we don't really need this catch. We are already catching `create-file.ts` and handling there the `FolderNotFoundError` or an unexpected exception.

PR #1172, opened 2025-12-05
Improve addon error handling

dajimenezriv-internxt:
## What

1. When we moved to async functions in C++, the wrapper was not using the format exception wrapper that we created for the sync functions. Now, we start using it so we don't get `Unknown exception`.
2. Improve error handling on `getFilePlaceholder` or `getFolderPlacholder` to check when the error is `NOT_A_PLACEHOLDER` or it's another one.


packages/addon/include/helpers/napi_safe_wrap.h
```diff
@@ -7,23 +7,29 @@
...
 
-template <typename Fn>
-napi_value napi_safe_wrap(napi_env env, napi_callback_info info, Fn&& fn, const char* function_name)
+inline std::string format_exception_message(const char* function_name)
 {
-    std::ostringstream oss;
-
     try {
-        return fn(env, info);
+        throw;
```
dajimenezriv, line 13: This throws again the last exception thrown.

src/node-win/tests/register-sync-root.infra.test.ts
```diff
@@ -0,0 +1,13 @@
+import { abs } from '@/context/local/localFile/infrastructure/AbsolutePath';
+import { Addon } from '../addon-wrapper';
+
+describe('register-sync-root', () => {
+  it('should throw error if path is too short', async () => {
```
sg-gs, line 5: _if the path is empty_ 

Could be a more semantic legend

dajimenezriv: This test is more to know that we are adding the error message in the error than to check anything in particular, but yes, we can rename it like that in a future PR. Anyway, we will implement more use cases for all C++ functions to make it more solid.

PR #1174, opened 2025-12-05
Join drive and backup persist file and folder

dajimenezriv-internxt:
## What

1. Add `PARENT_NOT_FOUND` error to create folder request.
2. Join create file and folder requests to be exactly the same in drive and backups. Also, now all these requests are done in the main process.

### DIFF
--- a/src/backend/features/remote-sync/update-in-sqlite/create-or-update-file.ts
+++ b/src/backend/features/remote-sync/update-in-sqlite/create-or-update-file.ts
@@ -14,14 +14,14 @@ export async function createOrUpdateFile({ ctx, fileDto }: { ctx: CommonContext;
   });
 }
 
-export async function createOrUpdateFiles({ context, fileDtos }: { context: CommonContext; fileDtos: FileDto[] }) {
+export async function createOrUpdateFiles({ ctx, fileDtos }: { ctx: CommonContext; fileDtos: FileDto[] }) {
   return await SqliteModule.FileModule.createOrUpdateBatch({
     files: fileDtos.map((fileDto) => ({
       ...fileDto,
       size: Number(fileDto.size),
       isDangledStatus: false,
-      userUuid: context.userUuid,
-      workspaceId: context.workspaceId,
+      userUuid: ctx.userUuid,
+      workspaceId: ctx.workspaceId,
     })),
   });
 }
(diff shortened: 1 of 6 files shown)


src/context/virtual-drive/files/infrastructure/HttpRemoteFileSystem.ts
```diff
@@ -1,65 +1,41 @@
...
-      ...offline,
-      bucket: ctx.bucket,
-      workspaceId: ctx.workspaceId,
-    };
-
-    const { data, error } = await HttpRemoteFileSystem.create(props);
-
-    if (data) return data;
-
-    if (error.code === 'FOLDER_NOT_FOUND') {
```
dajimenezriv, line 52: We don't need this anymore. This was restoring the parent folder in case it didn't exist when performing the create file request. However, technically we cannot reach this place because we are obtaining the parent uuid from the file explorer so the parent then needs to exist.

src/backend/features/remote-sync/update-in-sqlite/create-or-update-file.ts
```diff
@@ -1,21 +1,20 @@
...
-      userUuid: context.userUuid,
-      workspaceId: context.workspaceId,
+      userUuid: ctx.userUuid,
+      workspaceId: ctx.workspaceId,
     },
   });
 }
 
-export async function createOrUpdateFiles({ context, fileDtos }: { context: Config; fileDtos: FileDto[] }) {
+export async function createOrUpdateFiles({ context, fileDtos }: { context: CommonContext; fileDtos: FileDto[] }) {
```
egalvis27: Since you changed context above, you could change it here too.

src/backend/features/remote-sync/update-in-sqlite/create-or-update-folder.ts
```diff
@@ -1,8 +1,18 @@
...
+  return await SqliteModule.FolderModule.createOrUpdate({
+    folder: {
+      ...folderDto,
+      userUuid: ctx.userUuid,
+      workspaceId: ctx.workspaceId,
+    },
+  });
+}
+
+export async function createOrUpdateFolders({ context, folderDtos }: { context: CommonContext; folderDtos: FolderDto[] }) {
```
egalvis27: Since you changed context above, you could change it here too.

PR #1175, opened 2025-12-05
Join replace file from drive and backups

dajimenezriv-internxt:
## What

1. Create the `persistReplaceFile` function in the main process that it will be shared for drive and backups.
2. Simplify the create file and folder. Now we have created a function called `getParentUuid` that either returns the parent uuid in case it's already a placeholder or calls the folder creator with the parent path and returns the uuid.


src/apps/sync-engine/callbacks-controllers/controllers/add-controller.ts
```diff
@@ -1,38 +1,35 @@
...
 import { SyncModule } from '@internxt/drive-desktop-core/build/backend';
 import { ipcRendererSyncEngine } from '../../ipcRendererSyncEngine';
+import { FileCreator } from '@/context/virtual-drive/files/application/FileCreator';
 
 export class AddController {
   static async createFile({ ctx, path, stats }: { ctx: ProcessSyncContext; path: AbsolutePath; stats: Stats }) {
-    logger.debug({ msg: 'Create file', path });
+    ctx.logger.debug({ msg: 'Create file', path });
 
-    try {
```
dajimenezriv, line 14: We don't need this try catch here.

src/context/virtual-drive/files/application/FileCreator.ts
```diff
@@ -1,43 +1,39 @@
...
   ctx: ProcessSyncContext;
   path: AbsolutePath;
-  contents: {
-    id: ContentsId;
-    size: number;
-  };
+  size: number;
 };
 
 export class FileCreator {
```
dajimenezriv, line 14: Now, we have joined here all file creation operations.
1. Upload contents.
2. Get parent uuid (using the below file `get-parent-uuid.ts`).
3. Persist file in wip and sqlite.
4. Convert file to placeholder.

src/context/virtual-drive/files/application/get-parent-uuid.ts
```diff
@@ -0,0 +1,20 @@
...
+import { AbsolutePath } from '@internxt/drive-desktop-core/build/backend';
+import { FolderCreator } from '../../folders/application/FolderCreator';
+import { FolderUuid } from '@/apps/main/database/entities/DriveFolder';
+
+type Props = {
+  path: AbsolutePath;
+  ctx: SyncContext;
+};
+
+export async function getParentUuid({ ctx, path }: Props): Promise<FolderUuid> {
```
dajimenezriv, line 13: Here we have two options. Either that parent folder is already a placeholder a we can obtain the parent uuid. Or the parent folder is not a placeholder yet so we return the `FolderCreator.run` but with the parent path.

PR #1176, opened 2025-12-05
Simplify has to be moved checked

dajimenezriv-internxt:
## What

1. Remove lodash.debounce since we are already checking inside the `updateRemoteSync` that it's not running (https://github.com/internxt/drive-desktop/blob/37728eacfdd7cd40eb0a5e47075fab2ec2c6e25d/src/apps/main/remote-sync/handlers.ts#L36). Then this was only causing the webhooks to wait 5s before fetching the changes and updating the file explorer.
2. Remove the root folder for the drive traverser. In the end we don't need it, since we are never accessing to the parent path in the tree.
3. Open downloaded folder when download a backup (https://inxt.atlassian.net/browse/PB-5512).
4. Improve the check for knowing if a placeholder has to be moved locally because it has been moved in remote (https://inxt.atlassian.net/browse/PB-5508). Previously we were calling C++ code, but it's not necessary.
5. Use a `Map` instead of a `Record` when retrieving the local files and folder so we obtain the `undefined` when doing `Map.get`.

### DIFF
--- a/src/backend/features/remote-sync/file-explorer/update-folder-placeholder.test.ts
+++ b/src/backend/features/remote-sync/file-explorer/update-folder-placeholder.test.ts
@@ -23,7 +23,7 @@ describe('update-folder-placeholder', () => {
   let props: Parameters<typeof FolderPlaceholderUpdater.update>[0];
 
   beforeEach(() => {
-    validateWindowsNameMock.mockResolvedValue({ isValid: true });
+    validateWindowsNameMock.mockReturnValue({ isValid: true });
 
     props = mockProps<typeof FolderPlaceholderUpdater.update>({
       folders: new Map([['uuid' as FolderUuid, { path: 'localPath' as AbsolutePath }]]),
@@ -38,7 +38,7 @@ describe('update-folder-placeholder', () => {
 
   it('should do nothing if name is invalid', async () => {
     // Given
-    validateWindowsNameMock.mockResolvedValue({ isValid: false });
+    validateWindowsNameMock.mockReturnValue({ isValid: false });
     // When
     await FolderPlaceholderUpdater.update(props);
     // Then


src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
```diff
@@ -51,11 +51,7 @@ export class FolderPlaceholderUpdater {
   }
 
   static async run({ ctx, remotes, folders }: { ctx: SyncContext; remotes: ExtendedDriveFolder[]; folders: InMemoryFolders }) {
-    await Promise.all(
-      remotes.map(async (remote) => {
-        if (remote.absolutePath === ctx.rootPath) return;
```
dajimenezriv, line 56: We no longer have the root folder in the traverser.

src/backend/features/remote-sync/file-explorer/has-to-be-moved.test.ts
```diff
@@ -1,95 +1,60 @@
...
-  /**
-   * v2.5.6 Daniel Jiménez
-   * This is a use case that we cannot handle right now. Basically we cannot rename
-   * an item if the move event is in another folder because we cannot move an inner
-   * item. What will happen is that in the first sync iteration we will call move from
-   * folder1 to folder3 and in the second sync iteration now the path will be the same
-   * so we will rename the inner folder.
-   */
-  it('should return false if item has been renamed but both parents have the same uuid', async () => {
+  it('should return false if gran parent is moved', () => {
```
sg-gs: a d is missing here in gran

src/backend/features/remote-sync/file-explorer/has-to-be-moved.ts
```diff
@@ -1,27 +1,24 @@
...
 
-  const { data: remoteParentInfo } = await NodeWin.getFolderInfo({ ctx, path: remoteParentPath });
-  const { data: localParentInfo } = await NodeWin.getFolderInfo({ ctx, path: localParentPath });
+  remoteParentPath = dirname(remoteParentPath);
+  localParentPath = dirname(localParentPath);
 
-  if (!remoteParentInfo || !localParentInfo) return false;
-
-  const isMoved = remoteParentInfo.uuid !== localParentInfo.uuid;
+  const isMoved = remoteParentPath === localParentPath;
```
sg-gs: Why not !==?

dajimenezriv: In the end I had to revert most things since when reviewing the PR I discovered that just using the path or parent path is not enough. Basically we can't know just using the path if a file has been moved to another folder or if the parent folder has been renamed. We still need to use the parent uuid. However, we already have the parent uuid in the remote item, so we don't need to obtain it.

PR #1178, opened 2025-12-09
Improve sync remote changes to local

dajimenezriv-internxt:
## What

Previously when a file was modified in remote we were deleting the local item and creating the placeholder again. This is working when using the `usePolling` property of the chokidar watcher. However, since we want to disable it because of cpu problems (https://github.com/dotansimha/graphql-code-generator/issues/1796) this can cause the watcher to take the event of unlink the file and remove it in remote.

So, instead of deleting and recreating, we are going to dehydrate it and update the placeholder with the new size, which is safer. For that we need 2 new functions in C++: `set_pin_state` and `update_placeholder`.

### DIFF
--- a/src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.infra.test.ts
+++ b/src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.infra.test.ts
@@ -61,7 +61,7 @@ describe('sync-remote-changes-to-local', () => {
     await sleep(3000);
 
     // Then
-    calls(onAllMock).toMatchObject([
+    calls(onAllMock).toStrictEqual([
       { event: 'add', path },
       { event: 'change', path },
     ]);
--- a/src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.test.ts
+++ b/src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.test.ts
@@ -12,8 +12,8 @@ describe('sync-remote-to-local', () => {
   partialSpyOn(Addon, 'setPinState');
   const updatePlaceholderMock = partialSpyOn(Addon, 'updatePlaceholder');
 
-  const localDate = new Date('1999-01-01T00:00:00.000Z');
-  const remoteDate = new Date('2025-01-01T00:00:00.000Z');
+  const localDate = new Date('2000-01-01');
+  const remoteDate = new Date('2000-01-02');
   let props: Parameters<typeof syncRemoteChangesToLocal>[0];
 
   beforeEach(() => {
@@ -35,9 +35,9 @@ describe('sync-remote-to-local', () => {
     });
   });
 
-  it('should not sync when remote file is older', async () => {
+  it('should not sync when local file is older', async () => {
     // Given
-    props.local.stats.mtime = new Date('2026-01-01T00:00:00.000Z');
+    props.local.stats.mtime = new Date('2000-01-03');
     // When
     await syncRemoteChangesToLocal(props);
     // Then


packages/addon/native-src/placeholders_interface/Planceholders.cpp
```diff
@@ -46,24 +46,6 @@ winrt::file_handle Placeholders::OpenFileHandle(const std::wstring& path, DWORD
     return fileHandle;
 }
 
-void Placeholders::UpdateFileIdentity(const std::wstring& path, const std::wstring& placeholderId)
```
dajimenezriv, line 49: This was the old implementation of the new `updatePlaceholder` but it was not used in javascript, so maybe it was broken.

packages/addon/native-src/virtual_drive/dehydrate_file.cpp
```diff
@@ -17,7 +17,7 @@ void dehydrate_file(const std::wstring& path)
     LARGE_INTEGER offset;
     offset.QuadPart = 0;
     LARGE_INTEGER length;
-    GetFileSizeEx(fileHandle.get(), &length);
+    length.QuadPart = -1;
```
dajimenezriv, line 20: Setting as -1 dehydrates the whole file, we don't need to manually calculate the length.

packages/addon/native-src/main.cpp
```diff
@@ -1,26 +1,31 @@
-#include <iterator>
 #include <node_api.h>
+
+#include <iterator>
+
 #include "Wrappers.h"
 
 napi_value init(napi_env env, napi_value exports)
```
dajimenezriv, line 7: I just sorted them alphabetically and added these 2 new functions.

packages/addon/include/virtual_drive/Wrappers.h
```diff
@@ -2,15 +2,17 @@
...
-napi_value RegisterSyncRootWrapper(napi_env env, napi_callback_info args);
-napi_value GetRegisteredSyncRootsWrapper(napi_env env, napi_callback_info args);
-napi_value ConnectSyncRootWrapper(napi_env env, napi_callback_info args);
-napi_value DisconnectSyncRootWrapper(napi_env env, napi_callback_info args);
-napi_value UpdateSyncStatusWrapper(napi_env env, napi_callback_info args);
-napi_value GetPlaceholderStateWrapper(napi_env env, napi_callback_info args);
-napi_value ConvertToPlaceholderWrapper(napi_env env, napi_callback_info args);
-napi_value DehydrateFileWrapper(napi_env env, napi_callback_info args);
-napi_value HydrateFileWrapper(napi_env env, napi_callback_info args);
+napi_value ConnectSyncRootWrapper(napi_env env, napi_callback_info info);
```
dajimenezriv, line 5: I just sorted them alphabetically and added these 2 new functions.

src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.ts
```diff
@@ -12,44 +12,26 @@ type Props = {
...
-        creationTime: new Date(remote.createdAt).getTime(),
-        lastWriteTime: new Date(remote.updatedAt).getTime(),
-      });
-
-      ctx.logger.debug({
-        msg: 'File successfully synced from remote to local',
-        path: remote.absolutePath,
-        newSize: remote.size,
-      });
+      await Addon.setPinState({ path, pinState: PinState.Unspecified });
```
dajimenezriv, line 29: We want to ensure that the `pinState` is not `AlwaysLocal`, because in that case we cannot dehydrate the file.

src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.infra.test.ts
```diff
@@ -64,25 +61,24 @@ describe('sync-remote-changes-to-local', () => {
     await sleep(3000);
 
     // Then
-    calls(onAllMock).toStrictEqual([
-      { event: 'add', path: filePath },
-      { event: 'change', path: filePath },
+    calls(onAllMock).toMatchObject([
```
sg-gs: Why?

src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.test.ts
```diff
@@ -1,91 +1,71 @@
...
-      placeholderId: 'FILE:uuid',
-      size: 1024,
-      creationTime: time,
-      lastWriteTime: time,
-    });
+    calls(updatePlaceholderMock).toHaveLength(0);
   });
 
-  it('should not sync when remote file is not newer', async () => {
+  it('should not sync when file sizes are equal', async () => {
```
sg-gs, line 47: You should check at least the updated_at file also. Same sized-files with diff content will be accepted, otherwise. Or does this only apply to remote-only placeholders?

sg-gs (commented):
> So, instead of deleting and recreating, _we are going to dehydrate it and update the placeholder with the new size, which is safer_. For that we need 2 new functions in C++: set_pin_state and update_placeholder.

Is this safe? Was some important placeholder metadata used for checking sync things, changing before this modification? And is this criteria valid for all the cases or allows for false positives?

src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.test.ts
dajimenezriv: I'm not sure if I understood your question. We are already checking the size and updatedAt. Technically files with the same size and different modification time should be affected. The problem is that we weren't updating the modification time in remote with the `replaceFile` so using just the modification time will cause many local files to be dehydrated (because the modification time is an old one). Also, the `updatedAt` can be modified because of other things so it's not reliable. Currently the best we can do is use `size` and check that remote `updatedAt` is newer than local modification time (which we already do).

dajimenezriv:
@sg-gs the old implementation was safe. We were debouncing the event in change for 2s, and since the `usePolling` was obtaining the `stats` every 100ms, so the unlink+add was always considered as a modification. The new implementation is even safer. We never delete the placeholder, we just dehydrate, but the content is already in remote. Also we don't use any placeholder metadata, just the placeholder id (which is going to be the same) and the pin state. Not sure if that's what you meant?

sg-gs:
I meant that if you are not replacing the placeholder, only the size, what happens with the content?

dajimenezriv:
@sg-gs We have dehydrated it (removed the content locally), the content is no longer there (https://github.com/internxt/drive-desktop/pull/1178/files#diff-ef47047a026325ae971ff4da7d78a2023a07a84479cc10b0e68e55efd0349dd0R23)

src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.test.ts
sg-gs: > Currently the best we can do is use size and check that remote updatedAt is newer than local modification time (which we already do).

Yeah, this is what I am talking about. It should be done. But together. The legend of this test suggests that only checking the size is enough

dajimenezriv: But that's how we write test legends. We start with the premise of the successful/unsuccessful test in which we don't write a `// Given` statement because the props satisfy it (https://github.com/internxt/drive-desktop/blob/2be8e078b80237df54b952245a7b76e833552d74/src/backend/features/remote-sync/file-explorer/sync-remote-changes-to-local.test.ts#L56). Then, we write a different test with each case that breaks that successful/unsuccessful test. Do you have any reference or doc or external page about how you structure tests?

sg-gs: > we write a different test with each case

Yes, that's the thing. This case says (and per the code, suggests also) that: "if the file sizes are equal, this item does not need to sync". What I am saying is: is this reliable? 

And probably, considering the context, the syncRemoteChangesToLocal is probably checking the updated_at also, but the legend of your test does not suggest so. I am highlighting that this legend is confusing, not by its format, by what it says

dajimenezriv: But it's true. If the sizes are equal, the file is never uploaded. If the sizes are equal we don't care about the updatedAt.

PR #1180, opened 2025-12-10
Remove watcher use polling

dajimenezriv-internxt:
## What

1. Remove `usePolling` from watcher. This changes a bit how we handle the `change` events. Now we don't receive the previous and current stats, now we just received the path. So we need to obtain the stats from that path and check if it has been modified (`mtime`) or the metadata has changed (`ctime`) by comparing against `Date.now()`. Also, now when we create a placeholder it also emits a change event (it emits the `add` event and a new `change` event that we just ignore).
2. For some reason, the register sync root sometimes is failing when executing the infra tests in parallel. I just disabled parallelism for now since I couldn't find a solution yet. This error is not happening neither in local tests nor in the app running.


packages/addon/include/helpers/async_wrapper.h
```diff
@@ -17,17 +17,7 @@ struct AsyncWorkWrapper {
     std::string error;
```
dajimenezriv, line 17: Same code but a bit simplified.

packages/addon/include/helpers/napi_extract_args.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <node_api.h>
+#include <stdafx.h>
```
sg-gs, line 4: What is this new import doing?

dajimenezriv: Doing tests about the `register_sync_root` error I was not importing `napi_extract_args.h` in the `async_wrapper.h`. Since `async_wrapper.h` imported <stdafx.h>, `napi_extract_args.h` was already having that library. However, technically `napi_extract_args.h` needs <stdafx.h> because it's using the constant CF_PIN_STATE. It can be deleted or kept, because in C++ if you import in file1 and then that file is imported in file2, the imports from file1 work on file2.

PR #1184, opened 2025-12-11
Remove sonarcloud issue from sync walk

dajimenezriv-internxt:
## What

1. Reduce complexity of `process-item.ts` function. We split it into 3 functions.
2. Change `sync-wak.ts` to accept a `onError` function instead of returning the error in the return value.


src/context/local/localTree/infrastructure/FsLocalItemsGenerator.ts
```diff
@@ -14,40 +14,30 @@ type LocalFolderDTO = {
...
-  switch (code) {
-    case 'NON_EXISTS':
-      return 'ROOT_FOLDER_DOES_NOT_EXIST';
-    case 'NO_ACCESS':
-      return 'ROOT_FOLDER_DOES_NOT_EXIST';
-    default:
-      return code;
-  }
-}
+function parseStatError({ context, path, error }: { context: BackupsContext; path: AbsolutePath; error: StatError }) {
```
egalvis27: The definition of the parameters seems a bit large; you could move them to a type object.

PR #1187, opened 2025-12-12
Refactor traverser to wait for parent folder

dajimenezriv-internxt:
## What

Since we introduced async operations when create file and folder placeholder we could run into a bug that was causing to processs childreen placeholder before parent placeholder, so if for example we were trying to create a /parent/child but the /parent didn't exist we were receiving an error. Now, the traverser creates the placeholders and if the parent creation fails then it doesn't continue. Previously if the parent failed (even before moved to async operations) all childreen were tried and displaying an error.

We have included also the `checkDangledFiles` here.

We have reduced also the recovery sync from 60s to 30s to make it a bit faster.

### DIFF
--- a/src/backend/features/remote-sync/file-explorer/update-file-placeholder.ts
+++ b/src/backend/features/remote-sync/file-explorer/update-file-placeholder.ts
@@ -35,7 +35,7 @@ export class FilePlaceholderUpdater {
 
       if (isMoved) {
         ctx.logger.debug({
-          msg: 'Moving folder placeholder',
+          msg: 'Moving file placeholder',
           remotePath,
           localPath,
         });


src/apps/sync-engine/BindingManager.ts
```diff
@@ -42,7 +42,7 @@ export class BindingsManager {
      * This one is for the first case, since maybe the sync engine failed in a previous fetching
      * and we have some placeholders pending from being created/updated/deleted
      */
-    await refreshItemPlaceholders({ ctx });
+    await refreshItemPlaceholders({ ctx, runDangledFiles: true });
```
dajimenezriv, line 45: We just run dangled files in the first `refreshItemPlaceholders`.

src/context/virtual-drive/items/application/Traverser.ts
```diff
@@ -1,62 +1,60 @@
...
-      files: [],
-      folders: [],
-      trashedFiles: [],
-      trashedFolders: [],
-    };
-
-    this.traverse(tree, items, { absolutePath: ctx.rootPath, uuid: ctx.rootUuid });
 
-    return tree;
+    await Promise.all([filePromises, folderPromises]);
```
dajimenezriv, line 58: I don't know why these sonar issues appear, since they are promises.

src/backend/features/remote-sync/file-explorer/update-file-placeholder.ts
```diff
@@ -35,7 +35,7 @@ export class FilePlaceholderUpdater {
 
       if (isMoved) {
         ctx.logger.debug({
-          msg: 'Moving file placeholder',
+          msg: 'Moving folder placeholder',
```
sg-gs: Why? the function is for a file

dajimenezriv: Missclick. I was extracting the `check-if-moved` but the PR size was bigger than 500 so I had to split it and I forgot that word here. Fixed. It was fixed anyway in the next PR (https://github.com/internxt/drive-desktop/pull/1188/changes#diff-619a5b736a7c07396edf630b6a00ce9e32518d4757dd5da75c879cf586f55314).

src/apps/sync-engine/dangled-files/check-dangled-files.ts
```diff
@@ -1,33 +1,32 @@
...
 
-export async function checkDangledFiles({ ctx }: { ctx: ProcessSyncContext }) {
-  const { files } = await Traverser.run({ ctx });
+type Props = {
+  ctx: ProcessSyncContext;
+  file: ExtendedDriveFile;
+};
 
-  ctx.logger.debug({ msg: 'Checking dangled files', total: files.length });
+export async function checkDangledFiles({ ctx, file }: Props) {
```
sg-gs, line 15: Shouldn't this name be `checkDangledFile`?

dajimenezriv: Yes, but that file already exists, also I was in the limit of the PR size and since this function is temporal and it will be eventually deleted, I decided to keep it like this.

src/context/virtual-drive/items/application/Traverser.ts
```diff
@@ -1,62 +1,60 @@
...
 
       if (folder.status === 'DELETED' || folder.status === 'TRASHED') {
-        tree.trashedFolders.push(extendedFolder);
+        await deleteItemPlaceholder({ ctx, type: 'folder', remote, locals: folders });
       } else {
-        tree.folders.push(extendedFolder);
-        this.traverse(tree, items, extendedFolder);
+        const success = await FolderPlaceholderUpdater.update({ ctx, remote, folders });
+        if (success) {
+          await this.run({ ctx, items, files, folders, currentFolder: remote, runDangledFiles });
```
sg-gs, line 53: Have you sized this recursive call? I mean, how many times can this be called? Remember that this can cause a stack overflow if too many calls are done. This could perfectly be the reason of the high CPU/high RAM for customers with too much items

src/apps/sync-engine/dangled-files/check-dangled-files.ts
dajimenezriv: It's the same with `delete-placeholders`, now it should be renamed `delete-placeholder`, but for this case I renamed it in the next PR: https://github.com/internxt/drive-desktop/pull/1188/changes

src/context/virtual-drive/items/application/Traverser.ts
dajimenezriv: It will be called as much deep as a folder has. It's not a big deal. I think that we discovered the cause of high CPU usage last week. It's explained here: https://inxt.atlassian.net/browse/PB-5511. Hopefully this will be the first release in which we won't have high CPU usage.

dajimenezriv: High RAM is more related about how many processes we have, and we can remove all sync processes by next release, I already have some draft work.

sg-gs: > It will be called as much deep as a folder has

This would be still an issue for deep folders, why not translating it to an iterative approach?

dajimenezriv: Let me check it for next PR

src/context/virtual-drive/items/application/Traverser.ts
sg-gs: I think you should flatten or use ... as Promise.all expects an array of promises, not an array of arrays of Promises

dajimenezriv: That makes sense, I had this problem many times. Got it

PR #1191, opened 2025-12-15
Move watcher to main process

dajimenezriv-internxt:
## What

1. Move watcher to main process.
2. Move add pending items to main process. This is the function that runs on each workspace startup to check items that should be uploaded or hydrated.

### DIFF
--- a/src/apps/sync-engine/tests/create-placeholder.infra.test.ts
+++ b/src/apps/sync-engine/tests/create-placeholder.infra.test.ts
@@ -49,6 +49,11 @@ describe('create-placeholder', () => {
     await sleep(5000);
 
     // Then
+    calls(onAllMock).toMatchObject([
+      { event: 'add', path: file, stats: { size: 7 } },
+      { event: 'change', path: file, stats: { size: 7 } },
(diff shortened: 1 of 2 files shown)


src/apps/main/background-processes/sync-engine/services/spawn-sync-engine-worker.ts
```diff
@@ -89,6 +99,7 @@ export async function spawnSyncEngineWorker({ ctx }: TProps) {
       ctx,
       browserWindow,
       syncSchedule: scheduleSync({ ctx, manager }),
+      watcher: initWatcher({ ctx }),
```
dajimenezriv, line 102: We start the watcher and in the file below we close it if the sync engine "worker" is closed.

dajimenezriv:
@sg-gs Not really. Probabaly in the end it will be renamed and moved to another place but it was more because the `createFile` or `replaceFile` functions already existed. However, the persist functions were more temporal since we had the same funcionality for drive and backus. Anyway, I used persist since those functions persist the data to the wip and sqlite.

PR #1193, opened 2025-12-16
Use connection key for downloads

dajimenezriv-internxt:
## What

Since now all downloads will happen in the main process the addon needed a way to map each conneciton key with the function reference that performs the download or the cancellation. However, do this in the C++ code is more difficult so I decided to create a Map in the javascript code mapping the connectionKey with the context of the workspace. Now, when `connectSyncRoot` we receive a connection key that we will use to retrieve the context of the workspace when a file is going to be hydrated. Also, this connectionKey is used when disconnecting the sync root.

### DIFF
--- a/src/node-win/callbacks.ts
+++ b/src/node-win/callbacks.ts
@@ -33,7 +33,7 @@ export const cancelFetchDataFn: CancelFetchDataFn = (connectionKey, win32Path) =
   const path = abs(win32Path);
 
   if (ctx) {
-    ctx.logger.debug({ msg: 'Cencel fetch data callback', path });
+    ctx.logger.debug({ msg: 'Cancel fetch data callback', path });
     ctx.contentsDownloader.forceStop({ path });
     LocalSync.SyncState.addItem({ action: 'DOWNLOAD_CANCEL', path });
   } else {


src/apps/sync-engine/refresh-workspace-token.ts
```diff
@@ -6,6 +6,8 @@ type Props = {
 };
 
 export function refreshWorkspaceToken({ ctx }: Props) {
+  if (!ctx.workspaceToken) return undefined;
```
dajimenezriv, line 9: This will be executed also for the default drive. In that case it doesn't have a `workspaceToken` so we just don't create the interval.

packages/addon/include/helpers/napi_extract_args.h
```diff
@@ -9,6 +9,12 @@
 template <typename T>
 T napi_extract_value(napi_env env, napi_value value);
 
+template <>
+inline napi_value napi_extract_value<napi_value>(napi_env env, napi_value value)
+{
+    return value;
```
sg-gs, line 15: what

src/node-win/callbacks.ts
```diff
@@ -0,0 +1,47 @@
...
+    });
+  }
+};
+
+export const cancelFetchDataFn: CancelFetchDataFn = (connectionKey, win32Path) => {
+  const ctx = ctxs.get(connectionKey);
+  const path = abs(win32Path);
+
+  if (ctx) {
+    ctx.logger.debug({ msg: 'Cencel fetch data callback', path });
```
sg-gs: Typo here, is _Cancel_

packages/addon/include/helpers/napi_extract_args.h
dajimenezriv: This is a file that works as a helper for extracting the props from javascript call. Even if in this case the thing that we receive it's a `napi_value` we still need to add the tamplate to extract it. We extract it like this:

```
auto [syncRootPath, fetchDataCallback, cancelFetchDataCallback] =
        napi_extract_args<std::wstring, napi_value, napi_value>(env, info);
```

PR #1195, opened 2025-12-17
Remove effect

dajimenezriv-internxt:
## What

We have found that the library `Effect` is increasing the CPU usage so we have decided to remove it. This PR refactors the code from `Effect` to normal javascript.


src/node-win/watcher/events/on-change.ts
```diff
@@ -14,13 +14,13 @@ type Props = {
...
+    const { data: stats } = await fileSystem.stat({ absolutePath: path });
 
-    if (stats.isDirectory()) return;
+    if (!stats || stats.isDirectory()) return;
 
-    const { data: fileInfo, error } = await NodeWin.getFileInfo({ path });
+    const { data: fileInfo } = await NodeWin.getFileInfo({ path });
 
-    if (error) throw error;
+    if (!fileInfo) return;
```
dajimenezriv, line 23: This was throwing many errors because the on change event is also triggered when delete items, we just return in case there are not `stats` or `fileInfo`.

PR #1196, opened 2025-12-17
Trash item if paths do not match on delete

dajimenezriv-internxt:
## What

1. Node.js `stat` function return a property called `blocks` that tells how many blocks a file have in the local disk. However, this property is not reliable since when we create a new file it starts with 0 blocks. This was causing issues in the dehydrate file event since we expect blocks to be different than 0 (otherwise is already dehydrated). The solution is to return the property `onDiskDataSize` from C++ when retrieving the placeholder info.
2. When a file or folder is marked as TRASHED/DELETED in remote but that item still exists in the file explorer in another path, it means that we had an inconsistency. In this case, instead of deleting the item we are gonna move it to the trash so the user can decide what to do with it.

### DIFF
--- a/src/backend/features/remote-sync/file-explorer/delete-item-placeholder.ts
+++ b/src/backend/features/remote-sync/file-explorer/delete-item-placeholder.ts
@@ -19,14 +19,13 @@ export async function deleteItemPlaceholder({ ctx, type, remote, locals }: Props
     if (local.path !== remote.absolutePath) {
       /**
        * v2.6.4 Daniel Jiménez
-       * This is a temporary migration. Previously, we didn't remove items with mismatched paths, but
-       * we are unsure how many users have been affected. For now, we will revert the placeholder
-       * so the items reupload. In a few months, we will remove this and simply delete the items,
-       * regardless of the path, so the remote version's state prevails.
+       * If we reach this point, it means we had an inconsistency between remote and local,
+       * so instead of deleting the placeholder, we are going to send the item to the trash
+       * so the user can decide whether to delete it or retrieve it.
        */
 
       ctx.logger.error({
-        msg: 'Path does not match when removing placeholder',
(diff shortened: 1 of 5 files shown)


src/backend/features/remote-sync/file-explorer/delete-item-placeholder.ts
```diff
@@ -4,29 +4,39 @@ import { rm } from 'node:fs/promises';
...
+      ctx.logger.error({
+        msg: 'Path does not match when removing placeholder',
+        remotePath: remote.absolutePath,
+        localPath: local.path,
+        type,
+      });
 
-  try {
-    ctx.logger.debug({ msg: 'Delete placeholder', path: remote.absolutePath });
+      await trash(local.path);
```
dajimenezriv, line 34: We can't use `shell.trashItem` from electron since it fails with dehydrated files.

packages/addon/native-src/virtual_drive/convert_to_placeholder.cpp
```diff
@@ -11,14 +11,16 @@ void convert_to_placeholder(const std::wstring& path, const std::wstring& placeh
...
+    LPCVOID fileIdentity = static_cast<LPCVOID>(placeholderId.c_str());
+    DWORD fileIdentityLength = static_cast<DWORD>(placeholderId.size() * sizeof(wchar_t));
+
+    HRESULT hr = CfConvertToPlaceholder(
+        fileHandle.get(),
+        fileIdentity,
+        fileIdentityLength,
+        CF_CONVERT_FLAG_MARK_IN_SYNC,
+        nullptr,
+        nullptr);
```
dajimenezriv, line 23: Same funcionality but simplified in `convertToPlaceholder`. We were not using `convertUsn` or `overlapped`.

PR #1204, opened 2025-12-19
Simplify abort on change size

dajimenezriv-internxt:
## What

Refactor how we check if the file size has been changed during the upload. Before we were doing an interval but it can be risky if we forget to clear the interval since it will create a new one for every file upload.

The sonarcloud error cannot be solved since the progress callbacks expects a return void, and we are returning a Promise<void>.

### DIFF
--- a/src/infra/inxt-js/file-uploader/upload-file.ts
+++ b/src/infra/inxt-js/file-uploader/upload-file.ts
@@ -42,8 +42,7 @@ export function uploadFile({ ctx, fn, readable, size, abortSignal, path }: Props
         if (stats && stats.size !== size) {
           logger.debug({ msg: 'Upload file aborted on change size', path, oldSize: size, newSize: stats.size });
           stopUpload(state);
-          resolve();
-          return;
+          return resolve();
         }
 
         LocalSync.SyncState.addItem({ action: 'UPLOADING', path, progress });


src/infra/inxt-js/file-uploader/upload-file.ts
```diff
@@ -25,40 +24,35 @@ export function uploadFile({ ctx, fn, readable, size, abortSignal, path }: Props
...
-          }
-
-          return resolve({ error: processError({ path, err }) });
-        },
-        progressCallback: (progress) => {
-          LocalSync.SyncState.addItem({ action: 'UPLOADING', path, progress });
-        },
-      });
-
-      interval = setInterval(() => abortOnChangeSize({ path, size, resolve, stopUpload, state }), 5000);
```
dajimenezriv, line 50: We don't want this interval here since it will check every 5s if the file size has changed. However it's a bit risky if we forgot to clear the interval. For that reason we have moved the check file size login inside the `progressCallback`. Now every time we receive a progress we check the size and stop in case it has changed.

src/infra/inxt-js/file-uploader/process-error.ts
```diff
@@ -1,45 +1,34 @@
...
+    if (error.message === 'Max space used') {
+      logger.error({
         msg: 'Failed to upload file to the bucket. Not enough space',
         path,
-        error: err,
+        error,
       });
 
-      return new EnvironmentFileUploaderError('NOT_ENOUGH_SPACE', err);
+      addGeneralIssue({ error: 'NOT_ENOUGH_SPACE', name: path });
```
dajimenezriv, line 24: We were only doing something with this error, so instead of sending the error to above functions we just add the general issue here and that's all.

PR #1205, opened 2025-12-19
Simplify backups start

dajimenezriv-internxt:
## What

Refactor and simplify how we start backups and how we schedule them.

### DIFF
--- a/src/apps/main/background-processes/backups/launchBackupProcesses.ts
+++ b/src/apps/main/background-processes/backups/launchBackupProcesses.ts
@@ -1,7 +1,7 @@
 import { ipcMain, powerSaveBlocker } from 'electron';
 import { executeBackupWorker } from './BackukpWorker/executeBackupWorker';
 import { backupsConfig } from './BackupConfiguration/BackupConfiguration';
-import { logger } from '@/apps/shared/logger/logger';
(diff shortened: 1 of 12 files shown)


src/apps/main/background-processes/backups/BackupScheduler/BackupScheduler.ts
```diff
@@ -1,87 +1,34 @@
-import { getIsLoggedIn } from '../../../auth/handlers';
-import configStore from '../../../config';
+import { logger } from '@internxt/drive-desktop-core/build/backend';
+import { launchBackupProcesses } from '../launchBackupProcesses';
+import { electronStore } from '@/apps/main/config';
 
 export class BackupScheduler {
-  private static schedule: null | ReturnType<typeof setTimeout> = null;
+  private static timeout: NodeJS.Timeout | undefined;
```
egalvis27, line 7: Considering that these methods are static, would it be better to refactor them using a functional programming approach? and rename the file at once

dajimenezriv: Maybe, but that's outside the scope of the PR and implies a big change that will break the PR size limit.

PR #1206, opened 2025-12-19
Cleanup config

dajimenezriv-internxt:
## What

Previously we had a global config that was accessed using `getConfig()`. This made more or less sense taking into account that there were multiple processes, but now we just have one. The solution was to put all values into the `context` that most functions have. We remove this config that is no longer used.

We also simplify some things related with backups. Basically the devices already have the decrypted name so we don't need to the decrypt it.


src/apps/main/device/service.ts
```diff
@@ -11,13 +11,11 @@ import { client } from '@/apps/shared/HttpClient/client';
 import { driveServerWipModule } from '@/infra/drive-server-wip/drive-server-wip.module';
 import { addGeneralIssue } from '@/apps/main/background-processes/issues';
 import { getBackupsFromDevice } from './get-backups-from-device';
-import { decryptDeviceName } from '@/backend/features/device/services/decrypt-device-name';
 
 export type Device = {
-  name: string;
+  plainName: string;
```
dajimenezriv, line 16: Use the decrypted name instead of the encrypted name. Now we don't need the `decryptDeviceName` function.

PR #1208, opened 2025-12-22
Use sync functions in backups

dajimenezriv-internxt:
## What

Use the sync functions `createFile`, `createFolder` and `replaceFile` inside backups.

### DIFF
--- a/src/apps/backups/folders/create-folders.ts
+++ b/src/apps/backups/folders/create-folders.ts
@@ -29,14 +29,18 @@ export async function createFolders({ self, ctx, added, tree, tracker }: TProps)
 async function createFolder({ ctx, local, tree }: { ctx: BackupsContext; local: LocalFolder; tree: RemoteTree }) {
   const path = local.absolutePath;
 
-  const parentPath = pathUtils.dirname(local.absolutePath);
-  const parent = tree.folders.get(parentPath);
+  try {
+    const parentPath = pathUtils.dirname(local.absolutePath);
+    const parent = tree.folders.get(parentPath);
 
-  if (!parent) return;
+    if (!parent) return;
 
-  const folder = await Sync.Actions.createFolder({ ctx, path, parentUuid: parent.uuid });
+    const folder = await Sync.Actions.createFolder({ ctx, path, parentUuid: parent.uuid });
 
-  if (!folder) return;
+    if (!folder) return;
 
-  tree.folders.set(path, { ...folder, absolutePath: path });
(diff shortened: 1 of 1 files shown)


src/apps/backups/folders/create-folders.ts
```diff
@@ -4,50 +4,39 @@ import { LocalFolder } from '@/context/local/localFolder/domain/LocalFolder';
...
 
-    await createFolder({ context, local, tree });
+    await createFolder({ ctx, local, tree });
     self.backed++;
     tracker.currentProcessed(self.backed);
   }
 }
 
-async function createFolder({ context, local, tree }: { context: BackupsContext; local: LocalFolder; tree: RemoteTree }) {
-  if (local.absolutePath === context.pathname) return;
```
dajimenezriv, line 32: We don't need to check for the root directly since when trying to obtain the parent it doesn't have and it will be skipped.

src/apps/backups/Backups.ts
```diff
@@ -14,25 +13,24 @@ import { deleteFolders } from './folders/delete-folders';
...
+      rootPath: ctx.pathname,
+      rootUuid: ctx.folderUuid as FolderUuid,
     });
 
     const foldersDiff = calculateFoldersDiff({ local, remote });
     const filesDiff = calculateFilesDiff({ local, remote });
 
-    logger.debug({
-      tag: 'BACKUPS',
+    ctx.logger.debug({
```
AlexisMora, line 33: Why are you passing down the logger on the context rather than just importing it ?

src/apps/backups/Backups.ts
```diff
@@ -14,25 +13,24 @@ import { deleteFolders } from './folders/delete-folders';
...
-    const local = await LocalTreeBuilder.run({ context });
+  async run({ tracker, ctx }: Props) {
+    const local = await LocalTreeBuilder.run({ context: ctx });
     const remote = await Traverser.run({
-      userUuid: context.userUuid,
-      rootPath: context.pathname,
-      rootUuid: context.folderUuid as FolderUuid,
+      userUuid: ctx.userUuid,
+      rootPath: ctx.pathname,
+      rootUuid: ctx.folderUuid as FolderUuid,
```
AlexisMora, line 27: is this type assertion needed?

AlexisMora (approved):

src/apps/backups/Backups.ts
dajimenezriv: Yes, it is needed.

PR #1209, opened 2025-12-22
Create drive actions

dajimenezriv-internxt:
## What

Since now we have backups and drive running in the same process we can start joining the funcionality. We are going to have 3 different modules (drive, backups and sync). Sync will be used for both. In the previous PR we created the shared functions createFile, createFolder and replaceFile. Now, we create the drive functions for `createFile`, `createFolder` and `replaceFile`.

We change how the watcher works too. Previously we were making a `createFile` or `createFolder` for every event emitted of a file or a folder that didn't have a placeholder. However, this was causing a problem in which if one parent folder failed all childreen were still processed. Now, we are going to process only the files and folders that have a parent placeholder. Otherwise we will ignore them and after the parent folder is created we will do a `readdir` of the parent and process all children that do not have a placeholder.


src/infra/drive-server-wip/out/ipc-main.ts
```diff
@@ -17,6 +17,7 @@ import { HttpRemoteFileSystem } from '@/context/virtual-drive/files/infrastructu
 import { createOrUpdateFile } from '@/backend/features/remote-sync/update-in-sqlite/create-or-update-file';
```
dajimenezriv, line 17: This is just a temporal fix. These functions are still used in backups but they will be removed. Mantaining the { data, error } in `createOrUpdateFile/Folder` functions was causing many if else and providing no utility since the error was not not used anywhere. If the file or folder cannot be persisted in sqlite (returns undefined and logs), we just ignore and that's all.

PR #1210, opened 2025-12-22
Create pending items

dajimenezriv-internxt:
## What

Context: we had a problem in drive because when we were adding a folder with files inside, all events are processed at the same time, but if the parent folder failed all children files were tried to be uploaded. Now, we have change that so that in the watcher we only process the parent folder. So, when we process now the children folders? When the parent folder is uploaded we perform a readdir of the parent folder and start processing all children files and folders that are not a placeholder yet. We do that with this new function `createPendingItems`. This new function is going to replace the previous one `addPendingItems` when checking the file explorer state in the app startup to see if there are any folders or files pending of being uploaded.


src/backend/features/drive/actions/services/create-folder.ts
```diff
@@ -17,6 +18,13 @@ export async function createFolder({ ctx, path, parentUuid }: Props) {
     if (!folder) return;
 
     await Addon.convertToPlaceholder({ path, placeholderId: `FOLDER:${folder.uuid}` });
+
+    await createPendingItems({
```
dajimenezriv, line 22: When the parent folder is uploaded we perform a readdir of the parent folder and start processing all children files and folders that are not a placeholder yet.

src/apps/sync-engine/in/add-pending-items.ts
```diff
@@ -13,27 +9,17 @@ export async function addPendingItems({ ctx }: Props) {
   try {
```
egalvis27: Is there a need to keep this function? The only unique functionality it provides is logging execution time, and it is not used intensively. It appears only in spawn-sync-engine-worker.ts and its corresponding test file. It would be more efficient to replace it directly with the new function that has been added.

dajimenezriv: Why is logging execution time not a valid funcionality? Yes, it appears only once. And yes, it's going to be more "efficient" (it's just 2 lines of code) if we replace it with the new function, but then how can I measure execution time or how I know when the function finishes?

egalvis27: Why is this function not used in places where createPendingItems is called directly? Why is timing measurement only relevant when spawning the sync worker? Wouldn't this information be useful in other locations as well, thereby justifying the function's existence?

For example:

- create-folder.ts
- create-pending-folders.ts

dajimenezriv: Because it's only when called from `add-pending-items.ts` (that has the prop `isFirstExecution` to true) that we want to measure, since it iterates the whole file explorer uploading pending items. It helps us differenciate in the logs if the files that are being uploaded are from the pending items or from a regular upload. Maybe we can replace it in `create-folder.ts` but I think it's not necessary for now. We cannot replace it in `create-pending-folders` since it's a recursive function and it will be logged everytime.

PR #1211, opened 2025-12-23
Simplify tree builder in backups

dajimenezriv-internxt:
## What

This is the first of 2 PRs to implement the new sync functions (`createFile`, `createFolder` and `replaceFile`) in backups. First we need the whole `stats` property of the local files, so we have refactored and simplified how we create the local tree. The ouput is the same, but instead of the files inside the tree being of type `LocalFileDTO` now they are `SyncWalkItem`, which is the default type for an item from the file explorer.

### DIFF
--- a/src/apps/backups/process-files/create-files.test.ts
+++ b/src/apps/backups/process-files/create-files.test.ts
@@ -24,7 +24,7 @@ describe('create-files', () => {
       context: { abortController: new AbortController() },
       tracker: { currentProcessed: vi.fn() },
       remoteTree: { folders: new Map([[parentPath, { uuid: 'parentUuid' as FolderUuid }]]) },
-      added: [{ absolutePath: path, size: 1024 }],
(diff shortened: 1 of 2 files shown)


src/context/local/localTree/application/LocalTreeBuilder.ts
```diff
@@ -1,61 +1,52 @@
...
+  static async traverse({ context, tree }: { context: BackupsContext; tree: LocalTree }) {
+    const items = await syncWalk({
+      rootFolder: context.pathname,
+      onError: ({ path, error }) => {
+        parseStatError({ context, path, error });
+        logger.error({ tag: 'BACKUPS', msg: 'Error getting item stats', path, error });
+      },
+    });
 
-      if (file.size > SyncModule.MAX_FILE_SIZE) {
```
dajimenezriv, line 22: We don't want here to filter by size since we are already doing it inside the sync functions.

PR #1212, opened 2025-12-23
Fix C++ sonar analysis

dajimenezriv-internxt:
## What

For now we are going to disable the sonar analysis in the addon and the core packages, since they are pointing to their own repos when looking for PRs. In the code side everything is configured but commented.

Also, in windows we have pointed to the local core package instead of the published one, but we will keep the version for linux compatibility.


egalvis27 (approved):

src/infra/inxt-js/file-uploader/upload-file.ts
```diff
@@ -35,18 +36,19 @@ export function uploadFile({ ctx, fn, readable, size, path }: Props) {
...
-      progressCallback: async (progress) => {
-        const { data: stats } = await fileSystem.stat({ absolutePath: path });
+    });
 
-        if (stats && stats.size !== size) {
-          logger.debug({ msg: 'Upload file aborted on change size', path, oldSize: size, newSize: stats.size });
-          stopUpload(state);
-          return resolve();
-        }
+    async function progressCallback(progress: number) {
```
dajimenezriv, line 41: This is just because a sonar error, since `progressCalback` accepts a sync function and we were sending an async function.

dajimenezriv:
@sg-gs This new PR activates the sonar pipelines https://github.com/internxt/drive-desktop/pull/1257

PR #1215, opened 2025-12-24
Add abort controller to delete calls

dajimenezriv-internxt:
## What

1. Add abort controller to most requests.
2. Add workspace token to most requests.
3. Move `deleteBackupsFromDevice` ipc event to be defined when login and remove the listener when logout. This is because to delete a backup we needed the `AuthContext` that is created when login.

### DIFF
--- a/src/apps/main/config.ts
+++ b/src/apps/main/config.ts
@@ -1,6 +1,7 @@
 import Store, { Schema } from 'electron-store';
 import { AppStore } from '@/core/electron/store/app-store.interface';
 import { defaults } from '@/core/electron/store/defaults';
+import { PATHS } from '@/core/electron/paths';
 
 const schema: Schema<AppStore> = {
   backupInterval: { type: 'number' },
@@ -24,7 +25,14 @@ const schema: Schema<AppStore> = {
   'migrations.v2-6-3-move-checkpoint-to-sqlite': { type: 'boolean' },
 } as const;
 
-const configStore = new Store({ schema, defaults, accessPropertiesByDotNotation: false });
+const configStore = new Store({
+  schema,
+  defaults,
+  accessPropertiesByDotNotation: false,
+  cwd: PATHS.INTERNXT,
+  name: 'config',
+  fileExtension: 'json',
+});
 
 function get<T extends keyof AppStore>(key: T) {
   return configStore.get(key);
--- a/src/apps/shared/logger/services/open-logs.test.ts
+++ b/src/apps/shared/logger/services/open-logs.test.ts
@@ -7,6 +7,8 @@ import { INTERNXT_LOGS } from '@/core/utils/utils';
 import archiver from 'archiver';
 import { mockDeep } from 'vitest-mock-extended';
 import { pipeline } from 'node:stream/promises';
+import { join } from '@/context/local/localFile/infrastructure/AbsolutePath';
+import { PATHS } from '@/core/electron/paths';
 
 vi.mock(import('node:fs'));
 vi.mock(import('node:stream/promises'));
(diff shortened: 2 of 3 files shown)


src/infra/drive-server-wip/out/ipc.ts
```diff
@@ -1,22 +1,17 @@
...
   parentUuid: FolderUuid;
   path: AbsolutePath;
 };
 export type PersistMoveFolderProps = {
   ctx: CommonContext;
   uuid: FolderUuid;
-  workspaceToken: string;
   parentUuid: FolderUuid;
   path: AbsolutePath;
 };
```
egalvis27, line 17: These models are exactly the same; it would be good to be able to unify them.

src/infra/drive-server-wip/out/ipc-main.ts
```diff
@@ -1,14 +1,18 @@
...
 import { createOrUpdateFolder } from '@/backend/features/remote-sync/update-in-sqlite/create-or-update-folder';
+import { CommonContext } from '@/apps/sync-engine/config';
+import { FileUuid } from '@/apps/main/database/entities/DriveFile';
+import { AbsolutePath } from '@internxt/drive-desktop-core/build/backend';
+import { FolderUuid } from '@/apps/main/database/entities/DriveFolder';
 
-export async function deleteFileByUuid({ uuid, path, workspaceToken }: DeleteFileByUuidProps) {
-  const res = await driveServerWip.storage.deleteFileByUuid({ path, uuid, workspaceToken });
+export async function deleteFileByUuid({ ctx, path, uuid }: { ctx: CommonContext; path: AbsolutePath; uuid: FileUuid }) {
+  const res = await driveServerWip.storage.deleteFileByUuid({ ctx, context: { path, uuid } });
```
egalvis27, line 16: The parameter models for these functions are identical, so it would be beneficial to use a single shared model for both. I don't anticipate them diverging in the future, and even if modifications are needed, they will likely continue to share the same parameters due to the similar nature of these functions

src/infra/drive-server-wip/services/folders/create-folder.ts
```diff
@@ -18,19 +18,15 @@ class CreateFolderError extends DriveServerWipError {
   }
 }
 
-type Props = {
-  ctx: CommonContext;
-  context: { path: AbsolutePath; body: CreateFolderBody };
-};
-
-export async function createFolder({ ctx, context }: Props) {
+export async function createFolder({ ctx, context }: { ctx: CommonContext; context: { path: AbsolutePath; body: CreateFolderBody } }) {
```
egalvis27: This change is unnecessary. I don't see why we should remove the type to put it in the function and make that line longer.

egalvis27 (changes requested):

src/infra/drive-server-wip/out/ipc.ts
dajimenezriv: They are not the same, the uuid can be `FiileUuid` or `FolderUuid`, and we cannot do `FileUuid | FolderUuid`, sicne we lose the strict type for inner functions.

src/infra/drive-server-wip/out/ipc-main.ts
dajimenezriv: As said, one is `FileUuid` and the other is `FolderUuid`. It seems the same but it's not. We have been narrowing the types to be very strict forbiding us to call for example the move file function with a folder uuid because of this.

src/infra/drive-server-wip/services/folders/create-folder.ts
dajimenezriv: I agree, I don't why I did that change. Reverted.

PR #1217, opened 2025-12-24
Add abort controller to get and create requests

dajimenezriv-internxt:
## What

Now that we have all sync requests in the main process we have to improve the cleanup. Basically when we logout all requests must be aborted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "init:ci": "npm ci --ignore-scripts && node node_modules/electron/install.js",
     "init:dev": "npm install && node node_modules/electron/install.js && npm run build:dll && npm run reload-native-deps",
     "========== Code style ==========": "",
-    "lint": "eslint src --ext .ts,.tsx --max-warnings 360",
+    "lint": "eslint src --ext .ts,.tsx --max-warnings 332",
     "lint:fix": "npm run lint -- --fix",
     "format": "prettier src tests --check",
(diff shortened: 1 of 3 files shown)


src/infra/drive-server-wip/services/files.service.ts
```diff
@@ -20,21 +21,21 @@ export const FileModule = files;
 export type GetFilesQuery = paths['/files']['get']['parameters']['query'];
 type TCreateThumnailBody = paths['/files/thumbnail']['post']['requestBody']['content']['application/json'];
 
-async function getFiles(context: { query: GetFilesQuery }, extra?: { abortSignal: AbortSignal; skipLog?: boolean }) {
+async function getFiles({ ctx, context, skipLog }: { ctx: CommonContext; context: { query: GetFilesQuery }; skipLog?: boolean }) {
```
egalvis27, line 24: I think having two contexts in one function can be confusing. I understand that one is the request context and the other is the app context, but the request context could be renamed to improve readability and avoid confusion.

dajimenezriv: I agree with you and I don't like having two contexts either. I didn't want to change it now because it was the name already used in all requests and it will imply so many changes, so I would like to leave it for another PR. But yes, you are right.

egalvis27 (approved):

PR #1221, opened 2025-12-29
Add parcel watcher

dajimenezriv-internxt:
## What

Explained in ticket https://inxt.atlassian.net/browse/PB-5622
First PR of 2PRs. Next one here: https://github.com/internxt/drive-desktop/pull/1220

We did a change in the watcher in the previous release to stop using the `usePolling` feature from chokidar (which performed a stat to every folder every 100ms) to start using the native events of windows. However this added another problem. Some files seems to be locked while the watcher is running. We had that issue with a folder that Valentyna was trying to remove in the QA build, but we solved it and seemed a corner case. However, this error has happened to me too and it’s not a corner case. We need to check this in more detail because maybe we need to use the usePolling feature again (this time doing the statevery 1s) or look for other alternatives.

Solution: replace `chokidar` with `@parcel/watcher`.


src/backend/features/local-sync/watcher/events/move-file.test.ts
```diff
@@ -31,7 +31,7 @@ describe('move-file', () => {
     getFolderInfoMock.mockResolvedValue({ data: { uuid: 'parentUuid' as FolderUuid } });
   });
 
-  it('should perform move request and ignore unlink event', async () => {
+  it.skip('should perform move request and ignore unlink event', async () => {
```
dajimenezriv, line 34: Temporal change, this is removed in the next PR.

src/backend/features/local-sync/watcher/events/move-folder.test.ts
```diff
@@ -34,7 +34,7 @@ describe('move-folder', () => {
     });
   });
 
-  it('should perform move request and ignore unlink event', async () => {
+  it.skip('should perform move request and ignore unlink event', async () => {
```
dajimenezriv, line 37: Temporal change, this is removed in the next PR.

src/node-win/watcher/process-event.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    if (event.type === 'delete') {
+      await onUnlink({ ctx, path });
+      return;
+    }
+
+    const stats = await stat(path);
+
+    if (event.type === 'update') {
+      // if (stats.isFile()) debounceOnRaw({ ctx, path, stats });
```
dajimenezriv, line 27: Temporal change, this is uncommented in the next PR.

src/node-win/watcher/watcher.ts
```diff
@@ -39,8 +37,8 @@ export function initWatcher({ ctx }: Props) {
      * - we delete an item locally or when we delete it using sync by checkpoint.
      * - we move an item locally or when we move it using sync by checkpoint.
      */
-    .on('unlink', (path) => unlinkFile({ ctx, path: abs(path) }))
-    .on('unlinkDir', (path) => unlinkFolder({ ctx, path: abs(path) }))
+    // .on('unlink', (path) => unlinkFile({ ctx, path: abs(path) }))
+    // .on('unlinkDir', (path) => unlinkFolder({ ctx, path: abs(path) }))
```
dajimenezriv, line 41: Temporal change, this is uncommented in the next PR.

PR #1223, opened 2025-12-30
Implement custom watcher C++

dajimenezriv-internxt:
## What

Ticket https://inxt.atlassian.net/browse/PB-5622.

We are going to move all code to header files so we don't need to duplicate files to create the function definition and another for the function implementation. This is something allowed in C++ and doesn't have any important implication when building and linking the files.

Implement a C++ watcher:
1. Inside `watchPathWrapper`: get the props from the javascript call (`rootPath` and the `callback` function to call when an event is listened). Create the context so later it can be stopped. Launch the `watchPath` function inside a detached thread so it doesn't block the Node.js main thread. In case of exception call the event function with the `type=error`. After stop or exception, delete the context. Return the handle to javascript so we can send it later to `unwatch` it.
2. Open the `rootPath` handle and stop execution in `ReadDirectoryChangesW` until it receives an event or multiple events. Process events inside `processEvent`. Then stop again waiting in `ReadDirectoryChangesW`.


packages/addon/include/helpers/register_threadsafe_callback.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include <external.h>
+
+napi_threadsafe_function registerThreadsafeCallback(const std::string& resourceName, napi_env env, napi_value callback, napi_threadsafe_function_call_js callJsCallback)
```
dajimenezriv, line 5: This is a helper to register callbacks from javascript. It will be used later for the `hydrate file` and `cancel hydrate file`.

packages/addon/include/virtual_drive/watcher/watch_path.h
```diff
@@ -0,0 +1,108 @@
...
+    }
+}
+
+inline void watchPath(WatcherContext* ctx, const std::wstring& rootPath)
+{
+    auto hDirectory = Placeholders::OpenFileHandle(rootPath.c_str(), FILE_LIST_DIRECTORY, false);
+
+    BYTE buffer[4 * 1024];
+
+    while (!ctx->shouldStop) {
```
sg-gs, line 54: Is the performance of this ok or should you add a timer?

packages/addon/include/virtual_drive/watcher/watch_path.h
```diff
@@ -0,0 +1,108 @@
...
+                hDirectory.get(),
+                buffer,
+                sizeof(buffer),
+                TRUE,
+                FILE_NOTIFY_CHANGE_FILE_NAME | FILE_NOTIFY_CHANGE_DIR_NAME | FILE_NOTIFY_CHANGE_SIZE | FILE_NOTIFY_CHANGE_ATTRIBUTES,
+                &bytesReturned,
+                nullptr,
+                nullptr);
+
+            if (!success || ctx->shouldStop) break;
```
sg-gs: The watcher stops when is not able to watch the directory once or an error here means it is not able to watch it anymore?

packages/addon/include/virtual_drive/watcher/watch_path.h
dajimenezriv: It's ok, the while is not looping all time. When it executes `ReadDirectoryChangesW` it sleeps there until an event is received. We want to send the events to javascript asap. Maybe in the future we need to throttle, however, based on my tests, I couldn't reach a point in which many events were sent so fast, because when copying a folder inside the cloud provider it already takes time to copy the items one by one (so it's already throttled), and when we move we only receive an event from the parent folder.

packages/addon/include/virtual_drive/watcher/watch_path.h
dajimenezriv: Only when it's not able to watch the directory. In case of error it still iterates. I haven't tested when that is not `success` but I assume that if the directory cannot be watched (access denied for example) I don't want to keep looping and send the same error always. However, good catch, because in case of not success I was never logging anything, changing it.

packages/addon/include/virtual_drive/watcher/watch_path.h
```diff
@@ -0,0 +1,119 @@
...
+
+            wprintf(L"Remove watcher context\n");
+
+            napi_release_threadsafe_function(ctx->tsfn, napi_tsfn_release);
+            delete ctx;
+        } catch (...) {
+            auto error = format_exception_message("WatchPathThread");
+            wprintf(L"Error in watch path thread: %s\n", error.c_str());
+        }
+    }).detach();
```
dajimenezriv, line 114: This is the most dangerous part of the code because we are creating a thread and detaching it. If a C++ exception is not catched inside it will break the whole app. I added a try catch to everything inside so that cannot happen. This is the same that we did a month ago with all other addon functions.

PR #1224, opened 2026-01-01
Implement custom watcher JS

dajimenezriv-internxt:
## What

As explained in this ticket https://inxt.atlassian.net/browse/PB-5622 we had some problems with `chokidar` (high CPU usage or locked files) and we tried `@parcel/watcher`, but since it was throttling the events it wasn't the perfect match for us, so in the end we decided to build our own watcher in C++. That was done in the previous PR, this one implementes that watcher into the existing code. The idea is to match as much as possible the simple functionality of `@parcel/watcher`.

We are also going to join the map that tracks file and folder moves into just one for simplicity.

### DIFF
--- a/src/node-win/watcher/process-event.ts
+++ b/src/node-win/watcher/process-event.ts
@@ -23,6 +23,7 @@ export async function processEvent({ ctx, event, path }: Props) {
 
     if (event === 'update' && stats.isFile()) {
       debounceOnRaw({ ctx, path, stats });
+      return;
     }
 
     if (event === 'create') {


src/infra/drive-server-wip/out/error.types.ts
```diff
@@ -1,4 +1,4 @@
-export type TDriveServerWipError = 'UNKNOWN' | 'NETWORK' | 'SERVER' | 'ABORTED' | (string & {});
+export type TDriveServerWipError = 'UNKNOWN' | 'NETWORK' | 'SERVER' | 'ABORTED' | (string & { readonly brand?: unique symbol });
```
dajimenezriv, line 1: This was an error of sonarcloud.

src/backend/features/local-sync/watcher/events/unlink/get-parent-uuid.ts
```diff
@@ -21,9 +17,14 @@ export async function getParentUuid({ ctx, path }: Props) {
    * - if the parent doesn't exist it means that this item has been deleted because it's inside
    * of a folder that has been deleted and we need to find that folder to mark it as TRASHED.
    *
-   * Warning: Using just folderUuid is not going to work. There are some times in which we are
-   * getting the parentUuid even when the parent folder is deleted.
+   * Warning: events from C++ are too fast, so we need to wait before checking the parent folder.
+   * Otherwise we are going to get the parent info like if it still exists.
    */
-  if (parentInfo && stats) return parentInfo.uuid;
+  await sleep(2_000);
```
sg-gs, line 23: This seems a bit unreliable. An approach could be to add timestamps to the events so you can know in which order happens each thing

src/node-win/watcher/process-event.ts
```diff
@@ -1,39 +1,38 @@
...
       return;
     }
 
     const stats = await stat(path);
 
-    if (event.type === 'update') {
-      if (stats.isFile()) debounceOnRaw({ ctx, path, stats });
-      return;
+    if (event === 'update' && stats.isFile()) {
+      debounceOnRaw({ ctx, path, stats });
```
sg-gs, line 25: Should you return here?

src/backend/features/local-sync/watcher/events/unlink/get-parent-uuid.ts
dajimenezriv: The problem is that I cannot relate events between them, I don't even know if the first event to be retrieved is the one from the root folder. From the javascript perspective I just receive an `unlink` event. So what I do is, does the parent exists in the file explorer? If it exists, it means that the `unlink` event that I received is the root, otherwise I ignore it.

src/node-win/watcher/process-event.ts
dajimenezriv: Done

PR #1227, opened 2026-01-05
Load all file explorer at once

dajimenezriv-internxt:
## What

1. When refactoring I added a bug to the drive traverser. Previously we were obtaining the whole file explorer state into memory before doing the traverser, but I moved to only get the file explorer state of the current folder that we are traversing. This leads to a bug, because when checking the move operation we need to have the whole file explorer state to know if a file/folder exists in another folder or it was just removed.
2. In backups convert the dict of folders to just a list of the absolute paths, since it's the only things used. And make the walk to obtain all backup local state more performant.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,7 @@
         "react": "^17.0.2",
         "react-circular-progressbar": "^2.1.0",
         "react-dom": "^17.0.2",
-        "react-router-dom": "^6.2.1",
+        "react-router-dom": "^6.30.3",
         "reflect-metadata": "^0.1.13",
         "regenerator-runtime": "^0.13.9",
         "socket.io-client": "^4.4.1",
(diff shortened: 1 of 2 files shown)


src/backend/features/remote-sync/sync-items-by-checkpoint/load-in-memory-paths.ts
```diff
@@ -6,35 +6,39 @@ import { Stats } from 'node:fs';
...
 
 type Props = {
   ctx: SyncContext;
-  parentPath: AbsolutePath;
 };
 
-export async function loadInMemoryPaths({ ctx, parentPath }: Props) {
-  const files: InMemoryFiles = new Map();
-  const folders: InMemoryFolders = new Map();
+export async function loadInMemoryPaths({ ctx }: Props) {
```
dajimenezriv, line 16: Instead of loading just the folder items, now we walk through all the inner children if the folder is a placeholder.

src/context/local/localTree/application/LocalTreeBuilder.ts
```diff
@@ -1,52 +1,49 @@
...
     const tree: LocalTree = {
       files: {},
-      folders: {
-        [rootPath]: { absolutePath: rootPath },
-      },
+      folders: [rootPath],
     };
 
-    await this.traverse({ context, tree });
+    async function walk(parentPath: AbsolutePath) {
```
dajimenezriv, line 28: Same here, walk through all children to get the whole local backup state. Previously we were using `syncWalk` but it's not efficient because we were waiting to process all children of a folder before processing any child folder.

src/context/virtual-drive/items/application/Traverser.ts
```diff
@@ -4,57 +4,52 @@ import { ExtendedDriveFolder, SimpleDriveFolder } from '@/apps/main/database/ent
...
-  items: Items;
+  database: Database;
+  fileExplorer: FileExplorer;
   currentFolder: Pick<ExtendedDriveFolder, 'absolutePath' | 'uuid'>;
   isFirstExecution: boolean;
 };
 
-export class Traverser {
-  static async run({ ctx, items, currentFolder, isFirstExecution }: Props) {
-    const { files, folders } = await loadInMemoryPaths({ ctx, parentPath: currentFolder.absolutePath });
```
dajimenezriv, line 25: I've just removed this line, but everything appears as changed because I removed the `Traverser` class and converted to a simple function.

src/context/local/localTree/application/LocalTreeBuilder.ts
egalvis27: Is it necessary to have the walk function inside the run function?

src/backend/features/remote-sync/sync-items-by-checkpoint/load-in-memory-paths.ts
egalvis27: Is it necessary to have the walk function inside the loadInMemoryPaths function?

src/context/local/localTree/application/LocalTreeBuilder.ts
dajimenezriv: It makes easier to use the context and the tree since those are props from the parent function. Also, the `walk` function it's not "so complicated" and in terms of performance is the function definition is one done once. My first option was to write the `walk` function outside, but it feel that the code was not easier to read. Anyway, I don't have a strong opinion about writing a function inside a function, so I don't know. Do you have one or any thing about why we shouldn't use them,?

src/backend/features/remote-sync/sync-items-by-checkpoint/load-in-memory-paths.ts
dajimenezriv: Answered in the other comment.

src/context/local/localTree/application/LocalTreeBuilder.ts
egalvis27: I think it's fine to keep it that way, since the function only makes sense in the context of the other function.

egalvis27 (approved):

PR #1233, opened 2026-01-09
Use last id for recovery sync request

dajimenezriv-internxt:
## What

Use `lastId` to paginate files when sorting by uuid.


src/apps/renderer/hooks/antivirus/useAntivirus.tsx
```diff
@@ -16,8 +16,8 @@ export type UseAntivirusReturn = {
   showErrorState: boolean;
```
dajimenezriv, line 16: Fixing a sonarcloud error and some warnings.

sg-gs:
Did you benchmark any noticeable improvement on speeds / stability by using this approach @dajimenezriv-internxt ?

dajimenezriv:
@sg-gs Not really, here I have a good internet connection and everything feels pretty fast. Also in the `activate@internxt.com` account there are only 4k files, so that means just 4 calls. Also, in this case it's the recovery sync that runs every 30s and the performance here is not so relevant. The one more problematic is the request that filters by `updatedAt`.

PR #1236, opened 2026-01-13
Join widget and settings

dajimenezriv-internxt:
## What

We are going to join all frontend processes into just one so we make the app faster, more efficient and simpler. We want to simulate the windows behaviour by making these new windows draggable using `react-draggable`.

We are going to have just one renderer process that is going to fill the whole screen and create components inside that screen.


src/apps/renderer/pages/Widget/index.tsx
```diff
@@ -5,28 +5,50 @@ import useSyncStatus from '../../hooks/useSyncStatus';
...
+  function onMouseDown(e: React.MouseEvent<HTMLDivElement>) {
+    if (e.target === e.currentTarget) {
+      void globalThis.window.electron.hideFrontend();
+    }
+  }
+
   return (
-    <div className="flex h-screen flex-col overflow-hidden">
-      <Header setIsLogoutModalOpen={setIsLogoutModalOpen} />
+    <div className="relative h-screen w-screen bg-transparent" onMouseDown={onMouseDown}>
```
dajimenezriv, line 26: Now this div fills the whole screen and we have the settings and the widget inside it.

src/apps/renderer/pages/Widget/index.tsx
```diff
@@ -5,28 +5,50 @@ import useSyncStatus from '../../hooks/useSyncStatus';
...
+import Draggable from 'react-draggable';
+import { SETTINGS, useGetWorkArea } from './use-get-work-area';
 
 export function Widget() {
+  const { activeSection: settingsSection } = useSettingsStore();
   const { syncStatus } = useSyncStatus();
+  const { settings } = useGetWorkArea();
   const [isLogoutModalOpen, setIsLogoutModalOpen] = useState<boolean>(false);
 
+  function onMouseDown(e: React.MouseEvent<HTMLDivElement>) {
```
dajimenezriv, line 19: When we click outside of any component we hide the frontend.

PR #1242, opened 2026-01-16
Improve auth guard

dajimenezriv-internxt:
## What

### DIFF
--- a/src/apps/renderer/components/AuthGuard/index.tsx
+++ b/src/apps/renderer/components/AuthGuard/index.tsx
@@ -38,7 +38,7 @@ export function AuthGuard() {
   }
 
   return (
-    <div className="relative h-screen w-screen bg-transparent" role="button" onMouseDown={onMouseDown}>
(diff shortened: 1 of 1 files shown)


src/apps/renderer/pages/Widget/index.tsx
```diff
@@ -5,61 +5,28 @@ import useSyncStatus from '../../hooks/useSyncStatus';
...
-        </Draggable>
-      )}
-
-      {issuesSection && (
-        <Draggable handle=".draggable-handle" defaultPosition={issues?.positions} bounds={issues?.bounds} defaultClassName="absolute">
-          <div className="rounded-shadow-white" style={{ width: ISSUES.width }}>
-            <IssuesPage activeSection={issuesSection} />
-          </div>
-        </Draggable>
-      )}
```
dajimenezriv, line 44: We have already moved this draggable and the one from above to the `logged-page.tsx` component, so this file will be used only for the `Widget.tsx` (as it was before).

PR #1243, opened 2026-01-16
Remove auth browser window

dajimenezriv-internxt:
## What

Last PR removing the multiple processes in the frontend. Now we just have a process which is always visible. Based on the property `isUserLoggedIn` we display the login modal or the widget/settings/issues modals. While `isLoggedIn` is `null` we display the `WidgetSkeleton`. Now, there are not strange popups while first authentication, or logout or similar, since the flow of showing one modal or the other is very simple.

### DIFF
--- a/src/apps/renderer/pages/Widget/Logout.tsx
+++ b/src/apps/renderer/pages/Widget/Logout.tsx
@@ -5,7 +5,7 @@ type Props = {
   onLogout: () => void;
 };
 
-export function ModalLogout({ onClose, onLogout }: Props) {
+export function ModalLogout({ onClose, onLogout }: Readonly<Props>) {
   const { translate } = useI18n();
 
   return (


src/apps/main/tray/tray.ts
```diff
@@ -1,36 +1,27 @@
 import { app, Menu, nativeImage, Tray } from 'electron';
```
dajimenezriv, line 1: Same funcionality, just simplified.

src/apps/main/tray/tray.ts
```diff
@@ -44,58 +35,20 @@ export class TrayMenu {
...
-
-    return Menu.buildFromTemplate(contextMenuTemplate);
-  }
-
   updateContextMenu() {
-    const ctxMenu = this.generateContextMenu();
-    this.tray.setContextMenu(ctxMenu);
+    this.tray.setContextMenu(
+      Menu.buildFromTemplate([
+        { label: 'Show/Hide', click: toggleWidgetVisibility },
```
sg-gs, line 41: Should this be translated?

src/apps/main/tray/tray.ts
```diff
@@ -44,58 +35,20 @@ export class TrayMenu {
...
+        { label: 'Quit', click: () => void quitApp() },
```
sg-gs, line 42: Same here

src/apps/main/tray/tray.ts
dajimenezriv: It should, but it's beyond the scope of the PR. We don't yet have a way to translate things on the backend, and we haven't even considered how to do it.

PR #1354, opened 2026-04-21
[PB-6324] Fix stuck partial hydrated files

dajimenezriv-internxt:
## What

This situation can happen when the user performed an hydration of a folder with files inside and the files didn't complete the hydration before the app closes. The files stay in a stuck state forever. One option could be to resume the hydration, however, since when you hydrate a folder there is no way to cancel the hydration, maybe the user has forced the close of the app to finish it. It can also be the opposite, that the user wants to continue after opening the app again. We will keep the first use case since it's easier to implement, consumes less resources and in case the user wants to hydrate the folder he can perform the action again.

### DIFF
--- a/src/backend/features/remote-sync/file-explorer/update-file-placeholder.ts
+++ b/src/backend/features/remote-sync/file-explorer/update-file-placeholder.ts
@@ -14,48 +14,46 @@ type Props = {
   isFirstExecution: boolean;
 };
 
-export class FilePlaceholderUpdater {
-  static async update({ ctx, remote, files, isFirstExecution }: Props) {
-    const path = remote.absolutePath;
+export async function updateFilePlaceholder({ ctx, remote, files, isFirstExecution }: Props) {
+  const path = remote.absolutePath;
+  const { size } = remote;
 
-    try {
-      const { isValid } = validateWindowsName({ path, name: remote.name });
-      if (!isValid) return;
+  try {
+    const { isValid } = validateWindowsName({ path, name: remote.name });
+    if (!isValid) return;
 
-      const local = files.get(remote.uuid);
+    const local = files.get(remote.uuid);
 
-      if (!local) {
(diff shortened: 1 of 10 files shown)


src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
```diff
@@ -5,32 +5,36 @@ import { Addon } from '@/node-win/addon-wrapper';
 import { FileExplorerFolders } from '../sync-items-by-checkpoint/load-in-memory-paths';
 import { checkIfMoved } from './check-if-moved';
 
-export class FolderPlaceholderUpdater {
```
dajimenezriv-internxt, line 8: I've removed the class but everything else stays the same.

src/backend/features/remote-sync/file-explorer/update-file-placeholder.ts
```diff
@@ -13,36 +14,46 @@ type Props = {
   isFirstExecution: boolean;
 };
 
-export class FilePlaceholderUpdater {
```
dajimenezriv-internxt, line 16: I've removed the class.

src/backend/features/remote-sync/file-explorer/update-file-placeholder.ts
```diff
@@ -13,36 +14,46 @@ type Props = {
...
+        lastWriteTime: new Date(remote.updatedAt).getTime(),
       });
+
+      return;
+    }
+
+    await checkIfMoved({ ctx, type: 'file', remote, localPath: local.path });
+    await checkIfModified({ ctx, local, remote, isFirstExecution });
+
+    if (isFirstExecution) {
```
dajimenezriv-internxt, line 42: This is the only thing that has changed.

AlexisMora (approved):

PR #1361, opened 2026-04-23
Simplify electron autoupdater

dajimenezriv-internxt:
## What

Previously we were using the package `electron-updater` with uses the `autoUpdate` feature of electron. However, during an issue that sometimes we were having with the `latest.yml` and the hash I started investigating how the autoupdate was working by letting it work in my laptop. However, it was producing strange behaviours like this image when updating:

image

I was reading about how it worked, but the customization was very limited and it was working as a black box. I started lookign for alternatives and found a repo that was doing something similar without the hash verification. So, since the implementation it's "simple" and we just need to support windows, I decided to implement it manually so we can customize it as we want and decide when to install, when to ask the user, when to force the update and so.

I've added in this comment 2 videos showing how the update will be made and the dialog that we show the user. https://inxt.atlassian.net/browse/PB-6340?focusedCommentId=71094

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -28,6 +28,7 @@ module.exports = {
         'sonarjs/no-alphabetical-sort': 'off',
         'sonarjs/no-empty-test-file': 'off',
         'sonarjs/os-command': 'off',
+        'sonarjs/publicly-writable-directories': 'off',
       },
     },
   ],
--- a/.github/workflows/check-pr-size.yml
+++ b/.github/workflows/check-pr-size.yml
@@ -7,8 +7,9 @@ on:
 jobs:
   check_pr_size:
     # Ignore PRs with name merge-X-X-X-release
-    if: |
-      !(startsWith(github.head_ref, 'merge-') && endsWith(github.head_ref, '-release') && github.base_ref == 'main')
+    # if: |
+    #   !(startsWith(github.head_ref, 'merge-') && endsWith(github.head_ref, '-release') && github.base_ref == 'main')
+    if: false
     runs-on: ubuntu-latest
     timeout-minutes: 1
 
--- a/packages/core/src/backend/features/cleaner/utils/was-accessed-within-last-hour.test.ts
+++ b/packages/core/src/backend/features/cleaner/utils/was-accessed-within-last-hour.test.ts
@@ -17,10 +17,6 @@ describe('wasAccessedWithinLastHour', () => {
     });
   });
 
-  afterEach(() => {
-    vi.useRealTimers();
-  });
-
   it('should return true when file was modified within last hour', () => {
     // Given
     props.fileStats.mtimeMs = new Date('2025-09-19T11:00:00Z').getTime();
--- a/packages/core/src/backend/features/cleaner/utils/was-accessed-within-last-hour.test.ts
+++ b/packages/core/src/backend/features/cleaner/utils/was-accessed-within-last-hour.test.ts
@@ -17,6 +17,10 @@ describe('wasAccessedWithinLastHour', () => {
     });
   });
 
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
   it('should return true when file was modified within last hour', () => {
     // Given
     props.fileStats.mtimeMs = new Date('2025-09-19T11:00:00Z').getTime();
--- a/release/sign.ps1
+++ b/release/sign.ps1
@@ -36,5 +36,5 @@ $base64 = [System.Convert]::ToBase64String($bytes)
 Write-Host "Exe base64 hash: $base64"
 
 (Get-Content $yamlPath) `
-| ForEach-Object { $_ -replace '^(\s*sha512:\s*).+', "`$1$base64" } `
+| ForEach-Object { $_ -replace '^(\s*sha512:\s*).+', ('${1}' + $base64) } `
 | Set-Content $yamlPath
--- a/src/apps/main/background-processes/backups/BackupScheduler/backup-scheduler.test.ts
+++ b/src/apps/main/background-processes/backups/BackupScheduler/backup-scheduler.test.ts
@@ -17,10 +17,6 @@ describe('backup-scheduler', () => {
     BackupScheduler.stop();
   });
 
-  afterEach(() => {
-    vi.useRealTimers();
-  });
-
   it('should not schedule backups if there is no last backup', () => {
     // Given
     getMock.mockReturnValueOnce(-1);
--- a/src/apps/main/background-processes/backups/BackupScheduler/backup-scheduler.test.ts
+++ b/src/apps/main/background-processes/backups/BackupScheduler/backup-scheduler.test.ts
@@ -17,6 +17,10 @@ describe('backup-scheduler', () => {
     BackupScheduler.stop();
   });
 
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
   it('should not schedule backups if there is no last backup', () => {
     // Given
     getMock.mockReturnValueOnce(-1);
--- a/src/apps/main/electron/autoupdater/check-and-install.test.ts
+++ b/src/apps/main/electron/autoupdater/check-and-install.test.ts
@@ -1,14 +1,12 @@
-import { existsSync } from 'node:fs';
+import * as access from '@/infra/file-system/services/access';
 import { loggerFn } from '@/tests/vitest/mocks.helper.test';
-import { call, calls, deepMocked, partialSpyOn, TestProps } from '@/tests/vitest/utils.helper.test';
+import { call, calls, partialSpyOn, TestProps } from '@/tests/vitest/utils.helper.test';
 import { checkAndInstall } from './check-and-install';
 import * as showDialog from './show-dialog';
 import * as verifyHashModule from './verify-hash';
 
-vi.mock(import('node:fs'));
-
 describe('check-and-install', () => {
-  const existsSyncMock = deepMocked(existsSync);
+  const existsSyncMock = partialSpyOn(access, 'access');
   const verifyHashMock = partialSpyOn(verifyHashModule, 'verifyHash');
   const installReleaseMock = partialSpyOn(showDialog, 'installRelease');
 
@@ -17,13 +15,13 @@ describe('check-and-install', () => {
   };
 
   beforeEach(() => {
-    existsSyncMock.mockReturnValue(true);
+    existsSyncMock.mockResolvedValue(undefined);
     installReleaseMock.mockReturnValue(true);
   });
 
   it('should skip if file does not exist', async () => {
     // Given
-    existsSyncMock.mockReturnValue(false);
+    existsSyncMock.mockResolvedValue(new access.AccessError('NON_EXISTS'));
     // When
     const res = await checkAndInstall(props as any);
     // Then
--- a/src/apps/main/electron/autoupdater/check-and-install.test.ts
+++ b/src/apps/main/electron/autoupdater/check-and-install.test.ts
@@ -1,20 +1,19 @@
 import { existsSync } from 'node:fs';
 import { loggerFn } from '@/tests/vitest/mocks.helper.test';
 import { call, calls, deepMocked, partialSpyOn, TestProps } from '@/tests/vitest/utils.helper.test';
-import { checkExistingFile } from './check-existing-file';
+import { checkAndInstall } from './check-and-install';
 import * as showDialog from './show-dialog';
 import * as verifyHashModule from './verify-hash';
 
 vi.mock(import('node:fs'));
 
-describe('check-existing-file', () => {
+describe('check-and-install', () => {
   const existsSyncMock = deepMocked(existsSync);
   const verifyHashMock = partialSpyOn(verifyHashModule, 'verifyHash');
   const installReleaseMock = partialSpyOn(showDialog, 'installRelease');
 
-  const props: TestProps<typeof checkExistingFile> = {
+  const props: TestProps<typeof checkAndInstall> = {
     filePath: '/tmp/Internxt-Setup-10.0.0.exe',
-    latest: '10.0.0',
   };
 
   beforeEach(() => {
@@ -26,7 +25,7 @@ describe('check-existing-file', () => {
     // Given
     existsSyncMock.mockReturnValue(false);
     // When
-    const res = await checkExistingFile(props as any);
+    const res = await checkAndInstall(props as any);
     // Then
     expect(res).toBeUndefined();
     calls(verifyHashMock).toHaveLength(0);
@@ -34,7 +33,7 @@ describe('check-existing-file', () => {
 
   it('should install release if file exists and hash is valid', async () => {
     // When
-    const res = await checkExistingFile(props as any);
+    const res = await checkAndInstall(props as any);
     // Then
     expect(res).toBe(true);
     calls(verifyHashMock).toHaveLength(1);
@@ -46,7 +45,7 @@ describe('check-existing-file', () => {
     // Given
     verifyHashMock.mockRejectedValue(new Error('sha512 mismatch'));
     // When
-    const res = await checkExistingFile(props as any);
+    const res = await checkAndInstall(props as any);
     // Then
     expect(res).toBeUndefined();
     calls(installReleaseMock).toHaveLength(0);
--- a/src/apps/main/electron/autoupdater/check-and-install.ts
+++ b/src/apps/main/electron/autoupdater/check-and-install.ts
@@ -3,12 +3,12 @@ import { existsSync } from 'node:fs';
 import { installRelease } from './show-dialog';
 import { verifyHash } from './verify-hash';
 
-export async function checkExistingFile({ filePath, latest }: { filePath: string; latest: string }) {
+export async function checkAndInstall({ filePath }: { filePath: string }) {
   if (!existsSync(filePath)) return;
 
   try {
     logger.debug({ msg: 'Release already downloaded' });
-    await verifyHash({ filePath, latest });
+    await verifyHash({ filePath });
     return installRelease({ filePath });
   } catch (error) {
     logger.debug({ msg: 'Invalid downloaded release', error });
--- a/src/apps/main/electron/autoupdater/check-and-install.ts
+++ b/src/apps/main/electron/autoupdater/check-and-install.ts
@@ -1,10 +1,11 @@
 import { logger } from '@internxt/drive-desktop-core/build/backend';
-import { existsSync } from 'node:fs';
+import { access } from '@/infra/file-system/services/access';
 import { installRelease } from './show-dialog';
 import { verifyHash } from './verify-hash';
 
 export async function checkAndInstall({ filePath }: { filePath: string }) {
-  if (!existsSync(filePath)) return;
+  const error = await access(filePath);
+  if (error) return;
 
   try {
     logger.debug({ msg: 'Release already downloaded' });
--- a/src/apps/main/electron/autoupdater/check-existing-file.ts
+++ b/src/apps/main/electron/autoupdater/check-existing-file.ts
@@ -3,13 +3,13 @@ import { existsSync } from 'node:fs';
 import { showDialog } from './show-dialog';
 import { verifyHash } from './verify-hash';
 
-export async function checkExistingFile({ filePath, latestVersion }: { filePath: string; latestVersion: string }) {
+export async function checkExistingFile({ filePath, latest }: { filePath: string; latest: string }) {
   if (!existsSync(filePath)) return;
 
   try {
     logger.debug({ msg: 'Release already downloaded' });
-    await verifyHash({ filePath, latestVersion });
-    await showDialog({ filePath, latestVersion });
+    await verifyHash({ filePath, latest });
+    await showDialog({ filePath, latest });
   } catch (error) {
     logger.debug({ msg: 'Invalid downloaded release', error });
   }
--- a/src/apps/main/electron/autoupdater/check-existing-file.ts
+++ b/src/apps/main/electron/autoupdater/check-existing-file.ts
@@ -1,0 +1,16 @@
+import { logger } from '@internxt/drive-desktop-core/build/backend';
+import { existsSync } from 'node:fs';
+import { showDialog } from './show-dialog';
+import { verifyHash } from './verify-hash';
+
+export async function checkExistingFile({ filePath, latestVersion }: { filePath: string; latestVersion: string }) {
+  if (!existsSync(filePath)) return;
+
+  try {
+    logger.debug({ msg: 'Release already downloaded' });
+    await verifyHash({ filePath, latestVersion });
+    await showDialog({ filePath, latestVersion });
+  } catch (error) {
+    logger.debug({ msg: 'Invalid downloaded release', error });
+  }
+}
--- a/src/apps/main/electron/autoupdater/check-existing-file.ts
+++ b/src/apps/main/electron/autoupdater/check-existing-file.ts
@@ -9,7 +9,7 @@ export async function checkExistingFile({ filePath, latest }: { filePath: string
   try {
     logger.debug({ msg: 'Release already downloaded' });
     await verifyHash({ filePath, latest });
-    installRelease({ filePath });
+    return installRelease({ filePath });
   } catch (error) {
     logger.debug({ msg: 'Invalid downloaded release', error });
   }
--- a/src/apps/main/electron/autoupdater/check-existing-file.ts
+++ b/src/apps/main/electron/autoupdater/check-existing-file.ts
@@ -1,6 +1,6 @@
 import { logger } from '@internxt/drive-desktop-core/build/backend';
 import { existsSync } from 'node:fs';
-import { showDialog } from './show-dialog';
+import { installRelease } from './show-dialog';
 import { verifyHash } from './verify-hash';
 
 export async function checkExistingFile({ filePath, latest }: { filePath: string; latest: string }) {
@@ -9,7 +9,7 @@ export async function checkExistingFile({ filePath, latest }: { filePath: string
   try {
     logger.debug({ msg: 'Release already downloaded' });
     await verifyHash({ filePath, latest });
-    await showDialog({ filePath, latest });
+    installRelease({ filePath });
   } catch (error) {
     logger.debug({ msg: 'Invalid downloaded release', error });
   }
--- a/src/apps/main/electron/autoupdater/check-for-updates.ts
+++ b/src/apps/main/electron/autoupdater/check-for-updates.ts
@@ -1,0 +1,46 @@
+import { logger } from '@internxt/drive-desktop-core/build/backend';
+import { app } from 'electron';
+import { writeFile } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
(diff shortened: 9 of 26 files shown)


release/sign.ps1
```diff
@@ -36,5 +36,5 @@ $base64 = [System.Convert]::ToBase64String($bytes)
 Write-Host "Exe base64 hash: $base64"
 
 (Get-Content $yamlPath) `
-| ForEach-Object { $_ -replace '^(\s*sha512:\s*).+', "`$1$base64" } `
+| ForEach-Object { $_ -replace '^(\s*sha512:\s*).+', ('${1}' + $base64) } `
```
dajimenezriv-internxt, line 39: The build release script was sometimes generating wrongly the `latest.yml` because of the hash.

src/apps/main/electron/autoupdater/check-for-updates.ts
```diff
@@ -0,0 +1,43 @@
...
+      return;
+    }
+
+    const fileName = `Internxt-Setup-${latest}.exe`;
+    const filePath = join(tmpdir(), fileName);
+
+    logger.debug({ msg: 'New release available', latest, filePath });
+
+    const installing = await checkExistingFile({ latest, filePath });
+    if (installing) return true;
```
dajimenezriv-internxt, line 34: We want to check if the release was already downloaded to install it so we don't need to download it again, we just install it and open the app again.

packages/core/src/backend/features/cleaner/utils/was-accessed-within-last-hour.test.ts
```diff
@@ -17,10 +17,6 @@ describe('wasAccessedWithinLastHour', () => {
     });
   });
 
-  afterEach(() => {
-    vi.useRealTimers();
-  });
```
dajimenezriv-internxt, line 22: These are already removed with `clearMocks: true` in the vitest config.

sg-gs:
> I decided to implement it manually 

Can we rollback to `electron-autoupdater` if any issue arises or needed in the future? Which things are lost (the won ones are clear) by choosing this path? And how reliable is it in case of a download corruption?

sg-gs:
Adding @TamaraFinogina here to ensure there are no security regression issues here regarding authenticity, filesystem usage and security guarantees in general @dajimenezriv-internxt (against the current `electron-autoupdater` version)

dajimenezriv-internxt:
@sg-gs Regarding the first comment. In case the download is corrupted we are checking against the `hash`, so it will be retried the download in the next startup. What we lose in case something doesn't work is that the user will need to install the release manually or create another fast release going back to `electron-updater` (the implementation was very simple). Apart from that, we lose that we are not using the default electron update package, however, as pointed in the description it was working a bit strange (that happened in my laptop) and since we only use windows and github releases it wasn't difficult to implement it manually and we can add as much customization as we want. I think that we don't lose anything, I've been checking but I haven't found any better thing in `electron-updater`. The only thing is that it was checking against the blockmap and not downloading the whole file (however it was failing everytime and downloading the entire file in the end).

src/apps/main/electron/autoupdater/check-for-updates.ts
```diff
@@ -0,0 +1,43 @@
...
+import { downloadRelease } from './download-release';
+
+export async function checkForUpdates() {
+  if (!app.isPackaged) return;
+
+  try {
+    const res = await fetch('https://api.github.com/repos/internxt/drive-desktop/releases/latest');
+    const data = await res.json();
+    const release = data as { tag_name: string };
+    const latest = release.tag_name.replace(/^v/, '');
```
TamaraFinogina: Shall we add a check that the tag doesn't have `'../'` or some weird characters? I think we have a fixed name for all tags, but even just checking for that tag only has letters/numbers should do

```
if (!/^[0-9A-Za-z._-]+$/.test(latest)) {
  throw new Error("Wrong tag format");
}
```

TamaraFinogina: Actually, it's always with numbers v.X.Y.Z, so we can remove letters from the check

dajimenezriv-internxt:
About a comment made by @TamaraFinogina, she asked me why don't use `update-electron-app`. Is another library I considered but in the end I discarded it because it didn't include hash validation. However, I was wrong, that library uses internally `autoUpdater` from `electron` that already includes hash validation. So, I tried the library:

```ts
    updateElectronApp({
      updateSource: {
        type: UpdateSourceType.ElectronPublicUpdateService,
        repo: 'internxt/drive-desktop',
      },
      updateInterval: '5m',
      logger: {
        error(message) {
          logger.debug({ msg: message });
        },
        warn(message) {
          logger.debug({ msg: message });
        },
        info(message) {
          logger.debug({ msg: message });
        },
        log(message) {
          logger.debug({ msg: message });
        },
      },
    });
```

but it gave me some issues:

1. We need to try the app packaged to test anything (since they do a check of `app.isPackaged` inside the library). This happen with `electron-updater` and this library, which is another reason why I discarded `electron-updater` in the first place. Every test needed the build of the app (it takes around 250s).
2. We have been building the app using the target `nsis` but `update-electron-app` requires `squirrel`:
```
[2026-04-27 10:40:23.845] { header: '  - b -     ', msg: 'feedURL' }
[2026-04-27 10:40:23.846] { header: '  - b -     ', msg: 'requestHeaders' }
[2026-04-27 10:40:23.848] { header: '  - b -     ', msg: 'updater error' }
[2026-04-27 10:40:24.415] {
  header: '  - b -     ',
  msg: Error: Can not find Squirrel
      at AutoUpdater.checkForUpdates (node:electron/js2c/browser_init:2:9080)
      at initUpdater (C:\Users\dajim\AppData\Local\Programs\internxt-drive\resources\app.asar\dist\main\webpack:\internxt-drive\node_modules\update-electron-app\dist\index.js:125:28)
      at EventEmitter.<anonymous> (C:\Users\dajim\AppData\Local\Programs\internxt-drive\resources\app.asar\dist\main\webpack:\internxt-drive\node_modules\update-electron-app\dist\index.js:53:42)
      at EventEmitter.emit (node:events:520:35)
}
```

Btw, `squirrel` is deprecated in `electron-builder`.
The manual implementation is still simple, we can make it very customizable (for example, to force the install, if the user doesn't want to install manually when the update check, the install is the first thing the app does when opening again) and we can test with it in development.

.github/workflows/check-pr-size.yml
```diff
@@ -7,8 +7,9 @@ on:
 jobs:
   check_pr_size:
     # Ignore PRs with name merge-X-X-X-release
-    if: |
-      !(startsWith(github.head_ref, 'merge-') && endsWith(github.head_ref, '-release') && github.base_ref == 'main')
+    # if: |
+    #   !(startsWith(github.head_ref, 'merge-') && endsWith(github.head_ref, '-release') && github.base_ref == 'main')
+    if: false
```
TamaraFinogina, line 12: Is the PR size checker turned off for all?

dajimenezriv-internxt: Yes, because there are many tests and the lines were 550, to not split a test into another PR. It will be reverted in the next PR.

packages/core/src/backend/features/cleaner/utils/was-accessed-within-last-hour.test.ts
TamaraFinogina: Hmm, I think `clearMocks: true` calls [vi.clearAllMocks()](https://vitest.dev/api/vi.html#vi-clearallmocks) before each test. But `vi.clearAllMocks()` is not calling vi.useRealTimers, so timers won't be reset cause they have a separate API

src/apps/main/background-processes/backups/BackupScheduler/backup-scheduler.test.ts
```diff
@@ -17,10 +17,6 @@ describe('backup-scheduler', () => {
     BackupScheduler.stop();
   });
 
-  afterEach(() => {
-    vi.useRealTimers();
-  });
-
   it('should not schedule backups if there is no last backup', () => {
```
TamaraFinogina, line 24: same, I don't think timers are affected by clearMocks

src/apps/main/electron/autoupdater/check-existing-file.ts
```diff
@@ -0,0 +1,16 @@
...
+import { existsSync } from 'node:fs';
+import { installRelease } from './show-dialog';
+import { verifyHash } from './verify-hash';
+
+export async function checkExistingFile({ filePath, latest }: { filePath: string; latest: string }) {
+  if (!existsSync(filePath)) return;
+
+  try {
+    logger.debug({ msg: 'Release already downloaded' });
+    await verifyHash({ filePath, latest });
```
TamaraFinogina: No version downgrade prevention?

src/apps/main/electron/autoupdater/check-and-install.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    // Given
+    existsSyncMock.mockReturnValue(false);
+    // When
+    const res = await checkExistingFile(props as any);
+    // Then
+    expect(res).toBeUndefined();
+    calls(verifyHashMock).toHaveLength(0);
+  });
+
+  it('should install release if file exists and hash is valid', async () => {
```
TamaraFinogina, line 32: Can we add a test for trying to install an inferior version to what the user already has? Say the user has 11.0.0 and we try to install 10.0.0

src/apps/main/electron/autoupdater/check-for-updates.ts
```diff
@@ -0,0 +1,48 @@
...
+  tag_name: z.string().regex(/^v\d+\.\d+\.\d+$/, 'tag_name must match vX.X.X'),
+});
+
+export async function checkForUpdates() {
+  if (!app.isPackaged) return;
+
+  try {
+    const res = await fetch('https://api.github.com/repos/internxt/drive-desktop/releases/latest');
+    const data = await res.json();
+    const { tag_name } = ReleaseSchema.parse(data);
```
TamaraFinogina: Maybe use the `name`, so we don't have to remove v? Cause I think name and tag_name are the same except for v

Image

src/apps/main/electron/autoupdater/check-for-updates.ts
```diff
@@ -0,0 +1,48 @@
...
+    const latest = tag_name.replace(/^v/, '');
+
+    if (!isNewer(INTERNXT_VERSION, latest)) {
+      logger.debug({ msg: 'App is up to date', latest });
+      setTimeout(checkForUpdates, 60 * 60 * 1000);
```
TamaraFinogina, line 24: existsSync (called inside checkForUpdates) will pose the app. Are we sure we want to do the periodic checks? 

Link: https://www.geeksforgeeks.org/node-js/node-js-fs-existssync-method/

TamaraFinogina: In other words, I don't know how the file upload/download in progress will react to this interruption. @dajimenezriv-internxt will it be ok?

src/apps/main/electron/autoupdater/check-existing-file.ts
```diff
@@ -0,0 +1,16 @@
+import { logger } from '@internxt/drive-desktop-core/build/backend';
+import { existsSync } from 'node:fs';
+import { installRelease } from './show-dialog';
+import { verifyHash } from './verify-hash';
+
+export async function checkExistingFile({ filePath, latest }: { filePath: string; latest: string }) {
```
TamaraFinogina: technically, it's check and install

src/apps/main/electron/autoupdater/show-dialog.ts
```diff
@@ -0,0 +1,25 @@
...
+import { spawn } from 'node:child_process';
+import { iconPath } from '@/apps/utils/icon';
+
+export async function showDialog({ filePath, latest }: { filePath: string; latest: string }) {
+  const { response } = await dialog.showMessageBox({
+    type: 'info',
+    icon: nativeImage.createFromPath(iconPath),
+    title: 'Update Available',
+    message: `Version ${latest} is available`,
+    detail: 'Download and install the update now?',
```
TamaraFinogina: At this point, it's already downloaded

src/apps/main/electron/autoupdater/verify-hash.ts
```diff
@@ -0,0 +1,21 @@
...
+import { createHash } from 'node:crypto';
+import { readFile } from 'node:fs/promises';
+
+export async function verifyHash({ filePath, latest }: { filePath: string; latest: string }) {
+  const data = await readFile(filePath);
+  const actual = createHash('sha512').update(data).digest('base64');
+
+  logger.debug({ msg: 'Verifying release hash', actual });
+
+  const url = `https://github.com/internxt/drive-desktop/releases/download/v${latest}/latest.yml`;
```
TamaraFinogina: I think if we use api.github.com, we can get the latest without giving the release number

https://api.github.com/repos/internxt/drive-desktop/releases/latest

TamaraFinogina: https://github.com/internxt/drive-desktop/releases/latest/download/latest.yml

Wait, I think this one will work without a version number.

src/apps/main/token-scheduler/token-scheduler.test.ts
```diff
@@ -22,10 +22,6 @@ describe('token-scheduler', () => {
     vi.setSystemTime(new Date(0));
   });
 
-  afterEach(() => {
-    vi.useRealTimers();
```
TamaraFinogina, line 26: same about timers, don't think clearMocks affects  vi.useFakeTimers();

src/backend/features/backups/download/download-backup.test.ts
```diff
@@ -30,10 +30,6 @@ describe('download-backup', () => {
     vi.setSystemTime(new Date('2025-01-01T12:00:00Z'));
   });
 
-  afterAll(() => {
-    vi.useRealTimers();
```
TamaraFinogina, line 34: same

src/backend/features/local-sync/sync-state/services/add-item.test.ts
```diff
@@ -15,11 +15,6 @@ describe('add-item', () => {
     clearStore();
   });
 
-  afterEach(() => {
-    vi.clearAllTimers();
```
TamaraFinogina, line 19: same

.eslintrc.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
         'sonarjs/no-alphabetical-sort': 'off',
         'sonarjs/no-empty-test-file': 'off',
         'sonarjs/os-command': 'off',
+        'sonarjs/publicly-writable-directories': 'off',
```
TamaraFinogina, line 31: @dajimenezriv-internxt It's not fixing the Sonar warnings; it just disables them. Sonar keeps complaining but silently :)

dajimenezriv-internxt: It's because we are using the test path /tmp inside the tests, however the rule is only disabled inside the tests. The warnings inside the production code are not disabled.

dajimenezriv-internxt: https://github.com/internxt/drive-desktop/pull/1361/changes/BASE..e8464948dd462cae74b1bf650b8da0dd25dcbb08#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5L21

src/apps/main/electron/autoupdater/verify-hash.ts
dajimenezriv-internxt: True, it also applies to the `.exe`, I'm changing it.

src/apps/main/electron/autoupdater/check-existing-file.ts
dajimenezriv-internxt: It's checked here: https://github.com/internxt/drive-desktop/pull/1361/changes#diff-7413da9f63c5e485ce84aadfe7ab7bd545dc82013be911cd74f936a01b34025aR43

src/apps/main/electron/autoupdater/check-and-install.test.ts
dajimenezriv-internxt: It cannot reach this code, as explained in the above comment.

src/apps/main/electron/autoupdater/download-release.ts
```diff
@@ -0,0 +1,32 @@
...
+import { showDialog } from './show-dialog';
+import { verifyHash } from './verify-hash';
+
+export async function downloadRelease({ fileName, filePath, latest }: { fileName: string; filePath: string; latest: string }) {
+  try {
+    logger.debug({ msg: 'Downloading release', latest });
+
+    const time = await measurePerfomance(async () => {
+      const url = `https://github.com/internxt/drive-desktop/releases/download/v${latest}/${fileName}`;
+      const res = await fetch(url);
```
TamaraFinogina, line 16: @dajimenezriv-internxt Can you try this?


```
  const res = await fetch(`https://api.github.com/repos/${owner}/${repo}/releases/latest`, {
    headers: { 'Accept': 'application/vnd.github+json' }
  });

  if (!res.ok) throw new Error(`HTTP ${res.status}`);

  const release = await res.json();

  if (release.draft || release.prerelease) throw new Error('Latest release is not a stable release');

  const asset = release.assets.find((a: { name: string }) =>
    /^Internxt-Setup-[\d.]+\.exe$/.test(a.name)
  );

  if (!asset) throw new Error(`No setup .exe found in release ${release.tag_name}`);

  const response = await fetch(asset.browser_download_url);
```

src/apps/main/electron/autoupdater/check-for-updates.ts
dajimenezriv-internxt: Technically it's ok, also because it's something that we just do every hour, however I've been investigating more and probably we should `access` instead. I thought that it wasn't an async version because the `existsSync` was so fast than the overhead of making it async was not worth. It happens that with `better_sqlite3` for example. I'm going to change it here and I will do the same change in the loop that iterates the whole file explorer in the next PR since there it will be more useful.

src/apps/main/electron/autoupdater/download-release.ts
dajimenezriv-internxt: What's the benefit? Because the filename is constructed from the `latest.yml`? The same as the `browser_download_url`. So technically if an attacker achieves to inject a different `filename` he should be able to inject the asset.browser_download_url` too, right?

dajimenezriv-internxt: What we can do is to name all `.exe` as `Internxt-Setup.exe` and we have a fixed name and it doesn't change per release. However, I think that most open source projects include the version in the name as a standard.

.eslintrc.js
TamaraFinogina, line 31: @dajimenezriv-internxt We should return it, it's a useful check.

We can always pass the `latest` as input and just do the path at the moment. 


```
  const filePath = join(tmpdir(), `Internxt-Setup-${latest}.exe`);
```

dajimenezriv-internxt: The eslint rule doesn't apply there, it's just for the test files (https://github.com/internxt/drive-desktop/pull/1361/changes#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5L21) because I'm writing `/tmp` inside a test: https://github.com/internxt/drive-desktop/pull/1361/changes#diff-ab5bba20517d999790e6182db1ccbb8a9d39f9dad2951ddcb9b2df62087e7457R13.

src/apps/main/electron/autoupdater/check-for-updates.ts
```diff
@@ -0,0 +1,47 @@
...
+      return;
+    }
+
+    const fileName = `Internxt-Setup-${latest}.exe`;
+    const filePath = join(tmpdir(), fileName);
+
+    logger.debug({ msg: 'New release available', latest, filePath });
+
+    const installing = await checkAndInstall({ filePath });
```
TamaraFinogina, line 33: Ideally, I would bring hash and installation on top, something like:

```
const alreadyDownloaded = await checkIfAlreadyDownloaded({ filePath });
if(!alreadyDownloaded) downloadRelease({fileName});
await verifyHash({ filePath });
const installNow = await showDialog({ latest });
if(installNow) installRelease({ filePath });
```
Now installation happens in two different points: with user confirmation (immediately after download) and without (when already downloaded). I guess it's on purpose. 

Anyway, if we ever need to debug a client issue with updates, we will need to know in which path it happened, so the support will have to go back and forth asking for details (did the user press later or install, etc).

.eslintrc.js
TamaraFinogina: Never mind, you are right, it's for tests only

src/apps/main/electron/autoupdater/check-for-updates.ts
dajimenezriv-internxt: 1. We already know the path it took because of the logs. I'm going to add a log for the dialog response.
2. With the code you have provided we are not forcing the installation in the startup, you always show the dialog. I want to force the installation when the user opens the app.

TamaraFinogina: If the user selects 'later', in an hour there will be another check, and it will be installed, right?

dajimenezriv-internxt: Nop, the timeout only runs if there wasn't any update, so we check again every hour. If the user clicks on `Later` we don't check again. See that the `setTimeout` only runs when the version is up to date. It will be installed after the user closes the app and opens it again (when turning off and on the laptop for example).

TamaraFinogina:
```
const alreadyDownloaded = await checkIfAlreadyDownloaded({ filePath });
if (!alreadyDownloaded) downloadRelease({ fileName });
await verifyHash({ filePath });
const installNow = alreadyDownloaded ? true: await showDialog({ latest });
if (installNow) installRelease({ filePath });

```
Btw, this code will do what you need, but it's ok

dajimenezriv-internxt: It won't, because `downloadRelease` needs to be awaited and that will block the main thread, that's why we need to write twice the `verifyHash` and so. We want to check if the release is downloaded and install before doing anything. Otherwise, we want to download and so in background.

TamaraFinogina (dismissed):

src/apps/main/electron/autoupdater/check-for-updates.ts
TamaraFinogina: ok, then it's fine

dajimenezriv-internxt:
If you can do a final review with the changes @sg-gs

src/infra/file-system/services/access.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+describe('access', () => {
+  it('should return undefined if file exists', async () => {
+    // When
+    const error = await access(join(cwd(), 'package.json'));
+    // Then
+    expect(error).toBeUndefined();
+  });
+
+  it('should return NON_EXISTS if file does not exist', async () => {
```
sg-gs, line 13: One of these two legends (this and the below one) should be wrong, as it cannot behave in different ways for the same case

src/apps/main/main.ts
```diff
@@ -59,8 +59,7 @@ if (process.defaultApp) {
 const gotTheLock = app.requestSingleInstanceLock();
 
 if (!gotTheLock) {
-  app.quit();
-  process.exit(0);
+  app.exit(0);
```
sg-gs, line 62: Does this kill the process also?

src/apps/main/electron/autoupdater/verify-hash.ts
```diff
@@ -0,0 +1,21 @@
+import { logger } from '@internxt/drive-desktop-core/build/backend';
+import { createHash } from 'node:crypto';
+import { readFile } from 'node:fs/promises';
+
+export async function verifyHash({ filePath }: { filePath: string }) {
+  const data = await readFile(filePath);
+  const actual = createHash('sha512').update(data).digest('base64');
```
sg-gs: For an app of a couple hundred MBs, it's better to use a stream (pipeline(readable, hasher)) rather than loading the entire binary content in memory

src/apps/main/electron/autoupdater/check-and-install.ts
```diff
@@ -0,0 +1,17 @@
...
+export async function checkAndInstall({ filePath }: { filePath: string }) {
+  const error = await access(filePath);
+  if (error) return;
+
+  try {
+    logger.debug({ msg: 'Release already downloaded' });
+    await verifyHash({ filePath });
+    return installRelease({ filePath });
+  } catch (error) {
+    logger.debug({ msg: 'Invalid downloaded release', error });
```
sg-gs, line 15: Should we run a cleanup of the bad installer in case this fails to avoid accumulating binary content on the tmp dir every hour?

src/apps/main/electron/autoupdater/show-dialog.ts
```diff
@@ -0,0 +1,28 @@
...
+import { iconPath } from '@/apps/utils/icon';
+
+export async function showDialog({ filePath, latest }: { filePath: string; latest: string }) {
+  const { response } = await dialog.showMessageBox({
+    type: 'info',
+    icon: nativeImage.createFromPath(iconPath),
+    title: 'Update Available',
+    message: `Version ${latest} is available`,
+    detail: 'Install the update now?',
+    buttons: ['Update Now', 'Later'],
```
sg-gs, line 13: We may be interested in forcing updates always. We do not want clients to use problematic apps if something happens, it's preferable to release a new one and let the autoupdater autoupdate directly

src/apps/main/main.ts
dajimenezriv-internxt: In `electron` technically we cannot really close inmediatly a process, but yes, `app.exit(0)` is the closest to terminate the process. `process.exit(0)` doesn't really terminate the process either.

src/apps/main/electron/autoupdater/check-and-install.ts
dajimenezriv-internxt: Not needed, it overrides the file every time. Technically the `temp` folder in windows should do a cleanup of the files periodically so we won't save many version files (at least the theory). I can change to always be named as `Internxt-Setup.exe` so it's always override, as you say.

src/apps/main/electron/autoupdater/show-dialog.ts
dajimenezriv-internxt: I would like to, but if the user has the `Settings` window open or doing a backup or something it will stop in the middle and maybe is not the best UX. For me it would be the simplest, so I can change it. But not sure if what we want. Anyway, we can change between both in anytime.

src/apps/main/electron/autoupdater/check-and-install.ts
TamaraFinogina: @dajimenezriv-internxt I like it. With a fixed `Internxt-Setup.exe`, we won't need parameters for verifyHash and insallRelease and Sonar won't complain :)

dajimenezriv-internxt:
Hey, can you do a last check of the PR? @sg-gs @TamaraFinogina

TamaraFinogina (approved):

PR #1362, opened 2026-04-30
Move folder instead of delete

dajimenezriv-internxt:
## What

Currently when a folder was deleted in cloud we were deleting it in local using `trash`. However, this had 2 issues:
1. if the folder contained items that were not placeholders we were sending them to the trash when deleting the folder.
2. if the folder was big it could lead to a lot of `delete` events that could imply deleting files in the cloud (however just the root folder was deleted).

To solve the first issue we have implemented a C++ solution that iterates the folder and checks for non placeholder items. I've added a test that tests against an empty folder, a folder with all placeholders (dehydrated and hydrated) and also with a file that is not a placeholder.

To solve the second issue first we move the folder to a temporal folder in the same volume called `C:/.internxt_trash` and then we send it to the trash, so we just receive one `delete` event.


packages/addon/include/virtual_drive/get_first_non_placeholder.h
```diff
@@ -0,0 +1,61 @@
+#pragma once
```
dajimenezriv-internxt, line 1: Tested with a folder with 500 items and took 0.02s.

src/backend/features/remote-sync/file-explorer/delete-item-placeholder.test.ts
```diff
@@ -1,58 +1,91 @@
...
 import { deleteItemPlaceholder } from './delete-item-placeholder';
 
+vi.mock(import('node:fs/promises'));
+vi.mock(import('node:crypto'));
 vi.mock(import('trash'));
 
 describe('delete-item-placeholder', () => {
+  const mkdirMock = vi.mocked(mkdir);
+  const renameMock = vi.mocked(rename);
+  const rmMock = vi.mocked(rm);
```
egalvis27: Has the rule against using vi.mock been abolished? Because here you could use deepMocked to mock the functions in fs/promises.

src/backend/features/remote-sync/file-explorer/delete-item-placeholder.test.ts
```diff
@@ -1,58 +1,91 @@
...
+      type: 'file',
+    };
   });
 
   it('should skip if local item does not exist', async () => {
     // Given
     props.locals = new Map();
     // When
-    await deleteItemPlaceholder(props);
+    await deleteItemPlaceholder(props as any);
```
egalvis27, line 44: Why is this type assert necessary now when it wasn't necessary before?

src/backend/features/remote-sync/file-explorer/delete-item-placeholder.test.ts
dajimenezriv-internxt: True, you are right, changed.

src/backend/features/remote-sync/file-explorer/delete-item-placeholder.test.ts
dajimenezriv-internxt: I'm a trying with a new way by using `TestProps` in the `let props` instead of using `mockProps`. So now the type is `Partial` and needs to casting to `any`.

egalvis27 (approved):

PR #1365, opened 2026-04-30
[PB-6101]: feat/integrate-sentry-1

larryrider:
- Adds sentry integration


src/apps/shared/sentry/sentry.ts
```diff
@@ -0,0 +1,215 @@
...
+import { logger } from '../logger/logger';
+
+let isInitialized = false;
+
+export function getSentryEnvironment(): string {
+  if (process.env.NODE_ENV === 'production') {
+    return process.env.SENTRY_ENVIRONMENT || 'prod';
+  }
+  return process.env.SENTRY_ENVIRONMENT || 'dev';
+}
```
dajimenezriv-internxt: Can it be in zod `SENTRY_ENVIRONMENT` to be just `dev` or `prod`: `z.union([z.literal('dev'), z.literal('prod')])` and we ensure here that they are already set with that value, because you put 'min(1)' so I think this 2 conditions of `||` cannot happen since they at least contain one character.

src/apps/shared/sentry/sentry.ts
```diff
@@ -0,0 +1,215 @@
...
+
+export function initSentry(): void {
+  const dsn = process.env.SENTRY_DSN;
```
dajimenezriv-internxt: I think the same happens here, you put `min(1)` in zod, maybe it should be optional?

src/apps/shared/sentry/sentry.ts
larryrider: true 👍

src/apps/shared/sentry/sentry.ts
```diff
@@ -0,0 +1,215 @@
...
+  setUser(null);
+
+  logger.debug({ msg: 'Sentry user context cleared' });
+}
+
+export async function captureSentryException(error: unknown, extra?: Record<string, unknown>): Promise<void> {
+  if (!isInitialized) return;
+  captureException(error, { extra });
+}
+
```
dajimenezriv-internxt: I see that all sentry errors are `tags` which you use a common tag and then one or more extras. Why not combinate this with the logger? Something like this: `drive-desktop\src\apps\shared\logger\logger.ts`. We add a new function called `sentry`, or `sentryError` and so. For the tag we use the `msg`. The `msg` is always a fixed value that never concatenates with anything. So the error message it's always the same. And for the extras you put the rest of fields of the log. So, when logging we also send the sentry error. In case you want more extra that the log includes maybe we can add directly to the logs, so they are logged and also sent to sentry.

dajimenezriv-internxt: In case you need an special property for the tag, you can add it to function `logger.sentry({ tag: "whatever" })` instead of using the `msg` and we then use the `tag` for the `msg` log. So, we won't need as many functions to send every type of sentry error and also converting a log to a sentry error would be as easy as changing from `logger.error` to `logger.sentryError` for example.

dajimenezriv-internxt: Myabe even concatenate the `msg` with `[Sentry] ${msg}` or whatever, in case we want to differenciate the message, so we don't need the `beforeSend` message and we don't get like "duplicated" logs, and just one log includes everything.

larryrider: it looks good to me 👍

packages/core/src/backend/core/logger/logger.ts
```diff
@@ -9,6 +11,7 @@ export type TLoggerBody = {
   msg: string;
   workspaceId?: string;
   context?: Record<string, unknown>;
+  error?: unknown;
```
dajimenezriv-internxt: I think that this is not needed, since the `[key: string]: unknown` already includes this line.

packages/core/src/backend/core/sentry/sentry.ts
```diff
@@ -0,0 +1,86 @@
+import { init, setUser, captureException } from '@sentry/electron/main';
+import { arch, release, version } from 'node:os';
+
+import { logger } from '../logger/logger';
+
+let isInitialized = false;
+
+export function captureSentryException(exception: unknown, options?: Record<string, unknown>): void {
```
dajimenezriv-internxt: Remember that in windows we don't use return types, same for the other functions.

packages/core/src/backend/core/sentry/sentry.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  init({
+    dsn,
+    environment,
+    release: process.env.INTERNXT_VERSION,
+    defaultIntegrations: false,
+    integrations: (integrations) => {
+      return integrations;
+    },
+    beforeSend(event) {
```
dajimenezriv-internxt: Is this needed? Because I think it's going to fill the whole logger with a lot of errors. Why is this needed?

packages/core/src/backend/core/logger/logger.ts
```diff
@@ -98,8 +101,23 @@ function error(rawBody: TLoggerBody) {
   return new Error(rawBody.msg, { cause: rawBody.exc });
 }
 
+function sentryError(rawBody: TLoggerBody, sentryExtras?: Record<string, unknown>) {
+  const err = error(rawBody);
+
+  if (process.type === 'browser') {
+    const { msg, error, ...rest } = rawBody;
```
dajimenezriv-internxt: Since you want to pass an `error` and to ensure that we pass it as `error` and not `exc` or `err`, maybe you can extend the type `TLoggerBody` with `LoggerErrorBody` or similar that must contain an `error: unknown`. The `T` before `LoggerBody` is because previously we were naming types as `T` and interfaces as `I`. No longer needed.

src/backend/features/sync/actions/services/upload-file.ts
```diff
@@ -39,6 +39,11 @@ export async function uploadFile({ ctx, path }: Props) {
   } catch (error) {
     if (isBottleneckStop({ error })) return;
 
+    ctx.logger.sentryError(
+      { msg: 'Error uploading file', path },
+      { error, fileUuid: '', fileSize: size, sourcePath: path, uploadSource: 'sync-upload' },
+    );
```
dajimenezriv-internxt: You don't need this error here, you can search for the existing error with msg: `Failed to upload file to the bucket`. Also, you know if it comes from sync or backups because it has a `TAG` which is `BACKUPS` or `SYNC`. So, technically the `uploadSource` is not needed.

src/backend/features/remote-sync/file-explorer/update-file-placeholder.ts
```diff
@@ -54,6 +54,9 @@ export async function updateFilePlaceholder({ ctx, remote, files, isFirstExecuti
       }
     }
   } catch (error) {
-    ctx.logger.error({ msg: 'Error updating file placeholder', path, error });
+    ctx.logger.sentryError(
+      { msg: 'Error updating file placeholder', path, error },
+      { uuid: remote.uuid, type: 'file', operationType: 'update' },
```
dajimenezriv-internxt: What is the benefit of `type` or `operationType`? Arent's these included in the `msg`? The same question for the other logs. Basically because since they have a cost, maybe we can reduce it by storing less size if the don't provide any new information.

src/backend/features/remote-sync/file-explorer/update-folder-placeholder.ts
```diff
@@ -34,7 +34,10 @@ export async function updateFolderPlaceholder({ ctx, remote, folders }: Props) {
     await checkIfMoved({ ctx, type: 'folder', remote, localPath: local.path });
     return true;
   } catch (exc) {
-    ctx.logger.error({ msg: 'Error updating folder placeholder', path, exc });
+    ctx.logger.sentryError(
+      { msg: 'Error updating folder placeholder', path },
```
dajimenezriv-internxt: Here you have deleted the error from the logs, so we won't see it in the log file.

src/apps/sync-engine/download-contents/download-contents.ts
```diff
@@ -60,7 +60,15 @@ export async function downloadContents({
       return;
     }
 
-    ctx.logger.error({ msg: 'Error downloading file', path, size: file.size, chunk: chunk?.length, error });
+    ctx.logger.sentryError(
+      { msg: 'Error downloading file', path, error },
```
dajimenezriv-internxt: Add here also the `chunk.length` since it's the file was corrupted we want to know if the size was bigger, lower or what happened.

src/apps/main/main.ts
```diff
@@ -89,11 +98,11 @@ if (process.env.NODE_ENV === 'production') {
 process.on('unhandledRejection', (error, promise) => {
   if (isAbortError({ error })) return;
 
-  logger.error({ msg: 'Unhandled rejection', error, promise });
+  logger.sentryError({ msg: 'Unhandled rejection', error }, { promise, type: 'unhandledRejection' });
```
dajimenezriv-internxt: You need to put the `promise` inside the first prop, otherwise we won't see it in the log file. Also maybe the `type` is not needed since it's already included in the `msg`.

src/backend/features/sync/actions/services/upload-file.ts
```diff
@@ -39,6 +39,8 @@ export async function uploadFile({ ctx, path }: Props) {
   } catch (error) {
     if (isBottleneckStop({ error })) return;
 
+    ctx.logger.sentryError({ msg: 'Error uploading file', path, error }, { fileSize: size });
```
dajimenezriv-internxt, line 42: Maybe this error can be placed better here: `drive-desktop\src\infra\inxt-js\file-uploader\process-error.ts` since it also logs the retries errors so we know if something it's failing related with the infra.

src/apps/sync-engine/download-contents/download-contents.ts
dajimenezriv-internxt: Also the `size`.

src/backend/features/sync/actions/services/upload-file.ts
larryrider: Its a good idea but lets iterate and then we decide if its worth including it there or not 🤔

dajimenezriv-internxt (dismissed):

dajimenezriv-internxt (approved):

PR #1366, opened 2026-04-30
[PB-6101]: feat/integrate-sentry-2

larryrider:
- Adds sentry fingerprints with TTL in order to not duplicate requests and avoid spamming sentry


packages/core/src/backend/core/sentry/sentry.ts
```diff
@@ -5,9 +5,42 @@ import { logger } from '../logger/logger';
...
+}
+
+function shouldCaptureSentryError(fingerprint: string): boolean {
+  const now = Date.now();
+
+  for (const [key, timestamp] of sentryDedupCache) {
+    if (now - timestamp >= DEDUP_TTL_MS) {
+      sentryDedupCache.delete(key);
+    }
+  }
```
dajimenezriv-internxt, line 23: Why not just checking the key instead of iterating all? With the current approach any time that you try to capture an error message it iterates the whole Map. You could do something like this:

```ts
// get sentry dedup entry
// check if sentry dedup entry timestamp is lower than Date.now() + DEDUP_TTL_MS
// if is lower then ignore it (I wouldn't log it because it will add a lot of logs)
// if it higher just update the timestamp
```

packages/core/src/backend/core/sentry/sentry.ts
```diff
@@ -5,9 +5,42 @@ import { logger } from '../logger/logger';
...
+  if (sentryDedupCache.has(fingerprint)) {
+    logger.debug({ msg: 'Sentry exception throttled (duplicate)', ttlMs: DEDUP_TTL_MS, fingerprint });
+    return false;
+  }
+
+  sentryDedupCache.set(fingerprint, now);
+  return true;
+}
+
 export function captureSentryException(exception: unknown, options?: Record<string, unknown>) {
```
dajimenezriv-internxt: Why not use for the fingerprint just the `msg` of the `error` that is unique or maybe a combination of `${tag}-${msg}`, so you we can't skip all `instanceof Error` and so.

dajimenezriv-internxt: We can make the `captureSentryException` to receive the `rawBody: LoggerSentryErrorBody` and take it the `msg` or `${tag}-${msg}` for the fingerprint since the `msg` is always defined.

packages/core/src/backend/core/sentry/sentry.ts
larryrider: the idea is to remove the old ones too, not only the one being inserted. 

In that case all timestamps will grow that cache we dont know how much, it could take some important space if we do not clear the outdated items 🤔

dajimenezriv-internxt: It depends on the combination of keys. If we just use the `msg` as key there are maybe 300 unique keys, not more. So the cache will be very little. The thing is, if we continue using sentry in other places (as I assume we would do), there are errors that happen in loops of millions of items (when refreshing placeholders every 10min for example), so checking the whole map every time we need to add a error log can be a lot. And this is not as unusual as it seems. For example, when we had the error of paths longer than 255 characters we were filling the log of these paths every 10 min or less.

larryrider: oh i see, lets go for your version then 👍

dajimenezriv-internxt (approved):

PR #1386, opened 2026-05-18
Fix rename operation

dajimenezriv-internxt:
## What

Some days ago I made this PR (https://github.com/internxt/drive-desktop/pull/1380) simplifying the move operation, however, for some reason, I forgot about the rename operation. I've reverted the changes but used the `parentUuid` that we have available in the in memory file explorer instead of getting it through C++.

### DIFF
--- a/src/backend/features/remote-sync/file-explorer/needs-to-be-moved.test.ts
+++ b/src/backend/features/remote-sync/file-explorer/needs-to-be-moved.test.ts
@@ -1,0 +1,77 @@
+import { FolderUuid } from '@/apps/main/database/entities/DriveFolder';
+import { abs } from '@/context/local/localFile/infrastructure/AbsolutePath';
+import { mockProps } from '@/tests/vitest/utils.helper.test';
+import { needsToBeMoved } from './needs-to-be-moved';
+
+describe('needs-to-be-moved', () => {
+  const remotePath = abs('/drive/folder1/current');
+  const localPath = abs('/drive/folder2/current');
+
+  let props: Parameters<typeof needsToBeMoved>[0];
+
+  beforeEach(() => {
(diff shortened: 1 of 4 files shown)


src/backend/features/remote-sync/file-explorer/needs-to-be-moved.ts
```diff
@@ -0,0 +1,22 @@
...
+export function needsToBeMoved({ remote, local }: Props) {
+  if (remote.absolutePath === local.path) return false;
+
+  const remoteParentPath = dirname(remote.absolutePath);
+  const localParentPath = dirname(local.path);
+
+  const isRenamed = remoteParentPath === localParentPath;
+  if (isRenamed) return true;
+
+  const isMoved = remote.parentUuid !== local.parentUuid;
```
dajimenezriv-internxt, line 20: This is simplified: https://github.com/internxt/drive-desktop/pull/1380/changes#diff-311f6f8e82840b992f6987d8ca511827ef98afec812d96bc6d8c23cfb2bbb01fL22

src/backend/features/remote-sync/file-explorer/needs-to-be-moved.test.ts
```diff
@@ -0,0 +1,77 @@
+import { FolderUuid } from '@/apps/main/database/entities/DriveFolder';
```
dajimenezriv-internxt, line 1: This is the same test we had before: https://github.com/internxt/drive-desktop/blob/40120fd87b05d4b7676c9b51f613ba952c34eeba/src/backend/features/remote-sync/file-explorer/needs-to-be-moved.test.ts

AlexisMora (approved):

PR #1391, opened 2026-05-21
Add addon C#

dajimenezriv:
## What

Implement the first version of Addon C# for hydrate/dehydrate file and also the watcher. The first 2 functions are very easy and they are working as expected. About the watcher it's the feature more tested in the javascript part and all events are working perfectly. In case something goes wrong we can deactivate the new `AddonCs` and use the `addon`.


src/apps/main/preload/ipc-main.ts
```diff
@@ -52,7 +52,6 @@ export function setupLoggedPreloadIpc({ ctx }: { ctx: AuthContext }) {
   ipcPreloadMain.handle('driveChooseSyncRootWithDialog', () => chooseSyncRootWithDialog({ ctx }));
 
   // Backups
-  ipcPreloadMain.handle('deleteBackupsFromDevice', (_, props) => deleteBackupsFromDevice({ ctx, ...props }));
```
dajimenezriv-internxt: This is a fix needed because as said in the previous PR I resolved the conflicts wrongly and this line was duplicated.

packages/addon-cs/src/FileWatcher.cs
```diff
@@ -0,0 +1,154 @@
+using System.Runtime.CompilerServices;
```
dajimenezriv-internxt, line 1: Same funcionality as in C++.

packages/addon-cs/NativeMethods.txt
```diff
@@ -0,0 +1,14 @@
+CreateFile
```
dajimenezriv-internxt, line 1: Here we need to write all `NativeMethods` from C++ that we need in the C# code so they can be generated by `.NET` and used in the C# code.

packages/addon-cs/src/FileWatcher.cs
```diff
@@ -0,0 +1,154 @@
...
+    private const long EpochOffsetTicks = 116444736000000000L;
+
+    private readonly JSReference _callbackRef = new(onEvent);
+    private readonly JSSynchronizationContext _syncCtx = JSSynchronizationContext.Current
+            ?? throw new InvalidOperationException("No JSSynchronizationContext on current thread");
+    private readonly CancellationTokenSource _cts = new();
+
+    public void Start()
+    {
+        new Thread(Run) { IsBackground = true, Name = "FileWatcher" }.Start();
```
dajimenezriv-internxt, line 23: Now managing threads is much more simple.

PR #1392, opened 2026-05-22
Deduplicate create file

dajimenezriv-internxt:
## What

Explained in this ticket: https://inxt.atlassian.net/browse/PB-6411

When a user copies (not moves, copies) a folder with many files inside, the create events of the files are delayed in the watcher since the copy takes time. What can happen in this case is that the create events of those files are not ignored and are processed trying to create the files twice. If we detect that the create file promise is already happening, we ignore the new event.


packages/addon/include/virtual_drive/watcher/watch_path.h
```diff
@@ -69,9 +69,6 @@ inline void processEvent(FILE_NOTIFY_EXTENDED_INFORMATION* fni, const std::wstri
         case FILE_ACTION_MODIFIED:
             sendEvent(ctx, "update", path, fni);
             break;
-        case FILE_ACTION_RENAMED_OLD_NAME:
-            sendEvent(ctx, "rename_old", path, fni);
-            break;
```
dajimenezriv-internxt, line 74: We are not using the `rename_old` event, we use just the `rename_new`, so we don't need to spend resources sending the event to javascript.

packages/addon/include/virtual_drive/hydrate_file.h
```diff
@@ -11,7 +11,7 @@ inline void hydrate_file(const std::wstring& path)
     LARGE_INTEGER offset;
     offset.QuadPart = 0;
     LARGE_INTEGER length;
-    GetFileSizeEx(fileHandle.get(), &length);
+    length.QuadPart = -1;
```
dajimenezriv-internxt, line 14: This is the same, it was already used in dehydrate file.

AlexisMora (approved):

PR #1394, opened 2026-05-22
Stop download bottleneck on logout

dajimenezriv-internxt:
## What

1. Currently when the user logs out all requests and uploads are stopped. However, this is not happening for the downloads. Now, we create the download bottleneck inside the `AuthContext` and when the user logs out we stop also that bottleneck.
2. Add the context to the `getUsage` and `getLimit` so they can reuse the requests bottleneck.


src/apps/main/preload/ipc-main.ts
```diff
@@ -47,6 +46,7 @@ export function setupPreloadIpc() {
 }
 
 export function setupLoggedPreloadIpc({ ctx }: { ctx: AuthContext }) {
```
dajimenezriv-internxt, line 48: These are the handlers that are registered on login, so we need to unregister them also on logout. Otherwise, if we login twice it will throw an error because it will have 2 handlers registered to the same event.

AlexisMora (approved):

PR #1395, opened 2026-05-22
Add bottleneck to device requests

dajimenezriv-internxt:
## What

Devices was the only module that was not using yet the bottleneck for requests since it didn't have the `AuthContext` in their functions. Now, I've added the `AuthContext` to the functions, so they can use `ctx.client` in the requests. I've also fixed the ipc listeners since now they need to be set to listen when the user logs in and to remove when the user logs out.

With this change we can increase the requests per second from 2 to 3 (maybe even 4, but for now we can start just with 3), since now all requests are going to be using the same bottleneck.


src/apps/main/device/get-backups-from-device.ts
```diff
@@ -1,14 +1,21 @@
...
+type Props = {
+  ctx: AuthContext;
+  device: Device;
+  isCurrent?: boolean;
+};
+
+export async function getBackupsFromDevice({ ctx, device, isCurrent }: Props): Promise<Array<BackupInfo>> {
   try {
-    const { data: folder, error } = await driveServerWipModule.backup.fetchFolder({ folderUuid: device.uuid });
+    const { data: folder, error } = await driveServerWipModule.backup.fetchFolder({ ctx, context: { folderUuid: device.uuid } });
```
dajimenezriv-internxt, line 18: I know that using `ctx` and `context` is not the best, but it was the way they are already called. Maybe you can rename in the future the second `context` to `env` or something like that?

src/apps/main/remote-sync/handlers.ts
```diff
@@ -36,19 +34,15 @@ export async function updateAllRemoteSync() {
   );
 }
 
-export function setupRemoteSyncIpc() {
-  ipcMain.handle('get-item-by-folder-uuid', async (_, folderUuid): Promise<ItemBackup[]> => {
-    logger.debug({ msg: 'Getting items by folder uuid', folderUuid });
+export async function getItemsByFolderUuid({ ctx, folderUuid }: { ctx: AuthContext; folderUuid: string }): Promise<ItemBackup[]> {
```
dajimenezriv-internxt, line 37: Same funcionality but we extract it into a function.

src/apps/shared/HttpClient/client.ts
```diff
@@ -29,13 +29,12 @@ const middleware: Middleware = {
 
 export const client = createClient<paths>({
```
dajimenezriv-internxt, line 30: Now this client is just used for the `refresh` request, since it's the only request that can be done without being logged.

AlexisMora (approved):

PR #1397, opened 2026-05-27
Migrate fetch data to C#

dajimenezriv:
## What

Implemennt the fetch data callback in C#. This is the callback that happens when a user tries to hydrate a file using double click on that file. Basically we connect the callbacks in C# using `connectSyncRoot`, when a user tries to fetch a file we receive an event in C#. From C# we call the callback of javascript for download the file that he have registered in `connectSyncRoot`. From the javascript code we download the file and we call a callback in C# for every piece of bytes that we receive. The C# code starts hydrating the placeholder and sending the progress to windows.


packages/addon-cs/src/Addon.cs
```diff
@@ -14,48 +8,30 @@ public static class Addon
...
+    public static void UnwatchPath(string id)
     {
-        var id = (string)handle["id"];
         if (_watchers.TryRemove(id, out var w))
         {
             w.Stop();
         }
     }
 
-    public static Task HydrateFile(string path) => Task.Run(() =>
```
dajimenezriv-internxt, line 38: Just extract these functions into `Placeholder.cs`.

packages/addon-cs/src/Addon.cs
```diff
@@ -14,48 +8,30 @@ public static class Addon
 {
     private static readonly ConcurrentDictionary<string, FileWatcher> _watchers = new();
 
-    public static JSValue WatchPath(string rootPath, JSValue onEvent)
```
dajimenezriv-internxt, line 17: Instead of using an `object` for tracking active watchers we use an `guid string`.

packages/addon-cs/src/Placeholder.cs
```diff
@@ -0,0 +1,35 @@
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using Windows.Win32;
+using Windows.Win32.Foundation;
+using Windows.Win32.Storage.CloudFilters;
+using Windows.Win32.Storage.FileSystem;
+
+namespace Intx.Addon;
+
+internal static class Placeholder
```
dajimenezriv-internxt, line 10: The same code as before.

src/infra/inxt-js/file-uploader/upload-file.ts
```diff
@@ -45,6 +45,8 @@ export async function uploadFile({ ctx, readable, size, path, abortController, r
       progressCallback: (progress) => void progressCallback(progress),
     });
 
+    LocalSync.SyncState.addItem({ action: 'UPLOADING', path, progress: 1 });
```
dajimenezriv-internxt, line 48: After the download has been completed we put the progress to 100%, otherwise we don't receive the last progress notification from `inxt-js` and it leaves the file like at `96%` or whatever until is created on `drive-server-wip`.

packages/addon-cs/src/FileWatcher.cs
```diff
@@ -146,7 +146,7 @@ private void SendError(string message)
         {
             var evt = JSValue.CreateObject();
             evt["action"] = "error";
-            evt["path"] = message;
+            evt["message"] = message;
```
dajimenezriv-internxt, line 149: The variable is not really used in javascript (we just log the event when action is error), however, since in C++ we were using the same struct it was called path but it makes more sense to be called message since it contains the error message.

AlexisMora (approved):

PR #1399, opened 2026-05-28
[PB-6330] Retrieve and store the max file size by user

AlexisMora:
## What has been added/ changed
Now, uppon app startup or, when we recieve the `PLAN_UPDATED` event from the websocket, we will retrieve and store in the configStore the user `maxUploadFileSize`. If the API request fails, then we rely on the last stored value, if neither values are available, then we go "blind" and we dont pre-validate the users file size and rely on the backend.

### DIFF
--- a/.codegraph/.gitignore
+++ b/.codegraph/.gitignore
@@ -1,0 +1,16 @@
+# CodeGraph data files
+# These are local to each machine and should not be committed
+
+# Database
+*.db
+*.db-wal
+*.db-shm
+
+# Cache
+cache/
+
+# Logs
+*.log
+
+# Hook markers
+.dirty
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "clamav": "powershell release/clamAV.ps1",
     "prepare": "husky install",
     "swagger_dev": "openapi-typescript http://localhost:3004/api-json -o ./src/apps/shared/HttpClient/schema.ts",
-    "swagger_prod": "openapi-typescript https://gateway.internxt.com/drive/api-json -o ./src/apps/shared/HttpClient/schema.ts",
+    "swagger_prod": "openapi-typescript https://drive-server.internxt.com/api-json -o ./src/apps/shared/HttpClient/schema.ts",
     "schema": "sqlite3 %APPDATA%/internxt-drive/internxt_desktop.db .fullschema > src/infra/sqlite/schema.sql"
   },
   "build": {
--- a/src/apps/main/auth/handlers.ts
+++ b/src/apps/main/auth/handlers.ts
@@ -95,10 +95,10 @@ export async function emitUserLoggedIn(user: User) {
   eventBus.once('USER_LOGGED_OUT', () => {
     logger.debug({ tag: 'AUTH', msg: 'Received logout event' });
     clearSentryUserContext();
-    clearLoggedPreloadIpc();
     TokenScheduler.stop();
     BackupScheduler.stop();
     logout({ ctx });
+    clearLoggedPreloadIpc();
   });
 
   setupLoggedPreloadIpc({ ctx });
--- a/src/apps/main/auth/handlers.ts
+++ b/src/apps/main/auth/handlers.ts
@@ -102,7 +102,7 @@ export async function emitUserLoggedIn(user: User) {
   });
 
   setupLoggedPreloadIpc({ ctx });
-  await resolveUserFileSizeLimit({ ctx });
+  void resolveUserFileSizeLimit({ ctx });
   cleanAndStartRemoteNotifications({ ctx });
 
   const lastOnboardingShown = electronStore.get('lastOnboardingShown');
--- a/src/apps/main/auth/handlers.ts
+++ b/src/apps/main/auth/handlers.ts
@@ -8,6 +8,7 @@ import { AuthContext } from '@/apps/sync-engine/config';
 import { Marketing } from '@/backend/features';
 import { resetConfig } from '@/backend/features/auth/services/utils/reset-config';
 import { saveConfig } from '@/backend/features/auth/services/utils/save-config';
+import { resolveUserFileSizeLimit } from '@/backend/features/user/file-size-limit';
 import { BackupScheduler } from '../background-processes/backups/BackupScheduler/BackupScheduler';
 import { spawnSyncEngineWorkers } from '../background-processes/sync-engine';
 import electronStore from '../config';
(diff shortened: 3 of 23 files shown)


src/backend/features/remote-notifications/in/process-web-socket-event.ts
```diff
@@ -1,21 +1,30 @@
...
+    if (event.clientId === INTERNXT_CLIENT) {
+      logger.debug({
+        msg: 'Local notification received',
+        event: event.event,
+        clientId: event.clientId,
+        payload: event.payload,
+      });
+      return;
+    }
+    if (event.event === 'PLAN_UPDATED') {
```
AlexisMora, line 23: old behaviour is preserved, the only addition is that uppon plan updated webhook we fetch again for the user file size limit

package.json
```diff
@@ -43,7 +43,7 @@
     "clamav": "powershell release/clamAV.ps1",
     "prepare": "husky install",
     "swagger_dev": "openapi-typescript http://localhost:3004/api-json -o ./src/apps/shared/HttpClient/schema.ts",
-    "swagger_prod": "openapi-typescript https://gateway.internxt.com/drive/api-json -o ./src/apps/shared/HttpClient/schema.ts",
+    "swagger_prod": "openapi-typescript https://drive-server.internxt.com/api-json -o ./src/apps/shared/HttpClient/schema.ts",
```
AlexisMora, line 46: had to change this as the old url was throwing errors

.codegraph/.gitignore
```diff
@@ -0,0 +1,16 @@
+# CodeGraph data files
```
AlexisMora, line 1: [Codegraph](https://github.com/colbymchenry/codegraph) is a tool for claude/codex cli to discover and navigate far more easier

src/backend/features/user/file-size-limit/index.ts
```diff
@@ -0,0 +1,4 @@
+export { resolveUserFileSizeLimit } from './resolve-user-file-size-limits';
```
AlexisMora, line 1: I have added this index,ts so that we have a more organized module and when looking up directly, you already know which are the methods that are "exposed"

src/infra/drive-server-wip/services/files/get-user-file-size-limit.ts
```diff
@@ -0,0 +1,27 @@
...
+    promiseFn,
+    key,
+    loggerBody: { msg: 'Get user file size limit request' },
+  });
+
+  if (error) return { error };
+
+  return {
+    data: {
+      maxUploadFileSize: data.maxUploadFileSize,
```
dajimenezriv-internxt, line 24: What's the point of this and not doing just: `return await clientWrapper({ ... })`?

.codegraph/.gitignore
dajimenezriv-internxt: Why .db, or .db-wal and so? I mean, I know that they don't influence the code, but sometimes is the typical .gitignore autogenerated with a lot of files that you don't know what should be there or not

src/apps/main/auth/handlers.ts
```diff
@@ -101,7 +102,8 @@ export async function emitUserLoggedIn(user: User) {
   });
 
   setupLoggedPreloadIpc({ ctx });
-  cleanAndStartRemoteNotifications();
+  await resolveUserFileSizeLimit({ ctx });
```
dajimenezriv-internxt: With this you stop the whole main thread until the file size limit is fetched, if it retried or whatever it can be more time.

src/backend/features/user/file-size-limit/index.ts
dajimenezriv-internxt: When you do this and you try to go to the definition of the function from the code that is executing it you can have always to jump twice to go to the definition. But now you are the boss, you can ignore this comment.

.codegraph/.gitignore
AlexisMora: yes, it is the autogenerated .gititnore once you install the tool, could have been said it on the first message 😅

src/backend/features/user/file-size-limit/is-valid-max-upload-file-size.ts
```diff
@@ -0,0 +1,3 @@
+export function isValidMaxUploadFileSize(maxUploadFileSize: unknown): maxUploadFileSize is number {
```
dajimenezriv-internxt: `maxUploadFileSize` is already declared as `number` in the schema, why this?

src/backend/features/user/file-size-limit/resolve-user-file-size-limits.ts
```diff
@@ -0,0 +1,39 @@
+import configStore from '@/apps/main/config';
```
dajimenezriv-internxt: Use `electronStore` instead, it's typed. The point will be to stop using the default export.

src/infra/drive-server-wip/services/files/get-user-file-size-limit.ts
AlexisMora: because the result of `/files/limits` exposes one more property that we dont need, [you can check it here](https://drive-server.internxt.com/api#/File/FileController_getLimits:~:text=%7B%0A%20%20%22versioning%22%3A%20%7B%0A%20%20%20%20%22enabled%22%3A%20true%2C%0A%20%20%20%20%22maxFileSize%22%3A%200%2C%0A%20%20%20%20%22retentionDays%22%3A%200%2C%0A%20%20%20%20%22maxVersions%22%3A%200%0A%20%20%7D%2C%0A%20%20%22maxUploadFileSize%22%3A%200)
Thats why i thought i could just remove it returning it like this, so that we only return what we need

src/backend/features/user/file-size-limit/is-valid-max-upload-file-size.ts
AlexisMora: Trying last week we already got null, hence why i added this, to add more type safety

AlexisMora: but its typed as null already, we can get rid of this.

tests/vitest/setup.helper.test.ts
```diff
@@ -63,11 +63,10 @@ vi.mock(import('@internxt/drive-desktop-core/build/backend'), () => {
   };
 });
 
+// @ts-expect-error we cannot do importActual('electron') and use it inside the mock because sometimes it throws
+// Error: Electron failed to install correctly, please delete node_modules/electron and try installing again
```
dajimenezriv-internxt: We can revert this now, since the issue was something else.

dajimenezriv-internxt (approved):

PR #1400, opened 2026-05-28
[PB-6330] File size limits modal ui

AlexisMora:
## What has been Added

This pr introduces into the codebase the ui modal that will be used on the file size limits.
This is how it will be seen:
Captura de pantalla 2026-05-28 184220
Captura de pantalla 2026-05-28 184335


src/apps/renderer/pages/MaxFileSizeRejectionModal/upgrade-plan-list.tsx
```diff
@@ -0,0 +1,21 @@
+import { useI18n } from '../../localize/use-i18n';
+import { formatBytes } from '../../utils/format-bytes';
+import { upgradePlans } from './constants';
+import { getSuggestedUpgradePlan } from './service';
+
+export function UpgradePlanList({ suggestedPlan }: { suggestedPlan: ReturnType<typeof getSuggestedUpgradePlan> }) {
```
AlexisMora, line 6: Could have maintained this component in the same index.tsx but I think for readability can be separated

src/apps/renderer/pages/MaxFileSizeRejectionModal/format-bytes.ts
```diff
@@ -0,0 +1,9 @@
+export function formatBytes(bytes: number) {
```
AlexisMora, line 1: This is in the "utils" folder but it could be inside the modal as it is a really specific utility from this feature, later if its needed elsewhere we could move it again here.
I though as well on pass this value already formatted but i thought as well to have it here so we format it "on the frontend" the way we needed.
What do you think?

dajimenezriv-internxt: I also think that we should format in the frontend. About where to put it I don't know, I usually start by putting at first things closer to the feature because at the beginning they are used just in one place. In case they start being used in 2 places then I move them to a common place. Basically like don't trying to anticipate because maybe it's never used again.

vitest.config.renderer.mts
```diff
@@ -11,7 +11,7 @@ export default mergeConfig(
         reportsDirectory: './coverage/renderer',
       },
       environment: 'jsdom',
-      include: ['src/**/*.test.tsx'],
+      include: ['src/**/*.test.tsx', 'src/**/*.test.ts'],
```
dajimenezriv-internxt: The tests with `src/**/*.test.ts` are already running in unit testing, so now they will be running twice. I kept them in the unit testing since there they should be faster since you don't have to load the environment required for frontend tests.

packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@internxt/drive-desktop-core",
-  "version": "0.1.13",
+  "version": "0.1.14",
```
dajimenezriv-internxt, line 3: Just to point, until now when the change of core was just for windows I was not updating the version (same with the addon, I never update it), basically because you don't need to update the version in npm. Just when the core was going to be used in linux I was planning to update it.

dajimenezriv-internxt (dismissed):

dajimenezriv-internxt (approved):

PR #1401, opened 2026-05-29
feat: file size limit validation

AlexisMora:
## What has been Added/ Changed
This pull request introduces the upload size validation by the user max file size limit before uploading a file.
If the file exceeds the upload is skipped and a modal suggesting a plan upgrade with a cta is shown.

If the file exceeds the absolute 100gb cap then the upload is skipped as well, and this time we show the same modal but withouth upgrade and cta.

If the app does not have a stored max file size limit, local pre-validation is skipped and the backend remains the source of truth.

Multiple rejected files are grouped with a short debounce window, so folder uploads or batch operations show a single modal instead of one modal per file.


src/backend/features/user/file-size-limit/show-max-file-size-rejection-modal.ts
```diff
@@ -0,0 +1,39 @@
...
+    maximizable: false,
+    alwaysOnTop: true,
+    transparent: true,
+    webPreferences: {
+      preload: preloadPath,
+      nodeIntegration: true,
+    },
+  });
+  newWindow.on('closed', () => {
+    maxFileSizeRejectionModalWindow = undefined;
```
dajimenezriv-internxt, line 36: Doing this aren't we displaying a popup for each file that is above the size? Wasn't it just once during the whole app lifecycle?

src/backend/features/sync/actions/services/upload-file.ts
```diff
@@ -24,9 +26,17 @@ export async function uploadFile({ ctx, path }: Props) {
     return { contentsId: undefined, size, mtime };
   }
 
-  if (size > SyncModule.MAX_FILE_SIZE) {
-    ctx.logger.warn({ msg: 'File size is too big', path, size });
-    addSyncIssue({ error: 'FILE_SIZE_TOO_BIG', name: path });
+  const validation = validateUploadFileSize({ size, maxUploadFileSize: electronStore.get('maxUploadFileSizeInBytes') });
```
dajimenezriv-internxt, line 29: (Non blocking) Why not accessing the value directly in the `validateUploadFileSize` instead of sending the prop so you don't have to mock it in these tests?

src/backend/features/user/file-size-limit/max-file-size-rejection-state.ts
```diff
@@ -0,0 +1,95 @@
+import { showMaxFileSizeRejectionModal } from '@/backend/features/user/file-size-limit/show-max-file-size-rejection-modal';
```
dajimenezriv-internxt, line 1: I assume this is to show how many files during an upload have go over the limit right? Isn't it like too complex? Why not just the modal and that's all?

src/backend/features/user/file-size-limit/show-max-file-size-rejection-modal.ts
AlexisMora: no, we are displaying the popup only if we recieve the event and no modal is already shown

egalvis27 (dismissed):

larryrider (dismissed):

larryrider (approved):

PR #1412, opened 2026-06-12
Feat/pb 6323 public link sharing on desktop

AlexisMora:
## What has been changed/Added

This Pull Request adds public-link sharing from the File Explorer context menu. When the user right-clicks a single file or folder inside Internxt Drivet, a new `Copy Internxt share link option` is displayed in the primary context menu.

The context-menu entry is implemented as a native C++ [IExplorerCommand](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-iexplorercommand), which is the Windows API required for commands in the primary Windows 11 context menu. The command is hidden for items outside Internxt Drive, the root itself, empty selections, and multiple selections. Its is translated into English, Spanish, French, and German, and the Internxt icon is embedded directly in the [DLL](https://en.wikipedia.org/wiki/Dynamic-link_library).

The native command intentionally performs very little work. When clicked, it sends the selected filesystem path to Electron through a [Windows named pipe](https://learn.microsoft.com/en-us/windows/win32/ipc/named-pipes) and returns immediately. This prevents File Explorer from waiting.  I chosed this because both processes run on the same machine and only need to exchange a single filesystem path per click. Unlike an HTTP server, it does not require opening a TCP port, selecting and managing a port number, or exposing a network-accessible endpoint

Electron starts the named-pipe server once during application startup. It validates and decodes the UTF-16 path received from the extension, and resolves the local placeholder to its Drive UUID and item type.

Then, we create the public share link using the same general process as Drive Web. It generates a URL-safe share code, encrypts the mnemonic and code, creates the sharing entry through the Internxt SDK, decrypts the returned sharing code, and builds the final public URL. The resulting link is copied to the clipboard, and a native Windows notification reports whether the operation succeeded or failed (Its also has been added  English, Spanish, French, and German translations).

Registering a command in the Windows 11 primary context menu requires more than copying the C++ DLL into the application. Windows must know which [COM](https://learn.microsoft.com/en-us/windows/win32/com/component-object-model--com--portal) class the DLL exposes and where File Explorer can load it from. COM is the Windows mechanism used by File Explorer to discover and instantiate our `IExplorerCommand`. This registration is declared through a sparse  [MSIX](https://learn.microsoft.com/en-us/windows/msix/overview). A regular MSIX would package and install the complete application, but Internxt already uses NSIS as its installer. A sparse MSIX lets us keep NSIS for installing the Electron application while using MSIX only to provide the Windows package identity and context-menu registration metadata. The MSIX points to the native files placed by NSIS under resources/context-menu.

MSIX is Microsoft’s Windows application packaging format. Besides distributing application files, an MSIX manifest can declare an application’s identity and Windows integrations such as COM servers and File Explorer commands 

The DLL and minimal host executable are compiled with MSVC, Microsoft’s C++ compiler and build toolset. The build script discovers the installed MSVC toolset dynamically, following the existing native-project approach.

The `context-menu` files are installed in an isolated `resources/context-menu` directory. This prevents the sparse MSIX from changing permissions across the main Electron installation, The directory also contains a minimal host executable required by the MSIX package identity. It does not execute any sharing logic.

`npm run package` now produces a complete development installer. It builds the application and native extension, signs the required artifacts with a development certificate, bundles them, and generates the NSIS(Electron) installer. During installation, NSIS trusts the public development certificate when necessary and registers the MSIX(The one needed for the extension). Uninstalling the app removes both the extension registration and that specific certificate.

File Explorer may need to restart before showing the new command. The installer does not restart it automatically because doing so would close the user’s Explorer windows. The command appears after Explorer restarts naturally, after signing out, or after rebooting ( I have commented out the code just in case we end up needing it).

The release pipeline follows the equivalent production flow using DigiCert. It signs the native files, generates and signs the MSIX, builds and signs the final installer, and updates `latest.yml`.

The remaining validation is running the Windows release workflow and checking the DigiCert-signed production installer.


.github/workflows/windows-build-release.yml
```diff
@@ -52,8 +52,8 @@ jobs:
       - name: Install dependencies
         run: npm run init:ci
 
-      - name: Create package
-        run: npm run package
+      - name: Build application and context menu extension
+        run: npm run build:release
```
AlexisMora, line 56: Before this pull request, it was enought to sign the finall  `.exe`. Now with the new extension `context-menu`, since we are adding extra artifacts such as the DLL (The actual context-menu code) The host executable (The plaeholder required by MSIX) and the MSIX itself, which is the code in charge of the registration of the extension with windows The MSIX is especially strict because its manifest publisher must exactly match the certificate used to sign it. For that reason, production packaging can no longer create everything first and sign only at the end. `release/sign.ps1` now signs the DLL and host, reads the real DigiCert publisher from that signature, generates and signs the MSIX, builds the installer with those signed artifacts, and finally signs the installer

package.json
```diff
@@ -67,14 +76,43 @@
...
+        "from": "packages/context-menu/trust-development-certificate.ps1",
+        "to": "context-menu/trust-development-certificate.ps1"
+      },
+      {
+        "from": "packages/context-menu/unregister-context-menu-extension.ps1",
+        "to": "context-menu/unregister-context-menu-extension.ps1"
+      },
+      {
+        "from": "packages/context-menu/untrust-development-certificate.ps1",
+        "to": "context-menu/untrust-development-certificate.ps1"
```
AlexisMora, line 114: This entries are needed because the extension cannot live inside the `app.asar`. This is telling Electron builder to copy those files into the installed app under `resources/context-menu`. This way windows can register and load the extension from the fisical directory.

package.json
```diff
@@ -67,14 +76,43 @@
       "verifyUpdateCodeSignature": false,
       "artifactName": "Internxt-Setup-${version}.${ext}"
     },
+    "nsis": {
+      "include": "assets/context-menu-installer.nsh"
```
AlexisMora, line 80: This include lets us add an extra step after the installation, in this case, the context-menu install

src/apps/main/main.ts
```diff
@@ -118,6 +119,13 @@ async function start() {
     await app.whenReady();
     app.setAppUserModelId(INTERNXT_APP_ID);
 
+    const contextMenuPipe = startContextMenuPipe();
+    app.once('before-quit', () => {
+      logger.debug({ msg: 'App quitting, closing context-menu pipe' });
+      // ? What if we close the app abruptly.
+      contextMenuPipe.close();
```
AlexisMora, line 125: On graceful shutdown we close the server. On abrupt process exit windows releases the named-pipe automatically

packages/context-menu/register-context-menu-extension.ps1
```diff
@@ -0,0 +1,67 @@
...
+  -ForceUpdateFromAnyVersion
+
+Write-Host "Internxt context-menu package registered."
+
+# Explorer may need to restart before showing a newly registered extension.
+# We deliberately avoid restarting it during installation because that closes
+# the user's File Explorer windows and briefly refreshes the Windows desktop.
+#
+# Stop-Process -Name explorer -Force
+# Start-Process explorer.exe
```
AlexisMora, line 67: I commented this out just in case we need it in the future, but I dont like the idea of forcing the file explorer

packages/context-menu/InternxtContextMenu.vcxproj
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="utf-8"?>
```
AlexisMora, line 1: This file builds the actual native File Explorer extension. It compiles the C++ `IExplorerCommand` implementation into internxt_context_menu.dll, which is the DLL Explorer loads to decide whether the “Copy Internxt share link” option should appear, which title/icon it should show, and what should happen when the user clicks it.

larryrider (approved):

PR #1433, opened 2026-06-30
[BR-1995] Fix: backup file not updated after modified

AlexisMora:
## What has been Added/ Changed
In this pull request we modify the heuristic in which we determine if a local file has to be uploaded again to the backup because it has been modified or no, we now have into account the local file modification time and compare it with the remote one, if the local file has a modification time newer than the remote, then we add it into the "modified" list so that we upload it again.

## Why
Using only the file size comparison is not enough, small modifications on a backup file would not trigger since the file size would not change and not trigger the "backup" upload, hence why the modification time is a more precise way to check if a file was modified or not

### DIFF
--- a/src/apps/backups/diff/calculate-files-diff.ts
+++ b/src/apps/backups/diff/calculate-files-diff.ts
@@ -30,7 +30,7 @@ export function calculateFilesDiff({ local, remote }: TProps) {
       return;
     }
 
-    if (remoteFile.size !== local.stats.size) {
+    if (remoteFile.size !== local.stats.size || isLocalFileNewer({ local, remote: remoteFile })) {
       modified.push({ local, remote: remoteFile });
       return;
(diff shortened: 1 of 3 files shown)


src/apps/backups/diff/calculate-files-diff.ts
```diff
@@ -54,3 +54,12 @@ export function calculateFilesDiff({ local, remote }: TProps) {
     total,
   };
 }
+
+function isLocalFileNewer({ local, remote }: { local: StatItem; remote: ExtendedDriveFile }) {
+  const remoteTime = remote.modificationTime || remote.updatedAt;
```
larryrider: preffer ?? over ||

larryrider (dismissed):

larryrider (approved):

PR #1434, opened 2026-06-30
chore: husky pre push hook

AlexisMora:
## What is Changed / Added
Added a warning when using `||` over the `??` operator
Added a pre-push hook like we have in other company projects so that we dont have to wait for the errors on the pr pipeline
Added the `PULL_REQUEST_TEMPLATE` markdown file so that we have a common template amongts the linu/windows prs


.husky/pre-push
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
```
larryrider: if the husky dependency is updated to the latest version, then this two lines can be removed

like here: https://github.com/internxt/cli/blob/main/.husky/pre-commit

larryrider (approved):

PR #1448, opened 2026-07-16
Fix/improve load in memory paths

AlexisMora:
## What is Changed / Added
This PR improves the startup placeholder refresh path, especially the local placeholder scan performed by `loadInMemoryPaths`.

The file placeholder loading was also extracted into its own helper and changed to workers. This avoids creating one promise per file upfront when a folder contains many files. `statReaddir` now follows the same idea for filesystem stats, so a large directory does not launch one stat operation per entry at once

The recursive folder walk has been replaced with an iterative depth-first traversal helper under `backend/features/virtual-drive/tree-traversal`. I have made this function reusable so that we can use it in both `loadInMemoryPaths` and `traverse` keeping the traversal mechanics in one place while preserving the existing behavior.

---

## Why
 Users with very large drives have reported startup freezes and long periods where the app appears unresponsive. `loadInMemoryPaths` is part of that startup path, and it has to inspect a large local placeholder tree before comparing local state with remote state.

## Summary by CodeRabbit

* **Improvements**
  * Improved remote file synchronization for nested folders and large directory structures.
  * Added bounded processing to reduce resource usage during file and folder scans.
  * Improved handling of missing or unreadable files so processing can continue when individual entries fail.
  * Added support for stopping long-running folder scans when cancellation is requested.
  * Improved reliability when processing deeply nested folder hierarchies.


larryrider (dismissed):

AlexisMora:
Closing this PR because the `loadInMemoryPaths` traversal/concurrency changes were already merged into main through #1452

PR #1452, opened 2026-07-22
Fix: Optimize backups and virtual-drive traversal performance

AlexisMora:
## What is Changed / Added

In this pull request, I refactored `loadInMemoryPaths` to use the shared traversal helper. Its behavior is preserved: folders without placeholder metadata are skipped, files keep the resolved parent folder UUID, and the returned `fileExplorer` shape stays the same.

File placeholder loading and `statReaddir` were changed to use bounded worker loops. This way we can avoid creating multiple async tasks upfront when iterating through large folders, we are only generating a promise for the item we are processing being this method less resource consuming

The backup remote-tree query now asks SQLite directly for `EXISTS` files and folders instead of loading all statuses and filtering them in memory. To support this, `getByWorkspaceId` for files and folders now accepts optional `fileStatus` / `folderStatus` filters while preserving the existing behavior when no filter is passed.

Tests were added or updated to cover the important invariants: depth-first traversal order, subtree skipping, deep trees, parent UUID propagation, bounded worker behavior, SQLite status filtering, and backup remote-tree output order.

  ---

## Why

Users with large drives or large backup folders have reported startup freezes and long periods where the app appears unresponsive.

The previous implementation had a few scalability problems: recursive traversal, repeated full-array filtering per folder, and large promise batches when folders contained many items. Those patterns become expensive with tens or hundreds of thousands of files.

This PR keeps the current behavior but makes the work more predictable and bounded. The goal is to reduce unnecessary resource consumtion and avoid recursive stack growth, and make the traversal logic shared instead of duplicated across virtual-drive and backups.

## Summary by CodeRabbit

* **Performance**
  * Improved virtual-drive tree building using depth-first traversal and grouped indexing by parent.
  * Added bounded concurrency controls across directory statting and placeholder loading to prevent resource spikes.
* **Reliability**
  * Tree traversal now supports cancellation via abort signals.
  * Metadata/stat failures no longer halt processing; errors are reported while continuing.
* **Data Accuracy**
  * Workspace file and folder retrieval can now filter by status (e.g., `EXISTS`) so only relevant items are included.
* **Tests**
  * Expanded coverage for concurrency limits, traversal behavior, and in-memory path loading.


src/core/utils/concurrency.ts
```diff
@@ -0,0 +1,3 @@
+export function getWorkerCount({ concurrency, itemCount }: { concurrency: number; itemCount: number }) {
```
AlexisMora, line 1: This function comes originally from getFileProcessingWorkerCount in `src/context/virtual-drive/items/application/Traverser.ts` and since we are using it in multiple places i decided to add it in a common place. Also i decided to reduce complexity since the number type is already enforced by param hence why you dont see the `Number.isFinite`

src/context/virtual-drive/items/application/Traverser.ts
```diff
@@ -22,60 +28,18 @@ type Props = {
...
-    } else {
-      foldersByParentUuid.set(folder.parentUuid, [folder]);
-    }
-  }
-
-  return { filesByParentUuid, foldersByParentUuid };
+  const { filesByParentUuid, foldersByParentUuid } = indexDriveItemsByParentUuid(database);
+  const root: StackItem = { folder: currentFolder, requiresPlaceholderUpdate: false };
+
+  await traverseDepthFirst<StackItem>({
```
AlexisMora, line 36: What i have done really in this file is on one side, extract the traverse into its own function so the process can be used where needed as well as the pre-index, since the sync traverser and the backup traverser are essentially the same

larryrider (dismissed):

larryrider:
check sonarcloud and coderabbit before merge

larryrider (approved):

PR #1461, opened 2026-07-30
feat: use token refresh logic from @internxt/lib

AlexisMora:
## What is Changed / Added
in this pr I refactor the tokenScheduler so that it only handles the scheduling mechanism, the logic behind "when" will come from @internxt/lib dependency.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session validation to reliably detect expired or invalid authentication tokens.
  * Users are now signed out cleanly when token validation fails or an authentication error occurs.
  * Improved automatic token renewal timing to help prevent unexpected session interruptions, including extended expiration periods.

* **Refactor**
  * Centralized token validation and expiration handling for more consistent authentication behavior.
  * Improved handling and reporting of authentication token errors.


src/apps/main/auth/handlers.ts
```diff
@@ -47,9 +48,10 @@ export function checkIfUserIsLoggedIn() {
     return;
   }
 
-  const msToExpire = TokenScheduler.getMillisecondsToExpire();
-  if (msToExpire === null || msToExpire <= 0) {
-    logger.debug({ tag: 'AUTH', msg: 'User token is expired' });
+  const { data: tokenStatus, error } = validateTokenAndCheckExpiration();
+
+  if (tokenStatus === 'INVALID' || tokenStatus === 'EXPIRED' || error) {
```
larryrider: Hey! here it should be used the TokenStatus.INVALID and TokenStatus.EXPIRED types

src/apps/main/token-scheduler/TokenScheduler.ts
```diff
@@ -1,58 +1,31 @@
...
-    const msToRenew = this.getMillisecondsToRenew();
-    if (msToRenew === null) return;
+    logger.debug({
+      tag: 'AUTH',
+      msg: 'Token renew date',
+      expiresAt: new Date(expiresAt),
+      renewAt: new Date(renewAt),
+    });
 
     this.timeout = setTimeout(async () => {
```
larryrider, line 31: Hey I just found that `setTimeout` in Node caps out at ~24.8 days (2^31-1 ms), so anything bigger just fires almost immediately instead of waiting 🤔 . As now `msToRenew` comes straight from the backend's exp/iat, nothing stops it from exceeding that if a token ends up with a longer TTL, and we would end up refreshing it repeatedly.

A possible fix could be to use `Math.min(msToRenew, MAX_SAFE_TIMEOUT)`, so we dont rely on the backend keeping tokens under ~24 days.

larryrider (changes requested):

src/apps/main/token-scheduler/TokenScheduler.ts
AlexisMora: Nice catch, will make a fix

src/apps/main/token-scheduler/TokenScheduler.ts
```diff
@@ -1,77 +1,49 @@
...
+        this.schedule();
+        return;
+      }
+
       const { data } = await driveServerWip.auth.refresh();
 
       if (data) {
         updateCredentials({ newToken: data.newToken });
         this.schedule();
       }
```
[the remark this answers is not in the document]
AlexisMora: This issue is already handled on `src\infra\drive-server-wip\in\client-wrapper.service.ts`
the [clientWrapper ](https://github.com/internxt/drive-desktop/blob/main/src/infra/drive-server-wip/in/client-wrapper.service.ts#L36)has an [errorWrapper](https://github.com/internxt/drive-desktop/blob/main/src/infra/drive-server-wip/in/error-wrapper.ts#L13) that checks if its an error server via [isServerError](https://github.com/internxt/drive-desktop/blob/9fb3aeac41b311ab8b068b839bf7f3198c3a4f45/src/infra/drive-server-wip/in/helpers/error-helpers.ts#L66)
```TypeScript
export function isServerError({ response: { status } }: { response: Response }): boolean {
  if (status === 429) return true;
  return status >= 500 && status < 600;
}
```
and if so, in the [clientWrapper L70](https://github.com/internxt/drive-desktop/blob/9fb3aeac41b311ab8b068b839bf7f3198c3a4f45/src/infra/drive-server-wip/in/client-wrapper.service.ts#L70) we retry the query again

larryrider (approved):

PR #1481, opened 2026-08-25
[_]: Fix: Refactor addon watch path to send events in batches to reduce workload on electron

AlexisMora:
## What is Changed / Added
This PR changes the watcher strategy of processing events so that a large filesystem burst of events from the addon does not create an unmanageable amount of work on Electron main process.

 ### Native watcher: batch filesystem events before crossing the N-API boundary
 
[ReadDirectoryChangesExW](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-readdirectorychangesexw) can return many filesystem notifications in a short period. Originally, the native addon sent one N-API thread-safe callback per event to electron. Now we collect the events in an array and once we reach 500 items, we send them to electron, we keep repeating this process untill all of the events have been sent.

For operations of less than 500 events, we also sent them once we have finished to collect all of the notifications from `ReadDirectoryChangesExW`.

## Electron side: replace per-item timers with a shared scheduler
The previous strategy on the electron side was to have a separate `setTimeout` for each `internalId`.  A large copy could therefore create hundreds of thousands of timers. When their debounce period elapsed together, each timer started Sync work without a global concurrency limit.

The scheduler is scoped to a `SyncContext`. In practice, most users will only have one active context, but a personal Drive and a Business workspace can exist at the same time. Each context needs its own pending events, timer, and processing limit so that activity in one workspace cannot delay or reorder work in another one.

When an event arrives, the scheduler stores the latest event for that item in a `Map`, using `internalId` as the key. The `Map` is the source of truth: if Windows reports several changes for the same file while it is being copied, we only need to process the final known state once the file has been quiet for the debounce period. Events for different files are kept independently and are never replaced by each other.

The scheduler also keeps a small heap entry for each pending deadline. It only answers one question efficiently: “which item needs to be processed next?” A [min-heap](https://en.wikipedia.org/wiki/Heap_(data_structure)) is useful here because it can always expose the earliest deadline without scanning or sorting every pending item.
This lets the scheduler maintain one shared timer for the next deadline, instead of creating one `setTimeout` per file.

When an event for an item is replaced, the old heap entry is left in place and becomes stale. It is discarded when it reaches the top of the heap and no longer matches the latest version stored in the `Map`. This is intentional: removing an arbitrary item from a heap on every update would make the code and update path more complex, while lazy cleanup keeps the `Map` as the clear source of truth.

Once events are due, the scheduler moves them to an `async.queue` in small batches. The queue gives us a single, explicit concurrency limit for Sync work. Without it, a large copy could start thousands of asynchronous `processEvent` calls at the same time, competing for the Electron main process, filesystem, database, and network resources. If the queue is full, due events remain pending, so they are delayed and processed when its due time.

The scheduler also yields between promotion batches with `setImmediate`. This is important because the Electron main process must continue handling tray clicks, window actions, IPC, and other UI work while Sync is draining a large copy

Finally, an item that is already being processed is marked as active. If a newer event for that same item arrives before the previous one finishes, the newer event waits. This prevents two versions of the same file from being processed concurrently while still allowing unrelated files to use the configured concurrency.

### Important limitation: Windows watcher overflow
If `ReadDirectoryChangesExW` reports `bytesReturned === 0`, Windows has already lost the exact sequence of notifications. The addon reports: `ReadDirectoryChangesExW buffer overflow; reconciliation required`

**TL;DR:** This PR prevents large Windows filesystem bursts from overwhelming Electron. The native addon now sends watcher events to JavaScript in batches, while a per-workspace scheduler keeps only the latest pending event for each item, uses one shared debounce timer, and processes Sync work with bounded concurrency. Large copies may still take time to upload, but the Electron UI and tray should remain responsive instead of freezing

---

## Why
A QA reproduction showed that copying a 250k+ item folder into the Sync root could make the laptop appear frozen and prevent the tray widget from opening.


src/node-win/watcher/watcher.ts
```diff
@@ -1,16 +1,26 @@
...
+     * The addon has already grouped native notifications.
+     * Keep this callback deliberately small: the scheduler owns debounce/coalescing and bounded processing
+     * while this boundary avoids one N-API callback per item
+     */
+    onEvents: (events) => events.forEach((event) => onEvent({ ctx, event })),
   });
 
   return {
-    unsubscribe: () => Addon.unwatchPath({ handle }),
+    unsubscribe: () => {
```
victor-ferro, line 25: Something I'm not sure about though: getWatcherEventScheduler builds a new scheduler whenever it doesn't find one in the map. Since unwatchPath only sets a flag and batches already queued in the TSFN still get delivered, an event arriving right after disposeOnEventScheduler would create a fresh scheduler rather than hitting the if (state.disposed) return guard in addEvent — and nothing would dispose that one afterwards.

Probably a narrow window, just wanted to check whether it had come up. Thoughts?

src/node-win/watcher/on-event/event-scheduler/watcher-event-scheduler.ts
```diff
@@ -0,0 +1,156 @@
...
+  void state.workQueue.push(() => processQueuedEvent(state, pendingEvent));
+}
+
+/** async.queue invokes this work only when its global concurrency has capacity. */
+async function processQueuedEvent(state: SchedulerState, pendingEvent: PendingEvent) {
+  if (state.disposed) return;
+
+  const internalId = pendingEvent.event.internalId;
+  const latest = state.pendingEvents.get(internalId);
+  if (!latest || latest.version !== pendingEvent.version || pendingEvent.state !== 'queued') return;
```
victor-ferro: the early return here skips the finally { finishProcessing } below, and finishProcessing is the only thing that calls scheduleNextDebounceTimer again. Together with the delay <= 0 branch up top (which clears the timer and bails when the queue is at capacity), I think the queue can drain and free up capacity with nothing left to re-arm — so due events would sit in the map until the next filesystem event arrives. The version mismatch here looks fairly common during a bulk copy, and the tail of a copy is exactly where no more events come in.

If that holds up, I think this covers it:

```suggestion
  if (!latest || latest.version !== pendingEvent.version || pendingEvent.state !== 'queued') {
    scheduleNextDebounceTimer(state);
    return;
  }
```

AlexisMora: Thats a nice catch, im adding it into the pr

src/node-win/watcher/watcher.ts
AlexisMora: Even though thats a really thin corner-case, its still worth making a solution for it now that you got it

victor-ferro (approved):
Top!!!

PR #1484, opened 2026-08-27
[_] Fix: Add observedatMs property to addon event to preserve logic in onChange

AlexisMora:
## What is Changed / Added
In this pull request I move the reference timestamp used by the Windows watcher `update` heuristic from Electron to the native addon.

`onChange` originally used `ctimeMs` and `mtimeMs` from the Windows event to decide whether an `update` notification appears recent:

 ```ts
  now - event.ctimeMs <= 5000
  now - event.mtimeMs <= 5000
```
And `now` was calculated when `onChange` eventually processed the event in Electron. 
Now, with the current changes on the way that we process the events, it  can be substantially later than the moment the native watcher received the notification, especially when Electron is busy or when the watcher scheduler deliberately delays work to keep the application responsive.

The addon now captures `observedAtMs` immediately after `ReadDirectoryChangesExW` returns a result. The value is added to every `WatcherEvent` from that Windows result, transferred through `N-AP`I, retained by the watcher scheduler on Electron side, and passed to `onChange`.

`onChange` now evaluates the existing heuristic against `observedAtMs`:
```ts
  observedAtMs - event.ctimeMs <= 5000
  observedAtMs - event.mtimeMs <= 5000
```
Why? Since the way we process the events now is scheduled and consumed in a queue, tying the "now" on the exact moment the event is processed may cause a change in the original, intended logic. By attaching the "observedAtMs" from the addon, we can have a more precise 

---

## Why
The watcher scheduler may intentionally delay an event before it reaches `onChange`. This delay is necessary to avoid overwhelming Electron during large filesystem bursts, but it must not change the meaning of the existing timestamp heuristic.

  For example:
```
12:00:00  Windows reports a file update.
12:00:00  The addon receives the notification.
12:00:30  Electron finally processes onChange.
```
With `Date.now()` inside `onChange`, the event appears to be thirty seconds old.
The current heuristic can then classify it differently only because the app delayed its own work.

Capturing `observedAtMs` in the addon establishes a more accurate boundary:

```
How recent was the filesystem timestamp when the native watcher observed the Windows notification?
```
It excludes time spent waiting in the N-API queue, Electron event loop, debounce, and bounded Sync processing.

This does not make the five-second heuristic semantically reliable by itself. A Windows Changed notification does not identify the process or operation that caused it. A recent timestamp does not prove that a user edited the file, and an old timestamp does not prove that the user did not. The code contains a follow-up TODO to replace this timestamp-based inference with a more deterministic operation-aware and reconciliation-based approach


victor-ferro:
@AlexisMora 
Small robustness note, non-blocking 🙂

`observedAtMs` arrives from the addon unvalidated. If it ever came missing, `undefined - ctimeMs` is `NaN` and `NaN <= 5000` is `false` — `isChanged` and `isModified` would be permanently false and `onChange` would silently do nothing.

Nothing would catch it today, since `toMatchObject` ignores fields it doesn't list. One extra property in `src/node-win/watcher/tests/watcher-on-add.test.ts:48` (the only test that inspects a real event from the compiled addon) would cover it:

```ts
call(onEventSpy).toMatchObject({
  event: { action: 'create', type: 'file', size: 0, observedAtMs: expect.any(Number) },
});

victor-ferro (approved):

PR #1486, opened 2026-09-01
fix: remove orphan sync roots that Windows no longer resolves (BR-2314)

victor-ferro:
## What is Changed / Added

`unregisterVirtualDrives` gets a second pass that reads `SyncRootManager` straight from the registry and removes the entries Windows will not resolve. New `src/infra/windows-registry` module, no new dependencies.

## Why

Registering a virtual drive writes **two linked halves**:

```
HKLM\...\Explorer\SyncRootManager\{providerId}
    NamespaceCLSID, DisplayNameResource, IconResource, Flags, RecycleBinUrl
    UserSyncRoots[<user SID>] = <root path>

HKCU\...\Explorer\Desktop\NameSpace\{NamespaceCLSID}     ← what Explorer draws
HKCU\Software\Classes\CLSID\{NamespaceCLSID}
```

If the process dies mid-registration, the shell half is complete but the provider half is not — it stops at `NamespaceCLSID` + `DisplayNameResource`, with no `UserSyncRoots`. The result is unreachable from the Cloud Files API in **both** directions:

| | result |
|---|---|
| `GetCurrentSyncRoots()` | does not return it — no `Path()`, no `Version()` to build the object from |
| `Unregister(id)` | throws `0x80070490 ERROR_NOT_FOUND` on a key that is plainly there |
| Explorer | renders it, pinned, with `SFGAO_CANDELETE` **off** — hence no Delete in the context menu |

So `unregisterVirtualDrives` could never see it, and the user could never delete it. Deleting the folder from disk does not help either: the three registry keys survive and the entry stays, now pointing at nothing.

Only direct registry deletion clears it, which is what this PR does — and only for entries that pass four filters:

```ts
registration.displayName.toLowerCase().includes('internxt') &&
!registration.hasUserSyncRoots &&
!currentProviderIds.includes(registration.id) &&
!registeredIds.includes(registration.id),
```

`hasUserSyncRoots` separates the two reasons the API may skip an entry: **broken** (no `UserSyncRoots` at all) vs **belonging to another Windows user** (`UserSyncRoots` present, under a different SID). Without it, a shared machine with two Windows accounts using Internxt would have one account delete the other's registration.

`currentProviderIds` guards a real race: `spawnDrive` does not await `spawnSyncEngineWorker`, so our own registration may still be in flight when this runs — indistinguishable from a phantom at that moment.

No filesystem and no database access: the v2.5.1 rule about never deleting the folder is preserved.

## Verification

Reproduced deterministically on a real machine and driven through the real app, three scenarios:

| scenario | result |
|---|---|
| Phantom with its folder present | detected and removed |
| Phantom whose folder was deleted by hand | detected and removed |
| No phantom | `orphans: []`, nothing touched |

Plus a real account switch: the active drive was unregistered through the normal API pass and never appeared as an orphan.

```
Orphan sync root registrations  orphans: [ { id: 'syncRootID', displayName: 'Internxt',
  namespaceClsid: '{995BF3C7-...}', hasUserSyncRoots: false } ]
Removing orphan sync root registration
```

Files were verified untouched across the cleanup, including a file that only existed locally.

12 unit tests on `unregisterVirtualDrives` (5 of them pinning the filter boundaries) and 2 `.infra` tests reading the machine's real registry.

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically removes stale Windows sync-root registrations no longer associated with active folders.
  * Preserves valid registrations belonging to the current provider, other providers, or other users.
  * Opening a second app instance now correctly displays the existing widget when available.

* **Reliability**
  * Improved Windows registry access and cleanup so individual failures do not interrupt remaining operations.
  * More accurately identifies registrations using their associated folder paths.

* **Tests**
  * Added coverage for detecting, preserving, and removing Windows sync-root registrations.

### DIFF
--- a/src/apps/main/background-processes/sync-engine/services/unregister-virtual-drives.test.ts
+++ b/src/apps/main/background-processes/sync-engine/services/unregister-virtual-drives.test.ts
@@ -1,15 +1,19 @@
-import { call, mockProps, partialSpyOn } from 'tests/vitest/utils.helper.test';
+import { call, calls, mockProps, partialSpyOn } from 'tests/vitest/utils.helper.test';
+import { WindowsRegistry } from '@/infra/windows-registry/windows-registry.module';
 import { Addon } from '@/node-win/addon-wrapper';
 import { unregisterVirtualDrives } from './unregister-virtual-drives';
 
 describe('unregister-virtual-drives', () => {
   const getRegisteredSyncRootsMock = partialSpyOn(Addon, 'getRegisteredSyncRoots');
   const unregisterSyncRootMock = partialSpyOn(Addon, 'unregisterSyncRoot');
+  const getSyncRootRegistrationsMock = partialSpyOn(WindowsRegistry, 'getSyncRootRegistrations');
+  const removeSyncRootRegistrationMock = partialSpyOn(WindowsRegistry, 'removeSyncRootRegistration');
 
   let props: Parameters<typeof unregisterVirtualDrives>[0];
 
   beforeEach(() => {
     getRegisteredSyncRootsMock.mockReturnValue([{ id: '{PROVIDER_ID}', displayName: 'Internxt' }]);
+    getSyncRootRegistrationsMock.mockResolvedValue([]);
     props = mockProps<typeof unregisterVirtualDrives>({ currentProviderIds: ['{PROVIDER_ID}'] });
   });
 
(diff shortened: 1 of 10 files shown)


victor-ferro:
`686a41ab` — second, unrelated fix from the same ticket.

The widget is `skipTaskbar: true` and frameless, so the tray icon is the only way to open the app. On Windows 11 a reinstall drops that icon into the hidden overflow, leaving the desktop shortcut as the only entry point a user can find — and it did nothing, because the running instance only parsed a deeplink and returned:

```diff
 app.on('second-instance', (event, argv) => {
   processDeeplink({ argv });
+
+  if (getWidget()) showFrontend();
 });
```

From the ticket's log, twice, 45 seconds apart:

```
09:14:44.274  Registering protocol
09:14:44.277  Unknown deeplink, argv: [ 'C:\...\internxt-drive\Internxt.exe' ]
```

No `Starting app` line — `!gotTheLock` exits before it. `showFrontend()` is the same function the tray click already uses; the guard covers the window before `createWidget()` has run.

AlexisMora (dismissed):

victor-ferro:
`82010f28` — the `displayName` filter had a hole.

`Register` does not write the registration atomically. The order matters:

```
1. HKCU\...\CLSID\{clsid}\Instance\InitPropertyBag  TargetFolderPath
2. HKLM\...\SyncRootManager\{id}                    NamespaceCLSID
3. HKCU\...\Desktop\NameSpace\{clsid}               <- Explorer draws it from here
4. HKLM\...\SyncRootManager\{id}                    DisplayNameResource, IconResource, Flags, RecycleBinUrl
5. HKLM\...\SyncRootManager\{id}\UserSyncRoots\<SID>
```

The entry becomes visible at step 3 — one step before `DisplayNameResource` exists. So a phantom created in that window is drawn in Explorer with **no display name at all**, and the previous filter skipped it. The one thing it is guaranteed to have is `TargetFolderPath`, written at step 1.

```diff
-registration.displayName.toLowerCase().includes('internxt')
+registration.displayName.toLowerCase().includes('internxt') ||
+  registration.targetFolderPath.toLowerCase().includes('internxt')
```

This is coverage by construction, not by luck: no phantom can be visible to the user without having passed step 1.

The other three filters are untouched, so this only widens the candidate set, never the deletion criteria. A registration still has to have no `UserSyncRoots`, be absent from `GetCurrentSyncRoots()`, and not be a current provider id. Anything Windows can still resolve — ours or another provider's — is out on the second and third filter regardless of what its path is called.

AlexisMora (approved):

PR #1493, opened 2026-09-07
fix: stop propagating deletions from folders we could not sync (BR-2245)

victor-ferro:
## Problem

A folder whose name cannot be reconciled enters a rename loop that never converges.
`updateFolderPlaceholder` returns `false` and the traversal abandons the entire subtree,
silently, so a large share of the account never makes it to disk. When that content later
disappears from disk, the watcher reads it as a user deletion and replicates it to the
server 1:1, with no checks at all — thousands of deletion events within seconds, and the
same number of items sent to the cloud trash.

## What this PR does

1. **Validate windows names by what windows can address.** The rule is now whether windows can
   *reach* the name, not whether it can create it. We create placeholders through `\\?\` paths,
   which skip win32 name parsing, so a trailing space or dot is stored verbatim and explorer
   then cannot find it — that is BR-1796, a folder the user can see and cannot delete. Adds
   trailing dots and control characters; stops rejecting a leading space, which is addressable.

2. **Track folders we could not reconcile.** When a folder's reconciliation fails we store its
   local path, which is the one the watcher reports and can differ from the remote one — that
   difference is usually why it failed in the first place.

3. **Convergence counter.** Today `checkIfMoved` treats the reconciliation as done if `rename`
   does not throw. That is not true: it can return without an error and leave the folder where
   it was. When that happens `needsToBeMoved` asks for the same move again on the next pass,
   and again after that — over a thousand times on a single folder in the logs we analysed,
   without a single error logged.

   The only reliable signal that a move worked is that it **stops being requested**: if the
   rename at 13:30 had taken effect, there would be nothing left to move at 13:41.

   ```
   13:30:56  Moving placeholder  ~/<origin>  ->  ~/<destination>
   13:41:08  Moving placeholder  ~/<origin>  ->  ~/<destination>
   13:51:31  Moving placeholder  ~/<origin>  ->  ~/<destination>
   ```

   So we count consecutive attempts of the same **origin to destination pair**. The pair, rather
   than the folder, is what separates a retry from new work: if the move succeeded, the origin
   on the next pass is already a different path, because the folder has a different name. A
   user renaming a folder five times from the web produces five different pairs and resets the
   counter each time; a move that never takes effect repeats the same pair and adds up.

   After 10 attempts we stop trying and mark the folder, so the guard in point 4 covers its
   subtree. The limit is deliberately generous, because a move can also fail for transient
   reasons — a file held open inside the folder, for instance — and those resolve on their own
   and reset the count.

4. **Guard on deletion propagation.** A local deletion inside a folder we could not reconcile
   is not sent to the server.

## What it fixes

- The data loss in the main incident: the folder was still looping when it happened, so the
  counter would have marked it and the guard would have stopped every propagated deletion.
- A `CfCreatePlaceholders` loop caused by a trailing dot, which ran for over a thousand
  iterations and was never detected.
- Names containing a tab, which threw inside the addon instead of being reported.
- Names with a leading space, whose subtrees were being pruned for no reason.

## What it does NOT fix

- **The contents of folders with invalid names still do not sync.** They are detected and
  reported, but the subtree is still pruned. The fix is name sanitising, which will go in a
  branch on top of this one: create the placeholder with the name win32 can actually reach
  (`Report.` -> `Report`) so the subtree syncs. The link to
  the cloud does not break, because reconciliation goes by uuid, but it does require
  `needs-to-be-moved` and `on-unlink` to compare the sanitised name.
- **A mass deletion with no broken folder above it.** That needs a volume brake, already
  measured and pending: normal use never exceeded 6 propagated deletions per minute, while the
  four incidents were 15, 33, 53 and 58. It was left out of this PR because notifying the user
  relies on an Electron notification carrying a `TODO: Notification is not working` since
  v2.5.3, and we need to settle where the warning surfaces first.
- Backups, WebSocket and the context menu from the ticket.

## Pending outside this repo

The `INVALID_WINDOWS_NAME` strings in `packages/core` (en/es/fr) say "start/end with spaces";
they should now say "end with a space or a dot". Needs a PR against the submodule.

## Tests

699 unit tests and 33 infra tests. The infra one creates the folder and checks against real
windows whether it is addressable, so the rule set cannot drift back towards the documentation.
`watcher-on-unlink-unreconciled` reproduces the whole chain with the native watcher: a rename
that genuinely fails, the folder being marked, a real deletion on disk, and the assertion that
nothing reaches the server.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented deletions from syncing to the server when items are inside folders that failed reconciliation.
  - Restored deletion syncing after folders are reconciled.
  - Added safeguards to stop repeatedly retrying moves that cannot converge.
  - Improved handling of unreconciled folders during sync and logout.
  - Updated Windows filename validation to better match actual Windows behavior, including leading spaces and Unicode whitespace.

- **Tests**
  - Expanded coverage for unreconciled folders, move retries, deletion handling, and Windows-compatible names.

- **Documentation**
  - Added architecture decision records and guidance for documenting future architectural decisions.

### DIFF
--- a/src/apps/main/auth/logout.ts
+++ b/src/apps/main/auth/logout.ts
@@ -3,6 +3,7 @@ import { AuthContext } from '@/apps/sync-engine/config';
 import { LocalSync } from '@/backend/features';
 import { resetConfig } from '@/backend/features/auth/services/utils/reset-config';
 import { saveConfig } from '@/backend/features/auth/services/utils/save-config';
+import { onLogout as clearUnreconciledFolders } from '@/backend/features/remote-sync/unreconciled-folders';
 import { DriveServerWipModule } from '@/infra/drive-server-wip/drive-server-wip.module';
 import { clearAntivirus } from '../antivirus/utils/initializeAntivirus';
 import { clearIssues } from '../background-processes/issues';
@@ -35,6 +36,7 @@ export function logout({ ctx }: Props) {
 
     stopRemoteNotifications();
     LocalSync.SyncState.onLogout();
+    clearUnreconciledFolders();
     clearAntivirus();
     clearIssues();
 
--- a/src/backend/features/remote-sync/unreconciled-folders/defs.ts
+++ b/src/backend/features/remote-sync/unreconciled-folders/defs.ts
@@ -1,0 +1,8 @@
+/**
+ * BR-2245
+ * A placeholder move that keeps being requested with the same origin and destination has not
+ * converged: if the previous one had worked, the origin would already be the destination. We
+ * allow a generous amount of attempts because a move can also fail for transient reasons, like
+ * the user holding a file open inside the folder.
+ */
+export const MAX_MOVE_ATTEMPTS = 10;
--- a/src/backend/features/remote-sync/unreconciled-folders/index.ts
+++ b/src/backend/features/remote-sync/unreconciled-folders/index.ts
@@ -1,0 +1,9 @@
+import { clearStore } from './store';
+
+export { addUnreconciledFolder } from './services/add-unreconciled-folder';
+export { isInsideUnreconciledFolder } from './services/is-inside-unreconciled-folder';
+export { removeUnreconciledFolder } from './services/remove-unreconciled-folder';
+
+export function onLogout() {
+  clearStore();
+}
--- a/src/backend/features/remote-sync/unreconciled-folders/index.ts
+++ b/src/backend/features/remote-sync/unreconciled-folders/index.ts
@@ -1,8 +1,11 @@
 import { clearStore } from './store';
 
+export { MAX_MOVE_ATTEMPTS } from './defs';
 export { addUnreconciledFolder } from './services/add-unreconciled-folder';
+export { clearMoveAttempts } from './services/clear-move-attempts';
 export { isInsideUnreconciledFolder } from './services/is-inside-unreconciled-folder';
 export { removeUnreconciledFolder } from './services/remove-unreconciled-folder';
+export { trackMoveAttempt } from './services/track-move-attempt';
 
 export function onLogout() {
   clearStore();
--- a/src/backend/features/remote-sync/unreconciled-folders/services/add-unreconciled-folder.ts
+++ b/src/backend/features/remote-sync/unreconciled-folders/services/add-unreconciled-folder.ts
@@ -1,0 +1,12 @@
+import { FolderUuid } from '@/apps/main/database/entities/DriveFolder';
+import { AbsolutePath } from '@/context/local/localFile/infrastructure/AbsolutePath';
+import { store } from '../store';
+
+type Props = {
+  uuid: FolderUuid;
+  path: AbsolutePath;
+};
+
+export function addUnreconciledFolder({ uuid, path }: Props) {
+  store.folders.set(uuid, path);
+}
--- a/src/context/virtual-drive/items/validate-windows-name.ts
+++ b/src/context/virtual-drive/items/validate-windows-name.ts
@@ -6,13 +6,28 @@ type TProps = {
   name: string;
 };
 
+/**
+ * v2.5.3 Daniel Jiménez
+ * These characters are invalid in windows paths.
+ *
+ * v2.6.6 Daniel Jiménez
+ * Names that start or end with a space are rejected too (BR-1796).
+ *
+ * BR-2245
+ * The rule is not whether we can create the item, but whether the rest of windows can address
+ * it afterwards. We create placeholders through `\\?\` paths, which skip the win32 name
+ * parsing, so a trailing space or dot is stored verbatim and then explorer, which does parse,
+ * looks for the trimmed name and reports that the folder does not exist. That is exactly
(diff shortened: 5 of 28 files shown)


src/context/virtual-drive/items/validate-windows-name.ts
```diff
@@ -6,13 +6,28 @@ type TProps = {
...
+ * it afterwards. We create placeholders through `\\?\` paths, which skip the win32 name
+ * parsing, so a trailing space or dot is stored verbatim and then explorer, which does parse,
+ * looks for the trimmed name and reports that the folder does not exist. That is exactly
+ * BR-1796: a folder the user can see and cannot delete. Measured, of the names we were unsure
+ * about, only the trailing ones become unaddressable, and control characters fail on creation
+ * with EINVAL. A leading space survives win32 parsing, so it no longer makes the name invalid.
+ */
+// eslint-disable-next-line no-control-regex -- matching control characters is the point: windows refuses them with EINVAL
+const forbiddenCharacters = /[<>:"/\\|?*\x00-\x1f]/;
+const unaddressableEnding = /[\s.]$/;
```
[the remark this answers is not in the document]
victor-ferro: Applied in 2a91dc48. The check goes from `\s` to `[ .]`: measured with `GetFileAttributesW`, only U+0020 (the regular space) and U+002E (the dot) leave the folder unreachable; U+00A0, U+2003, U+200B, U+202F, U+3000, U+FEFF and U+1680 are found and removed without trouble.

The infra test asked .NET, which trims more trailing whitespace than win32 does, so it could not measure this case. It now asks win32 and covers a trailing non breaking space.

src/node-win/watcher/tests/watcher-on-unlink-unreconciled.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    await writeFile(join(parent, 'child', 'file.txt'), 'content');
+
+    getFolderInfoMock.mockResolvedValue({ data: { uuid: 'parentUuid' as FolderUuid } });
+    getFolderByNameMock.mockResolvedValue({ data: { uuid: 'remoteFolderUuid' as FolderUuid } });
+    getFileByNameMock.mockResolvedValue({ data: { uuid: 'remoteFileUuid' as FileUuid } });
+  });
+
+  afterEach(() => {
+    removeUnreconciledFolder({ uuid });
+  });
```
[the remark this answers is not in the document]
victor-ferro: Applied in 5842f578, here and in the infra test next to it, which leaked the same way.

The `rm` carries retries because this hook runs before the helper unsubscribes the native watcher, so the root can still be held open.

The four pre-existing watcher tests leak the same way; leaving them out to keep the PR's scope.

src/apps/main/auth/logout.ts
```diff
@@ -35,6 +36,7 @@ export function logout({ ctx }: Props) {
 
     stopRemoteNotifications();
     LocalSync.SyncState.onLogout();
+    clearUnreconciledFolders();
```
AlexisMora, line 39: I would leave the name `onLogout` because it is the proper, given name and it would be easier to navigate. No?

src/backend/features/remote-sync/unreconciled-folders/services/add-unreconciled-folder.ts
```diff
@@ -0,0 +1,12 @@
...
+import { AbsolutePath } from '@/context/local/localFile/infrastructure/AbsolutePath';
+import { store } from '../store';
+
+type Props = {
+  uuid: FolderUuid;
+  path: AbsolutePath;
+};
+
+export function addUnreconciledFolder({ uuid, path }: Props) {
+  store.folders.set(uuid, path);
```
AlexisMora, line 11: isnt it better to add the unreconciled folder here normalized so in `src/backend/features/remote-sync/unreconciled-folders/services/is-inside-unreconciled-folder.ts` you dont have to iterate through all of them hoping for a match? this way the lookup for an unreconciled folder in `isInsideUnreconciledFolder` goes from o(nFolders) to o(1)? that could save some time down the line

src/backend/features/remote-sync/unreconciled-folders/defs.ts
```diff
@@ -0,0 +1,8 @@
+/**
```
AlexisMora, line 1: I would rename this folder to constants to keep consistence along the project

src/backend/features/remote-sync/unreconciled-folders/index.ts
```diff
@@ -0,0 +1,12 @@
...
+export { MAX_MOVE_ATTEMPTS } from './defs';
+export { addUnreconciledFolder } from './services/add-unreconciled-folder';
+export { clearMoveAttempts } from './services/clear-move-attempts';
+export { isInsideUnreconciledFolder } from './services/is-inside-unreconciled-folder';
+export { removeUnreconciledFolder } from './services/remove-unreconciled-folder';
+export { trackMoveAttempt } from './services/track-move-attempt';
+
+export function onLogout() {
+  clearStore();
+}
```
AlexisMora, line 12: Isnt it better to just export and use clearStore and give the proper use where we consume it??

PR #1502, opened 2026-09-15
feat: add main mail bridge lifecycle process

AlexisMora:
## What is Changed / Added
In this pull request I introduce mail bridge ipc handlers for status, activation, and shutdown, and exposes it through electron preload. It also adds the mail bridge section and connects the core mail bridge views to real lifecycle states, including startup failures and the Mail-account-required state

---

## Why
 This change connects settings view to that lifecycle through a small ipc surface, allowing users to activate and manage the local bridge


src/apps/renderer/pages/Settings/index.tsx
```diff
@@ -28,17 +28,50 @@ type Props = {
...
+  const [mailBridgeViewModel, setMailBridgeViewModel] = useState(() => MailBridgeModule.createInitialViewModel());
   const { data: availableProducts, isLoading: isAvailableProductsLoading } = useGetAvailableProducts();
 
   const rootRef = useRef<HTMLDivElement>(null);
 
+  useEffect(() => {
+    void window.electron.mailBridge.getStatus().then((status) => {
+      if (status.status === 'stopped') setMailBridgeViewModel(MailBridgeModule.createInitialViewModel());
+      if (status.status === 'starting') setMailBridgeViewModel({ status: 'starting', error: null });
+      if (status.status === 'error') setMailBridgeViewModel({ status: 'error', error: status.error });
```
xabg2, line 40: nit: You can extract this to a separated function, using a switch and checking the status. Then, you can use it below, in onStatusChanged, as it is the same.

xabg2 (approved):

